PR #4, opened 2022-02-15
First test for MiniDexed

dcoredump:
I tried to get everything working and copied a lot directly from miniorgan. Compiling works, but I couldn't really test it.

Important: Synth_Dexed must still be updated from coderberg!


probonopd:
Thank you very much @dcoredump.

Since I don't know how to fix the git submodules related merge conflict, let's continue in #5.

PR #31, opened 2022-02-27
Fix order of members in minidexed.h

rsta2:
There were some warnings, because the member objects in *minidexed.h* weren't in the same order as the initializers in the constructor. This fixes it.


probonopd:
Thanks @rsta2.

PR #34, opened 2022-02-28
Fix: Invalid handling of .syx file indexes

rsta2:
There is another issue in *sysfileloader.cpp* left. Please see the commit description and diff for info.


probonopd:
Thanks @rsta2

PR #37, opened 2022-03-02
Restructure code and add new features

rsta2:
This pull request provides the following changes:

* Support multiple USB MIDI inputs at once (two on Raspberry Pi 1-3, four on Raspberry Pi 4).

* Now all MIDI inputs (USB, serial, PC keyboard) can work simultaneous.

* Program change and bank select work with serial MIDI.

* The synth configuration is editable for the user now, without modifying the code. There can be a file minidexed.ini on the SD card, which contains the configuration values (see src/minidexed.ini for a template). The configuration of the sound device ("pwm", "i2s" or "hdmi") is done in minidexed.ini now. The option `sounddev=` in cmdline.txt does not have an effect any more.

* Support headless operation on Raspberry Pi 4.

* Suppress frequent messages in MIDI dump (timing clock, active sensing).

* Some Dexed methods require to be guarded from being interrupted by other Dexed calls. This is done in the class CDexedAdapter.

* LCD writes from an IRQ handler are allowed now. The LCD code has been moved to the class CUserInterface, which can be extended to provide an user interface.

* Fix: Performance timer did not show correct percent value with HDMI.

* Code restructured and headers added to all files. This is a suggestion only and can of course be modified.

There is a problem with the default GPIO pin configuration for the LCD. The I2S interface uses GPIO18-21, and GPIO18-19 is currently also used by the LCD. So I2S will only work, when the LCD is disabled in minidexed.ini.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,8 +63,8 @@ jobs:
            mkdir -p sdcard
            cp -r ./circle-stdlib/libs/circle/boot/* sdcard
            rm -rf sdcard/config*.txt sdcard/README sdcard/Makefile sdcard/armstub sdcard/COPYING.linux
-           cp ./src/config.txt ./src/*img sdcard/
-           echo "usbspeed=full sounddev=sndpwm" > sdcard/cmdline.txt
+           cp ./src/config.txt ./src/minidexed.txt ./src/*img sdcard/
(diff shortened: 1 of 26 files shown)


probonopd:
Thanks for this massive PR @rsta2. Highly appreciated!

PR #41, opened 2022-03-12
Added support for CMSIS_5, so upstream Synth_Dexed can be used

dcoredump:
I fixed upstream [Synth_Dexed](https://codeberg.org/dcoredump/Synth_Dexed) so this library supports now Teensy **and** circle - no need for an own `Synth_Dexed` library. This makes working on `Synth_Dexed` easier because only one code base for both platforms is needed.

Therefore support for `CMSIS_5` was needed (and is also needed for further additions, like stereo panning and compressor). So `CMSIS_5` is now a submodule of `MiniDexed`.

For CMSIS_5 perhaps some optimizations are needed (like support for NEON).

I hope this PR is useful for further development.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,6 @@
 [submodule "Synth_Dexed"]
 	path = Synth_Dexed
 	url = https://codeberg.org/dcoredump/Synth_Dexed
-	branch = dev
 [submodule "CMSIS_5"]
 	path = CMSIS_5
 	url = https://github.com/ARM-software/CMSIS_5
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,8 @@
 	url = https://github.com/smuehlst/circle-stdlib
 [submodule "Synth_Dexed"]
 	path = Synth_Dexed
-	url = https://github.com/probonopd/Synth_Dexed
+	url = https://codeberg.org/dcoredump/Synth_Dexed
+	branch = dev
+[submodule "CMSIS_5"]
+	path = CMSIS_5
+	url = https://github.com/ARM-software/CMSIS_5
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/smuehlst/circle-stdlib
 [submodule "Synth_Dexed"]
 	path = Synth_Dexed
-	url = https://codeberg.org/dcoredump/Synth_Dexed
+	url = https://codeberg.org/dcoredump/Synth_Dexed.git
 [submodule "CMSIS_5"]
 	path = CMSIS_5
 	url = https://github.com/ARM-software/CMSIS_5
--- a/README.md
+++ b/README.md
@@ -104,6 +104,9 @@ cd -
 
 # Make zip that contains Raspberry Pi 4 boot files. The contents can be copied to a FAT32 formatted partition on a microSD card
 mkdir -p sdcard
+cd sdcard
+../getsysex.sh
+cd ..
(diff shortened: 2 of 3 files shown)


probonopd:
Thank you very much @dcoredump. Looking forward to multiple Dexed instances, stereo panning, and other effects :+1:

probonopd:
Oopsie. This destroyed Raspberry Pi Zero and 2 (did not test others).

* HDMI sound loud "random" noise
* PWM no sound

Which models did you test it on @dcoredump?

I perserved to build at https://github.com/probonopd/MiniDexed/releases/tag/CMSIS_5.

Please let me know how I can help debugging/testing.

dcoredump:
Oh, Strange. I tested with a RPi4. Will check this.

dcoredump:
I have now adjusted the build process in README.md a little bit (except for one problem, it works, but for the Raspi2 there is still `kernel=kernel8.img` instead of `kernel=kernel7.img` inside `config.txt` for `[pi2]` - I have adjusted this manually).

If I run everything freshly after the recipe inside README.md it runs with me for the Raspi2, 3 and 4 (audio over PWM)!

PR with fixes for the README.md is comming soon.

probonopd:
I think the `config.txt` was wrong. Changed it in the PR. (I wonder how it could have worked, ever...)

probonopd:
According to https://www.raspberrypi.com/documentation/computers/config_txt.html#model-filters,

Filter | Applicable model(s)
-- | --
[pi1] | Model 1A, Model 1B, Model 1A+, Model 1B+, Compute Module 1
[pi2] | Model 2B (BCM2836- or BCM2837-based)
[pi3] | Model 3B, Model 3B+, Model 3A+, Compute Module 3, Compute Module 3+
[pi3+] | Model 3A+, Model 3B+
[pi4] | Model 4B, Pi 400, Compute Module 4
[pi400] | Pi 400
[cm4] | Compute Module 4
[pi0] | Zero, Zero W, Zero 2 W
[pi0w] | Zero W
[pi02] | Zero 2 W
[board-type=Type] | Filter by Type number - see Raspberry Pi Revision Codes E.g [board-type=0x14] would match CM4.

And according to https://www.raspberrypi.com/documentation/computers/config_txt.html#kernel,

> kernel is the alternative filename on the boot partition to use when loading the kernel. The default value on the Raspberry Pi 1, Zero and Zero W, and Raspberry Pi Compute Module 1 is kernel.img. The default value on the Raspberry Pi 2, 3, 3+ and Zero 2 W, and Raspberry Pi Compute Modules 3 and 3+ is kernel7.img. The default value on the Raspberry Pi 4 and 400, and Raspberry Pi Compute Module 4 is kernel7l.img.
>
> arm_64bit
If set to non-zero, forces the kernel loading system to assume a 64-bit kernel, starts the processors up in 64-bit mode, and sets kernel8.img to be the kernel image loaded, unless there is an explicit kernel option defined in which case that is used instead. Defaults to 0 on all platforms.

So I think we can simplify the file, I did this in the PR.

dcoredump:
Thanks @probonopd !

I managed to change the name of the kernel file with `sed`. But perhaps it would be better to set the name for the kernel file to something like "minidexed.img"?

rsta2:
The Circle build system uses the standard naming convention `kernel*.img`. If you want to use an other naming, you have to copy the file after build. Please don't rename it, because some useful tools (e.g. serial bootloader) require the standard naming to work.

probonopd:
Agree @rsta2, let's stay to the conventions.

PR #47, opened 2022-03-14
Multi-core support

rsta2:
This PR modifies the following:

* Update submodule circle-stdlib to v15.12
	* System options can be defined cleaner using the "-o" option
	* Include KY-040 driver from Circle (removed from MiniDexed)
* Render sound on secondary CPU core 1
	* Enable multi-core support on Raspberry Pi 2-4
	* Does still work on the Raspberry Pi 1 with restrictions
	* Use CSoundBaseDevice::Write() instead overriding GetChunk()
	* CMiniDexed is not derived from the sound device classes any more
	* Add option SCREEN_DMA_BURST_LENGTH=1 to relieve bus congestion
* Add volume control to MIDI CC and UI

There is a new menu item in the UI, which controls the volume level.

The MIDI message dump and profiling should be disabled on the Raspberry Pi 1, otherwise one hears audio drops, when the screen scrolls.

Tested on RPi 1A+ (PWM, HDMI), 2B (PWM, I2S, HDMI) and 4B (PWM, I2S, HDMI). The UI has been tested with a RPi 1A+ only (because my hardware is connected there).


probonopd:
Thanks @rsta2. Here are my test results:

Raspberry Pi Zero with HDMI: Works if MIDI keyboard is directly attached to Pi. Does not work if MIDI keyboard is attached to Official Raspberry Pi Keyboard and Hub which is attached to Pi, getting `dwroot: Previous attempt to initialize device failed`. When I remove the Official Raspberry Pi Keyboard and Hub at that point and attach the MIDI keyboard directly, then I get 3x `dwroot: Previous attempt to initialize device failed`. Rebooting at that results in a working MIDI keyboard, I don't get the error message then.
As you write, there is stutter with the debug options enabled. Hence, I would disable the debug output by default.

Raspberry Pi 2 with HDMI: Works if MIDI keyboard is directly attached to Pi. Does not work if MIDI keyboard is attached to Official Raspberry Pi Keyboard and Hub which is attached to Pi, getting `usbdev0-1: Cannot get device descriptor (short)`. When I remove the Official Raspberry Pi Keyboard and Hub at that point and attach the MIDI keyboard directly, then it works (without the need for a reboot). Additionally attaching the Official Raspberry Pi Keyboard and Hub gives `usbdev0-1: Cannot get device descriptor (short)` and does not result in a working ASCII keyboard, also not after a reboot.

Did not test Raspberry Pi 3 and 4 yet.

Volume knob on MIDI keyboard works. Pitch Bend and modulation don't seem to do anything that I could hear.

rsta2:
@probonopd Thanks for testing. There seems to be a problem with the Official Raspberry Pi Keyboard and Hub with Circle. This is probably independent from this PR. It should have happened with the previous MiniDexed version too? Can you try the option `usbpowerdelay=1000` in *cmdline.txt* with this keyboard?

The debug output is disabled by default, when the respective options are not found in *minidexed.ini*. But they are present and enabled in the default .ini file. I think, we should remove them there for production.

I tried to implement Pitch Bend and modulation, but because it did not work here, I did not included this in the PR. So it will not work at the moment.

What happened to the lost MIDI events? Does this still happen?

Nice, that you like the LCD volume display. :)

rsta2:
I think, we had this problem with your USB keyboard and hub before on the RPi 4. It was gone after the boot ROM update. Strange.

rsta2:
Yes, please remove the `usbspeed=full` option.

Maybe the RPi 1 (and Zero) is too slow, to handle USB MIDI without event loss in time. The USB HCI controller emits 8000 interrupts per second. That's huge.

rsta2:
> Note that I can reproduce "stuck notes" on the Raspberry Pi 2 using the MIDI keyboard as well. (To reproduce: Use "ORGAN 1" voice and play rapid glissando; after a while you will hear a "stuck" note.)

Is this with `usbspeed=full` or without?

probonopd:
Without.

rsta2:
OK. That's bad. You should try this on the RPi 3, when you have some time. mt32-pi did work for you, and it does not run on the RPi 2.

I did not hear stuck notes at all with all MiniDexed versions so far with my AKAIpro LPK25 keyboard. It seems to behave differently then yours. I cannot reproduce this.

probonopd:
> mt32-pi did work for you

To be fair, I didn't stress-test it to the same extent ;-)

As the next step, let me see if I can reproduce this using a Raspberry Pi Pico acting as a MIDI keyboard playing notes in really fast succession...

rsta2:
Tried "E.ORGAN 1" (voice 17) from the default bank again, sliding for one minute up and down on the keyboard, using a RPi 2B. No stuck note.

Perhaps this is more a test for midi_adapter on the Pico, because it handles the USB in this scenario.

probonopd:
Strange...
I am not sure whether this is really missed key presses or the synth engine getting stuck somehow. 

If I play the same note several times, then the MIDI OFF signal should make that note go silent no matter how often a MIDI ON signal for that note was previously received, correct? So by playing each note slowly I should get all notes back to OFF state. But this is not what happens. Some notes keep playing.

(I was not able to reproduce this with a Rpi Zero test script yet.)

rsta2:
This synth engine behaves differently. This is something, that I noticed too. If you send two "note on" for a specific note, you also have to send two "note off" for that note. Otherwise it will not stop generating sound. I can easily reproduce this here, because I can send single MIDI bytes here via the serial line.

probonopd:
With [MiniDexed_2022-03-14.zip](https://github.com/probonopd/MiniDexed/releases/download/continuous/MiniDexed_2022-03-14.zip) I _cannot_ reproduce the issue. I only get the issue with the build from this PR.

rsta2:
Yes, the question remains, why some "note off" events get lost, and if this happens on the RPi 3, and with mt32-pi? I suppose with `usbspeed=full`, it will also work with the version with PR applied. So it would be no step back, at least. BTW. Your Official RPi keyboard does obviously not work with USB full speed.

probonopd:
> So it would be no step back, at least.

That's true.
With `usbspeed=full` I don't get suck notes but the Official Raspberry Pi Keyboard and Hub doesn't work. Regardless in which version.

> the question remains, why some "note off" events get lost, and if this happens on the RPi 3, and with mt32-pi?

I'll try that next.

probonopd:
Without `usbspeed=full` I _do_ get stuck notes also on the Rpi 3 using the version from this PR.

Using mt32-pi-0.11.0 on the same hardware and using an Organ-style voice (= a voice that plays forever if you keep the key pressed forever) I cannot reproduce the stuck notes issue.

I notice it has (only) `fast=true` in _cmdline.txt_.

rsta2:
Can you try mt32-pi without `fast=true`? With `fast=true` and the class `CCPUThrottle` in the system, the RPi 3B runs twice as fast (1200 MHz). Perhaps this is the right time to add support for enabling the full CPU speed to MiniDexed? It can be switched with this *cmdline.txt* option.

probonopd:
With _quite_ some forceful effort I can get stuck notes on mt32-pi without `fast=true`. But the difference is that mt32-pi stops playing the stuck note when I hit the same key again. (Which is possibly not according to the MIDI spec?)

rsta2:
I'm quickly preparing a full CPU speed patch for MiniDexed with PR #47.

Yes, this is a difference in the synth engine.

rsta2:
Here it is:

[full-speed.zip](https://github.com/probonopd/MiniDexed/files/8248658/full-speed.zip)

You still have to add the option `fast=true` to *cmdline.txt*. Watch out for the SpeedFactor in one of the log first lines. It should be greater now.

probonopd:
Does the full-speed patch have any downsides (as long as `fast=true` is not actively set)? If it doesn't hurt could you apply the patch to https://github.com/rsta2/MiniDexed/tree/multi-core-support; once you do a new build will be triggered here automatically which would ease my testing. Thanks!

rsta2:
Done.

rsta2:
BTW. With the option `socmaxtemp=78` you can set the SoC temperature, at which the CPU speed is throttled again, when it gets too hot. 78°C is the maximum, 60°C the default.

probonopd:
Can still produce stuck keys with (only) `fast=true` on Raspberry Pi 2 and 3.

rsta2:
I would merge it, because we need the multi-core support anyway, and it's not a step back. The problem was there before. mt32-pi is very sophisticated after a long development. I think we cannot expect to reach the same after a month. There is room for improvement, that's for sure.

probonopd:
Thanks @rsta2

rsta2:
You are welcome.

Issue #9, opened 2022-02-20
Support the volume knob, modulation, and pitch bend on MIDI controllers


probonopd:
Support the volume knob on MIDI controllers.

(Once #8 is implemented I can say which MIDI message the volume knob sends.)

dcoredump:
Will do this with adding MIDI processing support for Synth_Dexed (see also #10, #11)

probonopd:
The volume MIDI messages are `B0 07 00` (knob all the way left) to `B0 07 7F` (knob all the way right).

probonopd:
Likewise, `B0 01 ...` for modulation and `E0 00 ...` for Pitch Bend.

rsta2:
I was so bold and implemented the volume control (using MIDI CC and via the UI) already in #47. ;) I also tried modulation (using `Dexed::setModWheel()`) and pitch bend (using `setPitchbend()`), but this didn't work. Or I wasn't able to "move the wheel" quick enough, because I can only simulate the MIDI messages by sending hex bytes manually. I also don't know, if it depends on the loaded voice, if these functions have an effect.

BTW. Pitch bend was already implemented, but I think this MIDI message has two parameter bytes (`0xE0 LSB MSB`), and the MSB is not considered yet.

dcoredump:
> I was so bold and implemented the volume control (using MIDI CC and via the UI) already in #47. ;) I also tried modulation (using `Dexed::setModWheel()`) and pitch bend (using `setPitchbend()`), but this didn't work. Or I wasn't able to "move the wheel" quick enough, because I can only simulate the MIDI messages by sending hex bytes manually. I also don't know, if it depends on the loaded voice, if these functions have an effect.

ModWheel depends on the routing of the voices. Some voices don't have a target for the ModWheel. 



> BTW. Pitch bend was already implemented, but I think this MIDI message has two parameter bytes (`0xE0 LSB MSB`), and the MSB is not considered yet.

Yes, Pitchbend is 14 bit (LSB, MSB)!

probonopd:
> E0 00 ... for Pitch Bend.

To be precise, `E0 00 40` (normal pitch wirthout bending) all the way down to `E0 00 00` and all the way up to `E0 7F 7F`.

rsta2:
@dcoredump Ah OK, I probably used a voice, which don't use the ModWheel. Thanks for info.

[...]

Issue #10, opened 2022-02-20
Support the sustain button on MIDI controllers


probonopd:
Support the sustain button on MIDI controllers.

(Once https://github.com/probonopd/MiniDexed/issues/8 is implemented I can say which MIDI message the sustain button sends.)

dcoredump:
Will do this with adding MIDI processing support for Synth_Dexed (see also #9, #11)

rsta2:
Here is a small patch to implement MIDI CC Sustain (Damper). The behavior is different from the damper pedal of my Yamaha home keyboard. When you release the sustain button, the note does not go off. You have to press the key again, to switch it off. This is managed by Synth_Dexed. @dcoredump I don't know, if it is intended?

[sustain.zip](https://github.com/probonopd/MiniDexed/files/8256855/sustain.zip)

probonopd:
On my M-AUDIO Keystation Mini 32 Mk3 a LED goes on when I press the sustain key (sending `B0 40 7F`), and the LED goes off when I press the same key again (sending `B0 40 00`).

So I think it is correct as it is implemented: A sustain button works differently from a sustain pedal.

probonopd:
Pressing the sustain key works.

But pressing it again, the already sustained notes keep playing forever. This is not like it should be.

rsta2:
The sustain code in Dexed is [here](https://github.com/asb2m10/dexed/blob/master/Source/PluginProcessor.cpp#L386). I think line 386-393 have to be added to `Dexed::setSustain()` in Synth_Dexed. This will trigger the `keyup()` events automatically, when the sustain button/pedal is released.

rsta2:
Here is a patch for Synth_Dexed, which enables the full sustain support:

```
diff --git a/src/dexed.cpp b/src/dexed.cpp
index be43a27..a93521a 100644
--- a/src/dexed.cpp
+++ b/src/dexed.cpp
@@ -406,6 +406,15 @@ void Dexed::setSustain(bool s)
     return;
 
   sustain = s;
+
+  if (!sustain) {
+      for (int note = 0; note < max_notes; note++) {
+          if (voices[note].sustained && !voices[note].keydown) {
+              voices[note].dx7_note->keyup();
+              voices[note].sustained = false;
+          }
+      }
+  }
 }
 
 bool Dexed::getSustain(void)
```

It's tested, it works.

probonopd:
@dcoredump could you add it please? Thanks.

[...]

PR #49, opened 2022-03-15
Implement Pitch Bend and Modulation CC, thanks @rsta2

probonopd:
https://github.com/probonopd/MiniDexed/issues/9#issuecomment-1067662867


probonopd:
Merged since it is definitely an improvement; however it may still need refinement; see #9 and #10.
Thanks @rsta2

PR #50, opened 2022-03-19
Support multiple Dexed instances

rsta2:
This PR adds support for multiple Dexed instances (Tone Generators, TG). It's not possible to load the performance configuration from a file at the moment. I will add this soon.

Currently voice bank, voice number, volume and assigned MIDI channel (1-16, OMNI mode or OFF) can be selected from the UI for each TG. The TGs can be switched by double clicking the switch. When you hold the switch for one second, you will go back to the home position (TG1, VOICE) in the menu. By default TG1 is in OMNI mode, all other TGs are not assigned to a MIDI channel. The PC keyboard fakes MIDI events on channel 1.

The mixer is not perfect yet. That's why the volume is reduced, when only a few TGs are active. I had to reduce the `MaxNotes` value for Raspberry Pi 1 (8 now) and increased the chunk size default for RPi 1 to 1024. Otherwise there are audio drops. The chunk size is limited to a value of 4096 now.

I tested this on RPi 3B (with 32 and 64-bit and UI) and RPi 1 (without UI) only. It is noticeable, that the CPU load could rise to a critical level with eight active TGs. CPU core 1 handles two TGs, core 2 and 3 handle three TGs each. Core 0 is free to handle all other stuff (MIDI, sound driver, UI). `fast=true` in *cmdline.txt* may be required, at least on the RPi 2.

Let me know, when you need modifications.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -38,7 +38,11 @@ void CConfig::Load (void)
 	m_SoundDevice = m_Properties.GetString ("SoundDevice", "pwm");
 
 	m_nSampleRate = m_Properties.GetNumber ("SampleRate", 48000);
+#ifdef ARM_ALLOW_MULTI_CORE
 	m_nChunkSize = m_Properties.GetNumber ("ChunkSize", m_SoundDevice == "hdmi" ? 384*6 : 256);
+#else
+	m_nChunkSize = m_Properties.GetNumber ("ChunkSize", m_SoundDevice == "hdmi" ? 384*6 : 1024);
+#endif
 	m_nDACI2CAddress = m_Properties.GetNumber ("DACI2CAddress", 0);
 
 	m_nMIDIBaudRate = m_Properties.GetNumber ("MIDIBaudRate", 31250);
--- a/src/config.h
+++ b/src/config.h
@@ -30,7 +30,13 @@
 class CConfig		// Configuration for MiniDexed
 {
 public:
+#if RASPPI == 1
+	static const unsigned ToneGenerators = 1;
+	static const unsigned MaxNotes = 8;		// polyphony
+#else
+	static const unsigned ToneGenerators = 2;
 	static const unsigned MaxNotes = 16;		// polyphony
+#endif
 
 #if RASPPI <= 3
 	static const unsigned MaxUSBMIDIDevices = 2;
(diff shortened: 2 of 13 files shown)


probonopd:
Wow, this is HUGE. :1st_place_medal: Testing now.

probonopd:
When I leave TG1 as it is and switch TG2 to OMNI and use a different voice, then I still hear only TG1.
Am I doing it wrong? (RPi 2.)

probonopd:
Wonderful. Now imagine there'd be controls (similar to volume) for Stereo balance and for detune (an existing Dexed parameter). Instant TX802 feeling :+1:

probonopd:
fwif, so far I did not notice the need for `fast=true` on RPi 2.

rsta2:
@probonopd Thank you very much! :)

I think Stereo balance cannot be implemented in Synth_Dexed that easy, at least not in the `Dexed` class, because the delivered sound data is Mono at the moment. One can multiply the samples with different factors to create a Stereo shift. But the clipping problem can occur then again. I don't know, if @dcoredump wants to change this, to provide Stereo sound data. For the Reverb FX it is also necessary to have Stereo.

I'm not sure at the moment because of the Detune parameter, which exists in each of the six OPs in DX7/Dexed. So if you have a Detune setting of (say) +4 in the Performance, how does this influence the six Detune values for the OPs? We have to find this out, before we can add this to the UI menu.

Ah OK, I did not test it on the RPi 2. But I have seen load values of over 60% on the RPi 3B with all TGs active, so I thought it could be tight on the RPi 2. But it's good, when it is working there too.

BTW. I have to go somewhere today and will be back tomorrow.

probonopd:
> I'm not sure at the moment because of the Detune parameter, which exists in each of the six OPs in DX7/Dexed.

I am looking for the detune that detunes a whole Dexed instance overall (TG). This exists in the Dexed GUI.

rsta2:
> I am looking for the detune that detunes a whole Dexed instance overall (TG). This exists in the Dexed GUI.

Great hint! After searching for this in the Dexed code, I found that it has a `masterTune` parameter and Synth_Dexed already [has it](https://codeberg.org/dcoredump/Synth_Dexed/src/branch/master/src/dexed.cpp#L811) too.

PR #51, opened 2022-03-20
Add detune and Stereo balance

rsta2:
This adds Detune and Stereo balance (pan) items to the UI menu.

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -41,6 +41,9 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	m_SerialMIDI (this, pInterrupt, pConfig),
 	m_bUseSerial (false),
 	m_pSoundDevice (0),
+#ifdef ARM_ALLOW_MULTI_CORE
+	m_nActiveTGsLog2 (0),
+#endif
 	m_GetChunkTimer ("GetChunk",
 			 1000000U * pConfig->GetChunkSize ()/2 / pConfig->GetSampleRate ()),
 	m_bProfileEnabled (m_pConfig->GetProfileEnabled ())
@@ -50,6 +53,7 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	for (unsigned i = 0; i < CConfig::ToneGenerators; i++)
 	{
 		m_nVoiceBankID[i] = 0;
+		m_nPan[i] = 64;
 
 		m_pTG[i] = new CDexedAdapter (CConfig::MaxNotes, pConfig->GetSampleRate ());
 		assert (m_pTG[i]);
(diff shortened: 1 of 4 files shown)


probonopd:
@rsta2 this is so cool!

Can it be that left and right are reversed in the GUI?

rsta2:
@probonopd Thanks! Normally not. But swapping the left and right channels did occur earlier, especially with PWM in Circle. With which RPi model and which sound device did you noticed this?

probonopd:
RPi 2, HDMI. If useful I can test other combinations tomorrow.

rsta2:
I tested RPi 3B with PWM and HDMI and the left/right seems to be OK there. I will test RPi 2 with HDMI tomorrow here.

probonopd:
I am going from the RPi 2 to a HDMI display then via its headphone jack to my headphones. Cannot exclude the possibility that the display is mixing it up. Will do systematic tests tomorrow.

rsta2:
OK. Yes, it's fun! ;)

rsta2:
I tested the Stereo balance on two TV sets, and you are right, my own TV set swaps the channels with HDMI. Because I used this TV set as a reference, now the Circle HDMI driver swaps the channels and all existing Stereo application probably have the wrong channel assignment. :( Maybe some developers have recognized this already and swapped the channels back in their software.

I think we need an option in *minidexed.ini*, which allows to assign the channels, as it is wanted. Perhaps there are other TV sets out there, which need this.

BTW. I think this problem exists only with HDMI. PWM and I2S should be OK.

rsta2:
PR #52 fixes the problem.

probonopd:
Confirmed. Thanks @rsta2.

One strange observation:

PWM is relatively noisy and in one of the Stereo positions it is even more grave than in all others. I wonder why this is. To reproduce:

* Use a Rpi 2 with PWM (did not test this on other models so far) and ("good") headphones
* Restart RPi for default settings
* Bring the volume down to a useful level
* Play one high-pitched note very lightly and keep it pressed. Notice that it is relatively noisy while the note is playing
* While the note is still playing, via the GUI, change the stereo shifting and pay attention to the amount of noise. In the setting where the 9th dot from the left = 8th from the right is selected, the noise appears to be much more than in any other Stereo setting - can this be explained?

rsta2:
I would guess, that this is caused by rounding errors. There are only 16 bits resolution of the audio samples and for PWM it has to be round down to 11 bits.

With the mixer for the signals of the different TGs and the Stereo pan, we come to the situation, that not all audio processing is done in Synth_Dexed any more. It's a difference to program some audio, and to program it in a way without noise and other problems. I'm not a professional audio programmer and maybe I should keep away from such tasks.

BTW. I cannot hear this noise. Maybe because I'm on the RPi 3B, because my headphones are not good enough, or my ears. ;) There is one thing, I can imagine. Maybe the TG mixer has to be protected by a spin lock, because when changing the Stereo balance, this may happen while the mixer is running and may influence the resulting signal.

probonopd:
Thanks, will do more testing on other devices as well.

PR #52, opened 2022-03-21
Fix Stereo channel swap with HDMI

rsta2:
There has been a swap of the Stereo channels with HDMI. With the new
option "ChannelsSwapped=0|1" in minidexed.ini one can swap the channels
again, in case it is needed.


probonopd:
Confirmed, this works for me on Rpi 2 HDMI. Thanks @rsta2

PR #57, opened 2022-03-26
Add note limits and shift to performance.ini

rsta2:
This adds note limits and note shift to *performance.ini*. These parameters are currently not editable from the UI. The note numbering is as follows:

https://computermusicresource.com/midikeys.html


probonopd:
Cool! Works for me Thanks @rsta2

PR #59, opened 2022-03-30
Added code for enabling compressor (compressor code is in Synth_Dexed) / Fixes for README.md / Added stereo plate reverb

dcoredump:
I now added compressor code to Synth_Dexed so it is available inside MiniDexed. The compressor is enabled in [src/dexedadapter.h](https://github.com/probonopd/MiniDexed/compare/main...dcoredump:main#diff-2dd9dd0384db1fd7d8c7f63eae2fddd83630d3abf933cccdff6bb2384cc35798).

The initial setup of the compressor is:

- Threshold: -20.0 dB
- Ratio: 5.0
- Attack: 5ms
- Release: 200ms
- HP-Filter: enabled

List of new methods in Synth_Dexed:
```
void setCompressor(bool comp);
bool getCompressor(void);
void setCompressorPreGain_dB(float32_t pre_gain);
void setCompressorAttack_sec(float32_t attack_sec);
void setCompressorRelease_sec(float32_t release_sec);
void setCompressorThresh_dBFS(float32_t thresh_dBFS);
void setCompressionRatio(float32_t comp_ratio);
float32_t getCompressorPreGain_dB(void);
float32_t getCompressorAttack_sec(void);
float32_t getCompressorRelease_sec(void);
float32_t getCompressorThresh_dBFS(void);
float32_t getCompressionRatio(void);
```

I also fixed some build instructions inside the README.md.

Also now added first implementation of the stereo plate reverb (currently with fixed parameters).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,27 +88,29 @@ __CAUTION:__ All GPIO numbers are [chip numbers](https://pinout.xyz/), not heade
 E.g., to build for Raspberry Pi 4 on a Ubuntu 20.04 build system, you can use the following example. See [`build.yml`](../../tree/main/.github/workflows/build.yml) for complete build steps that create versions for Raspberry Pi 1, 2, 3,and 4 in 32-bit and 64-bit as required.
 
 ```
+# Choose your RPi
+export RPI=4
+
 git clone https://github.com/probonopd/MiniDexed
 cd MiniDexed
+mkdir -p kernels sdcard
 
 # Recursively pull git submodules
 git submodule update --init --recursive
 
-# Choose your RPi
-export RPI=4
-
 # Install toolchain
 if [ "${RPI}" -gt 2 ]
 then
-	wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf.tar.xz
+	wget https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf.tar.xz
 else
-	wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-arm-none-eabi.tar.xz
+	wget https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-arm-none-eabi.tar.xz
 fi
-tar xf gcc-arm-*-*.tar.xz 
+tar xvf gcc-arm-*-*.tar.xz 
 export PATH=$(readlink -f ./gcc-*/bin/):$PATH
 
 # Build dependencies and MiniDexed
 ./build.sh
+cp ./src/kernel*.img ./kernels/
 
 # Get Raspberry Pi boot files
 cd ./circle-stdlib/libs/circle/boot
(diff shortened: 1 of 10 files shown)


probonopd:
Thank you very much @dcoredump. I tested this on a RPi 2 with the headphone jack and with HDMI. With HDMI I get all sorts of strage (and interesting sounding! but definitely unintentional) sounds, which I don't get with PWM (headphone jack). Did you have a chance to test this with HDMI yet?

dcoredump:
> Isn't it possible to hand over the `ReverbBuffer[]` directly to `m_pSoundDevice->Write()` in _minidexed.cpp_ line 607, without copying it to `SampleBuffer[]` before? This would save some processing time. I think both buffers have the same contents after copying.

The reverb handling code came in by mistake...

I still have to adjust some things, but I'm on a good way. I will add the reverb internally to the signal, then the reverb array is omitted.

dcoredump:
> Thank you very much @dcoredump. I tested this on a RPi 2 with the headphone jack and with HDMI. With HDMI I get all sorts of strage (and interesting sounding! but definitely unintentional) sounds, which I don't get with PWM (headphone jack). Did you have a chance to test this with HDMI yet?

I'm sorry, but I haven't been able to test it with HDMI yet. I'll get myself a HDMI->audio converter.

I would be very surprised that it works with PWM but not with HDMI. I actually didn't do anything else but put the compressor at the end of `getSamples()`.

Can you try setting Dexed::setCompressor(true) to false in `minidexed.cpp:36` and try again?

probonopd:
> Can you try setting Dexed::setCompressor(true) to false in minidexed.cpp:36 and try again?

I don't see that [there](https://github.com/dcoredump/MiniDexed/blob/2a20eb82a08d7a59385c3efa98960af0c346be12/src/minidexed.cpp#L36)...

probonopd:
@rsta2 does [this](https://github.com/probonopd/MiniDexed/suites/5894295837/artifacts/200187085) build work for you with HDMI?

rsta2:
@probonopd Yes, the *kernel7.img* works with PWM and HDMI. Please note that the size of these kernel images is hard at the limit or over the limit of the default maximum kernel size of 2 MBytes. For bigger kernel images you have to specify the option `-o KERNEL_MAX_SIZE=0x400000` (e.g. for 4 MBytes) to `configure` in *build.sh*. That's why the other kernels do not work, because they are bigger.

probonopd:
That seems to have done the trick!

Wow, what a cool sound. Thanks @dcoredump.

> I still have to adjust some things, but I'm on a good way. I will add the reverb internally to the signal, then the reverb array is omitted.

Please let me know when this is ready for being merged. Thanks!

dcoredump:
We need a mixer object, so we can do a mix of all instances (what is currently working) and a mix of the reverb sends of the instances, so every instance can be adjusted of how many reverb should be send.

I think it max also be better to use a float32_t as type for the signal chain instead of int16_t. Why? Most effects use internaly floats because they are easier to handle. For "small" microcontrollers, which do not have a FPU, integeger arithmetic must be used, but for a Raspi this is not needed. With floats we can drop every integer->float converting (and back) for every effect we want to use...

I will try to create a mixer class and switch to floats in the signal chain. Only at the end of the signal chain `arm_float_q15()` has to be called to use the Neon-DSP functions to convert as fast as possible to integer.

dcoredump:
Forgot to say: you can merge this and perhaps add the reverb setup to the UI. Or you can wait until I am rwady with my tasks from above.

probonopd:
Thanks @dcoredump. @rsta2 wdyt, should I merge this now so that it could be integrated into the GUI?

rsta2:
@probonopd You can merge PR #59 now. Sounds really cool.

probonopd:
Merged!

probonopd:
@dcoredump: What does the __Send__ parameter do? I'd like to add it to the [documentation](https://github.com/probonopd/MiniDexed/wiki/Menu).

PR #67, opened 2022-04-03
userinterface: Reverse click and double-click events

rsta2:
* Double-click is select
* Single-click is back


probonopd:
Changed https://github.com/probonopd/MiniDexed/wiki/Menu accordingly.

PR #68, opened 2022-04-03
uimenu: Fix: Edit OP parameter from selected TG

rsta2:
When editing an OP parameter, the selected TG was not used,
but always TG1 instead.


rsta2:
Sorry for the mistake. Unfortunately there is not enough room on the LCD to display both TG and OP numbers and the name of the parameter in the first line. That's why I didn't noticed it so far.

probonopd:
No worries! Thanks for the fix. :+1:

PR #69, opened 2022-04-04
Improve compressor and reverb effects

rsta2:
* Can be switched on/off in "Effects" sub-menu in UI
* Compressor is switched on/off for all TGs
* Read effects parameters from performance.ini

Unfortunately I had to change the suggested menus a little bit, because for instance "Edit Reverb" and "Diffusion" does not fit on one LCD line. So there is only one "Reverb" menu, which covers all reverb parameters.

### DIFF
--- a/src/dexedadapter.h
+++ b/src/dexedadapter.h
@@ -33,11 +33,6 @@ public:
 	CDexedAdapter (uint8_t maxnotes, int rate)
 	: Dexed (maxnotes, rate)
 	{
-		Dexed::setCompressor(true);
-		if(Dexed::getCompressor()==true)
-			printf("Dexed-Compressor: enabled\n");
-		else
-			printf("Dexed-Compressor: disabled\n");
 	}
 
 	void loadVoiceParameters (uint8_t* data)
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -113,8 +113,11 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	}
 #endif
 
+	SetParameter (ParameterCompressorEnable, 1);
+
 	// BEGIN setup reverb
 	reverb = new AudioEffectPlateReverb(pConfig->GetSampleRate());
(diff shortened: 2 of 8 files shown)


probonopd:
It's probably better this way. Less deep menus = better.

probonopd:
This is interesting. In the default performance switching the compressor off makes the sound feel _louder_. I would have expected the opposite?

probonopd:
Would you have anything against renaming "Compress" to "Compressor" and "Compress Effects" to "Compressor"? I think that way it would be more consistent with the rest.

rsta2:
Yes, it's getting louder, when the compressor is off, because the audio signal is not compressed any more. ;) The compressor reduces the peak volume to prevent clipping.

"Compress" and "Effects" are two different items. "Effects" is the menu name and is automatically displayed in the upper right corner. "Compress" is the parameter name and is displayed in the left corner. This cannot be changed easily. We could rename "Effects" to "FX" for instance, then there would be enough room for "Compressor".

But the several voice parameters do have also have short names, which are not that nice, thought. I think, we have to live with this (having only 16 character positions). An other possibility would be like that:

```
+----------------+
|         Effects|
|Compressor=On   |
+----------------+
```

This is more like the original DX7 is doing it and was my first try. But it would not work with the volume and pan bars and there is also only small room for the parameter values then.

probonopd:
> Yes, it's getting louder, when the compressor is off, because the audio signal is not compressed any more. ;) The compressor reduces the peak volume to prevent clipping.

Are you describing a limiter here? So far, I understood compressors differently: they compress the sound so that the resulting perceived volume is higher despite the peaks not being higher. And indeed, e.g., the _KORG opsix native_ VST has a compressor enabled by default. When one _disables_ the compressor, then the sound feels "less loud".

Maybe something needs to be parametrized differently?

probonopd:
I am merging this for now, as it is a big improvement to have these in the menu. Have opened https://github.com/probonopd/MiniDexed/issues/71 to keep track of the remaining optimizations.

PR #70, opened 2022-04-04
Add shortcut for changing the TG in parameter edit

rsta2:
When a parameter is edited in the UI, the current TG can be changed
by pressing the switch and turning the knob left or right. The
selected TG remains active, when the parameter editor is left. The
menu home position is entered by triple click now, reboot after
holding the switch for ten seconds.

### DIFF
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -37,6 +37,7 @@ const CUIMenu::TMenuItem CUIMenu::s_MenuRoot[] =
 	{0}
 };
 
+// inserting menu items before "TG1" affect TGShortcutHandler()
 const CUIMenu::TMenuItem CUIMenu::s_MainMenu[] =
 {
 	{"TG1",		MenuHandler,	s_TGMenu, 0},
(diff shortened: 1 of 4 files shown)


probonopd:
Thanks @rsta2, testing it now.

> The menu home position is entered by triple click now

Why did you change this? I think a triple click is much harder to do thatn a long press. Technical reasons or personal preference?

rsta2:
Because pressing the switch while turning the knob would trigger the "Home" function too. It cannot technically be decided, if the longer press is for the shortcut or the Home function.

probonopd:
Logic-wise it could be done like this:
For the rotary controller button, Never act directly on "key down" events. Wait for the "key up" event. If the knob was rotated in between "key down" and "key up", then it was not a long-press but a press-and-rotate operation. (Not sure whether this can be translated into code easily.)

rsta2:
It does not act directly on switch down events, but acts, when the switch is down and the knob is turned. Otherwise you wouldn't see an direct reaction. You would have to turn the knob "in the blind", which does not work, because the reaction is not as direct. You cannot count say 2 turns, and will move the TG number by 2 further.

probonopd:
> It does not act directly on switch down events, but acts, when the switch is down and the knob is turned. Otherwise you wouldn't see an direct reaction. 

Sure. But in that case we also know that it was not a "long press".
Maybe I misunderstand the problem.

Why is it possible to detect a 10 second long press (for rebooting) but not a shorter one (for going to the home screen)?

In any case, merging, since it is an improvement.

PR #73, opened 2022-04-07
Signal chain in float32_t

### LINKED ISSUE #71: Revisit final sound processing stages
Continuation from https://github.com/probonopd/MiniDexed/pull/69#issuecomment-1087892398:

> Ah yes, I tried the compressor in the VLC media player, and it's getting louder there, when it is switched on. As I read it, @dcoredump wants to re-work the whole final signal processing stage (use floats instead of int16_t, pan, mixer, compressor, reverb), which makes absolutely sense to me. At the moment, each TG has its own compressor, and as I said earlier, my pan processing + mixer is not ideal. There is much room for improvement in it.

This ticket is a reminder that we revisit the way we set up and use limiter/anti-clipping, compressor, pan processing, and mixer once @dcoredump has re-worked the final signal processing stage. No hurries, just so that it won't be forgotten.

probonopd:
Closes #71


dcoredump:
Now I got upstream merged into my fork and switched the complete audio signal path into type float32_t. MiniDexeds output seems to be louder now (without clipping). The reverb still has a noisy tail, but I think this is a problem of the PWM output. I hope I can build a setup with a PCM5102A soon.

probonopd:
Thanks @dcoredump. Significant improvement. Tested good with HDMI on RPi 2.

rsta2:
`CXXFLAGS += -DARM_MATH_NEON -DHAVE_NEON` in Synth_Dexed.mk has no effect. Use `DEFINE += -DARM_MATH_NEON -DHAVE_NEON` instead to define these macros.

dcoredump:
> `CXXFLAGS += -DARM_MATH_NEON -DHAVE_NEON` in Synth_Dexed.mk has no effect. Use `DEFINE += -DARM_MATH_NEON -DHAVE_NEON` instead to define these macros.

Thanks @rsta2 .

You mentioned this earlier and I forgot to fix this in my fork.

rsta2:
I did some "code review" to learn about the changes. I think `ProcessSound()` without `ARM_ALLOW_MULTI_CORE` for the RPi 1 produces Stereo audio now, but is initialized for Mono [here](https://github.com/probonopd/MiniDexed/blob/main/src/minidexed.cpp#L203). That does not match. It's good that the output is much louder now. For what is the file *src/minidexed.cpp.O*?

PR #84, opened 2022-04-11
uimenu: Add shortcut for changing OP

rsta2:
Allows to quickly change the current OP, when editing OP parameters.


probonopd:
Thank you very much @rsta2.

The MiniDexed menu structure already now starts to feel more logical to me than the original DX7 user interface. Yes, it's still a lot of parameters hidden away in menu structures, but it's all very logical and can actually be navigated rather quickly.

I think the guy who decided that the original DX7 instrument should have only one "data entry" slider would be proud of our 1 rotary encoder only solution :+1:

PR #94, opened 2022-04-16
Enable filter

probonopd:
Continued from https://github.com/probonopd/MiniDexed/pull/93

PR #93, opened 2022-04-16
Enable filter in Synth_Dexed, UI and performance.ini

rsta2:
@dcoredump Requires the following patch in Synth_Dexed for successful build:

```
diff --git a/src/PluginFx.cpp b/src/PluginFx.cpp
index 9e850e1..4e40731 100644
--- a/src/PluginFx.cpp
+++ b/src/PluginFx.cpp
@@ -48,11 +48,11 @@ inline static float tptlpupw(float & state , float inp , float cutoff , float sr
 //    return (param) * (max - min) + min;
 //}
 
-/*
+#ifdef USE_FX
 static float logsc(float param, const float min, const float max, const float rolloff = 19.0f) {
   return ((EXP_FUNC(param * LOG_FUNC(rolloff + 1)) - 1.0f) / (rolloff)) * (max - min) + min;
 }
-*/
+#endif
 
 PluginFx::PluginFx() {
   Cutoff = 1.0;
```

PR #98, opened 2022-04-17
Enable filter cutoff, resonance, and real-time parameter control

### LINKED ISSUE #80: Real-time parameter control
It would be useful if changes to voice parameters would have immediate effect on already-playing notes.
This would be especially neat in combination with #63.

For some parameters this is already the case (e.g, volume), but apparently not for all of them. More investigation needed.

Do you perhaps know more about this subject @dcoredump?

### LINKED ISSUE #90: Expose Cutoff and Resonance global Dexed paramters in menu
Similar to (De)tune, we should expose the Cutoff and Resonance global Dexed paramters in the menu and save them in `performance.ini`.

image

probonopd:
Closes #90, closes #80

Continued from #93 and #94
Thanks @rsta2 and  @dcoredump


probonopd:
Build succeeded this time! Testing time :+1:

Issue #121, opened 2022-04-20
Support pan, cutoff frequency/resonance and reverb level changes using standard Midi CCs


ChKK1963:
Thanks for already implementing volume, modulation and pitch bend! Next obvious candidates would be pan, cutoff frequency/resonance and reverb level as there are standard Midi CCs defined for them:

- 10 = Pan
- 71 = Resonance
- 74 = Frequency Cutoff
- 91 = Reverb Level

This is independent of the future user-assignable / learnable CCs.

probonopd:
Let's also add * 94 = Detune Level

probonopd:
@ChKK1963 can you please test the version from https://github.com/probonopd/MiniDexed/suites/6193386464/artifacts/218198239 and let me know whether it works as intended, thank you.

ChKK1963:
Hi, just tested your release:

CC10 (Pan), 71 (Resonance) and 74 (Cutoff Frequency) are working fine.

For CC 94 (Detune) I am not sure if this works as intended - this only seems to tune things up. If I have one track running without using the 94 CC and another one with the CC 94 sent, they are only in tune for a CC value 0. Everything other value tunes the track up.

For the 0 value this seems to be in track with the Midi specification which says for CC 94: "0 to 127, with 0 being no celeste effect applied at all.

But isn't detune supposed to be up or down? If you detune two tracks, you will often pitch one down a couple of cents and the other one up accordingly.

Not sure what is a good way to handle this ... I could imagine either using 0 for no detune and 1 - 62 for tuning down 63 in the middle for no detune and 64 to 127 for tuning up. This would allow for "sweeping" from low to high detuning (could be nice with Midi-based LFOs from the sequencer).

Or to let CC 94 as is and use one of the "unused" CCs for tuning things down.

Searched a bit in the Midi specification and the CC charts of a couple of manufacturers but found no "standard" way how this is assumed to behave (except for 0 being "off") ...

Thanks for your work!

probonopd:
Thanks for testing @ChKK1963.

> But isn't detune supposed to be up or down? 

Most definitely! We need to fix this.

> using 0 for no detune and 1 - 62 for tuning down 63 in the middle for no detune and 64 to 127 for tuning up

I was blindly assuming that this was how it was intended, but thinking about it, the DX7 was made at a time when Yamaha (mis)read the MIDI spec for the first time and thought values should go from 0 to 99.

[...]

PR #122, opened 2022-04-20
Support more MIDI CCs

probonopd:
Add support for MIDI CCs

- 10 = Pan
- 71 = Resonance
- 74 = Frequency Cutoff
- 91 = Reverb Send
- 94 = Detune Level

Closes https://github.com/probonopd/MiniDexed/issues/121

### DIFF
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -31,10 +31,15 @@
 #define MIDI_AFTERTOUCH		0b1010			// TODO
 #define MIDI_CONTROL_CHANGE	0b1011
 	#define MIDI_CC_BANK_SELECT_MSB		0	// TODO
-	#define MIDI_CC_MODULATION		1
-	#define MIDI_CC_VOLUME			7
+	#define MIDI_CC_MODULATION			1
+	#define MIDI_CC_VOLUME				7
+	#define MIDI_CC_PAN_POSITION		10
 	#define MIDI_CC_BANK_SELECT_LSB		32
 	#define MIDI_CC_BANK_SUSTAIN		64
+	#define MIDI_CC_RESONANCE			71
+	#define MIDI_CC_FREQUENCY_CUTOFF	74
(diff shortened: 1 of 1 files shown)


probonopd:
I think I was able to implement them except for 91 = Reverb Level, because there is no `m_pSynthesizer->SetReverbLevel` like for the other parameters. Probably we have to implement it in `minidexed.cpp`. Since this is a global (non-TG) parameter I was not sure how to do this.

rsta2:
Regarding 91 (Reverb Level) I'm honestly not sure. Perhaps it's better to use `TGParameterReverbSend` instead, because it is local to the TG, and let Reverb Level only be modified from the UI. You can change `TGParameterReverbSend` from *mididevice.cpp* like that:

```
m_pSynthesizer->SetTGParameter (CMiniDexed::TGParameterReverbSend, nValue,  nTG);
```

Some parameters are not in the MIDI range 0..127. They must be mapped to that range.

probonopd:
> Perhaps it's better to use TGParameterReverbSend instead, because it is local to the TG

Yes, it'd be more consistent with the other params indeed.

> Some parameters are not in the MIDI range 0..127. They must be mapped to that range.

https://github.com/probonopd/MiniDexed/blob/43d1b3ab939213ea49e7b19cb6d35ad6ca9eb09a/src/minidexed.cpp#L386

should take care of that, right?

rsta2:
`constrain()` only checks, if the parameter is in the given range, but does not map it. To map from one range (MIDI) to another use:

```
nReverbSend = (unsigned) maplong ((long) nReverbSend, 0, 127, 0, 99);
```

probonopd:
10 = Pan: **0-127**
71 = Resonance: **0-99**
74 = Frequency Cutoff: **0-99**
91 = Reverb Send: **0-99**
94 = Detune Level: **-99-99**

PR #129, opened 2022-04-21
Add MIDI through feature

rsta2:
* Option "MIDIThrough=from,to" in minidexed.ini to enable
* This forwards MIDI events from the device "from" to "to"
* from/to can be:
	* umidiN (USB MIDI keyboard, N = 1..4)
	* ttyS1 (serial interface)
	* ukbd1 (PC keyboard)
* See src/minidexed.ini for an example


probonopd:
Works perfectly with Synthesia on Android! Thank you very much @rsta2

PR #133, opened 2022-04-21
Patch in WM8960 support

probonopd:
Temporary patch to add WM8960 support until https://github.com/rsta2/circle/pull/286 is integrated into Circle

Thanks @matemaciek

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
       run: |
            set -ex
            cd circle-stdlib/libs/circle
-           git checkout develop
+           git checkout ae22928 # develop
            cd -
     - name: Install toolchains
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,13 +20,11 @@ jobs:
       run: |
            set -ex
            git submodule update --init --recursive
-    - name: Patch in WM8960 support
+    - name: Patch in WM8960 support until Circle PR #286 is merged
       run: |
            set -ex
-           # As long as https://github.com/rsta2/circle/pull/286 is not merged yet
            cd circle-stdlib/libs/circle
            patch -p1 < ../../../src/patches/WM8960.diff
-           rm *.diff
            find . -type f -name "*.orig" -delete
            cd -
     - name: Install toolchains
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,16 @@ jobs:
       run: |
            set -ex
            git submodule update --init --recursive
+    - name: Patch in WM8960 support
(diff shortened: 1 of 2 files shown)


probonopd:
Tested with MiniDexed and a WM8960 DAC, works for me! Both the headphone jack on the WM8960 board and also the speakers connected to the built in amplifier +1

Here is the build:
https://github.com/probonopd/MiniDexed/suites/6210021520/artifacts/219385493

Now it would be good if someone with a PCM5102A or PCM5122 based DAC could also test this in order to ensure that it doesn't break those.

probonopd:
@matemaciek I am mirroring and locally applying your patch in MiniDexed for now so that we can enjoy our WM8960s while an architecturally cleaner solution is being worked out in Circle.

ChKK1963:
> It would be good if someone with access to a PCM5122 DAC could test the build from this branch (linked above) to ensure this doesn't break the PCM5122 which also uses i2c

Just tested the build with my new Raspiaudio+ V2 DAC which has a PCM5102A DAC. Seems to work fine.

probonopd:
Thanks for testing @ChKK1963. We still need to find someone with a PCM5122 DAC though, as that one has i2c while the PCM5102A doesn't.

matemaciek:
> Also, since this is auto-detecting i2c addresses, should we remove the related code from MiniDexed?
Is there any? I only see passing of optional I2C address, and it still makes sense - if user provides specific address, auto-detecting is skipped.

rsta2:
I tried a PiFi DAC+ v2.0 HAT (uses a PCM5122 DAC) attached to a RPi Zero 2 W using this build:

https://github.com/probonopd/MiniDexed/suites/6210021520/artifacts/219385493

First it initialized OK, but after some seconds the sound was strange and after a restart, the I2C initialization failed (-2 from all probed addresses). I aborted the test, because I don't want to destroy something.

With a kernel image, built from the *main* branch, everything works OK.

matemaciek:
I've ordered PiFi, I'll look into it. Previous version was just assuming PCM5122 is present and was blindly sending its init sequence on its two known addresses. Can you try setting `DACI2CAddress=76` and if that doesn't work `DACI2CAddress=77` in `minidexed.ini`?

rsta2:
@matemaciek I guess the *i2cdetect* test, you have implemented, does not work reliable with some devices. I will wait for the license information and then merge your PR, maybe with some modifications. Then I will test again.

matemaciek:
@rsta2 They've just added GPLv3 (-:

probonopd:
WM8960 works. If PCM5xxx also still works, we might want to consider merging.

probonopd:
PCM5xxx confirmed, hence merging.

PR #145, opened 2022-04-22
Blink on MIDI data

probonopd:
For now, using the built-in LED. If this works with no bad side effects, we could also do the same on a GPIO

https://github.com/probonopd/MiniDexed/issues/143

### DIFF
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -82,7 +82,7 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 {
 	assert (m_pSynthesizer != 0);
 	
-	CActLED.Blink (1, 100, 100);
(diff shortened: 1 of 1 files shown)


probonopd:
Maybe we'd need a LED task? Maybe it'd interfere with sound quality?

https://github.com/rsta2/circle/tree/master/sample/19-tasks

probonopd:
Here it is described how it needs to be done:
https://github.com/probonopd/MiniDexed/issues/143#issuecomment-1107425050

probonopd:
See https://github.com/dwhinham/mt32-pi/blob/ae0e146141255425c117a1239b8dcc439358747a/src/mt32pi.cpp for an example.

probonopd:
Exceeds my C++ skills for the moment. If anyone wants to implement it, a PR would be welcome.

PR #155, opened 2022-04-23
Simplified click behaviour

fp64lib:
The current implementation for the menu navigation is following a "GUI" approach with a mouse with Double Click for selecting an item & Single Click for upwards navigation. With just a rotary encoder, especially changing Operator behaviour needs a lot of clicking. I tried to optimize the behaviour towards less clicks with the following rules:

Single click: "Confirm", i.e.
- if the currently selected item is a menu, it selects the menu and goes "down" into the menu
- if the currently selcted item is a parameter (e.g. "Volume"), it confirms the current value and returns the the menu level above
Double click: "Up", i.e. goto into the previous menu
Triple click: "Home", i.e. goto the root menu (no change here)

Here a typical use case on how this reduces the number of necessary clicks:  Select a new voice and adjust "Volume"  & "Pan" for it, e.g. for TG1
- previously: DblClick on "TG1", select "voice", DblClick, select a voice, Click for Up, select "volume", DblClick, Change Volume, Click for Up, select "Pan", DblClick, Select Pan --> 10 clicks
- now: Click on "TG1", select "voice", Click, select a voice, Click for confirm, select "Volume", Click, change Volume, Click for confirm, select "Pan", Click "Pan", Select Pan --> 6 clicks

As user interface behaviour is sometimes very tied to the user and the inventor, I will not fight for this pull. This change works for me very good, I have the feeling that I am much faster at tweeking paramenters. But before rejecting it, give a try, start a jam session for 1 hour, where you experiment with the voices and the parameter settings, tweaking operators etc. 

Otherwise: awesome project, keep developing it. With that, I do not have a dedicated hardware "DX7" in my setup!

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -67,6 +67,9 @@ void CConfig::Load (void)
 			}
 		}
 	}
+	
+	m_bMIDIRXProgramChange = m_Properties.GetNumber ("MIDIRXProgramChange", 1) != 0;
+
 
 	m_bLCDEnabled = m_Properties.GetNumber ("LCDEnabled", 0) != 0;
 	m_nLCDPinEnable = m_Properties.GetNumber ("LCDPinEnable", 17);
@@ -124,6 +127,11 @@ const char *CConfig::GetMIDIThruIn (void) const
 const char *CConfig::GetMIDIThruOut (void) const
 {
 	return m_MIDIThruOut.c_str ();
+}
+
+bool CConfig::GetMIDIRXProgramChange (void) const
+{
+	return m_bMIDIRXProgramChange;
 }
 
 bool CConfig::GetLCDEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -84,6 +84,7 @@ void CConfig::Load (void)
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 5);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 6);
 	m_nEncoderPinSwitch = m_Properties.GetNumber ("EncoderPinSwitch", 26);
+	m_bEncoderClickIsConfirm = m_Properties.GetNumber ("EncoderClickIsConfirm", 0) != 0;
 
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
@@ -192,6 +193,11 @@ unsigned CConfig::GetEncoderPinData (void) const
 unsigned CConfig::GetEncoderPinSwitch (void) const
 {
 	return m_nEncoderPinSwitch;
+}
+
+bool CConfig::GetEncoderClickIsConfirm (void) const
+{
+	return m_bEncoderClickIsConfirm;
 }
 
 bool CConfig::GetMIDIDumpEnabled (void) const
--- a/src/config.h
+++ b/src/config.h
@@ -73,6 +73,8 @@ public:
 	unsigned GetMIDIBaudRate (void) const;
 	const char *GetMIDIThruIn (void) const;	// "" if not specified
 	const char *GetMIDIThruOut (void) const;	// "" if not specified
+	bool GetMIDIRXProgramChange (void) const;	// true if not specified
+	
 
 	// HD44780 LCD
 	// GPIO pin numbers are chip numbers, not header positions
@@ -108,6 +110,7 @@ private:
 	unsigned m_nMIDIBaudRate;
 	std::string m_MIDIThruIn;
 	std::string m_MIDIThruOut;
+	bool m_bMIDIRXProgramChange;
 
 	bool m_bLCDEnabled;
 	unsigned m_nLCDPinEnable;
--- a/src/config.h
+++ b/src/config.h
@@ -93,6 +93,7 @@ public:
 	unsigned GetEncoderPinClock (void) const;
 	unsigned GetEncoderPinData (void) const;
 	unsigned GetEncoderPinSwitch (void) const;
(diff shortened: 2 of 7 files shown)


probonopd:
Hello @fp64lib, welcome to the project. I will test this for a while and see if it makes one's life indeed easier.

craigyjp:
Yeah, keep backing out of menus by mistake, much better implementation.

probonopd:
For those who want to test:
https://github.com/probonopd/MiniDexed/suites/6232936519/artifacts/221146568

probonopd:
The logic problem arises when you can select something (go "down" in the menu) __and__ can go out of something ("up" in the menu). Then you have to double-click. Isn't that hard to explain and understand?

fp64lib:
> The logic problem arises when you can select something (go "down" in the menu) **and** can go out of something ("up" in the menu). Then you have to double-click. Isn't that hard to explain and understand?

No, the default is to select, i.e. to go down. Single Click will always go down, DblClick will always go up. So fo the "middle" layers, where you can go  up and down, Click will always go down. This is basically just swapping the behaviour of Click vs. DblClick.

Only, when you can no longer go down, Single Click will confirm and automatically go up, thus avoiding the additional DoubleClick needed to escape the lowest level.

probonopd:
I am hesitant. While I see how this could speed things up, it also makes things illogical to explain and may be confusing.

miotislucifugis:
For what its worth,my vote is for the simplified click behavior.   Ive been using this version the past week and definitely prefer a single click vs double click to advance.   For me( and i suspect many others,) a single click is way more logical and intuitive.  I hope that this gets merged.... or at least a option to choose.

fp64lib:
> ... or at least a option to choose.
I made it now configurable. If the parameter "EncoderClickIsConfirm=0" in minidexed.ini or missing, it will stay with the default behaviour (DBLClick is down, Click is Up). When the parameter "EncoderClickIsConfirm=1" or any other value, then it will use the bahaviour described above. So everyone can be happy and it will work seemlessly with old minidexed.ini files.

miotislucifugis:
Nice!

probonopd:
Available for testing:
https://github.com/probonopd/MiniDexed/suites/6332305248/artifacts/227858855

probonopd:
No ill side effects in my testing.

So the only downside is that the code gets a bit more complex, and that it becomes more complex to explain and document how things work.

How would we describe this feature in the documentation so that it is short and understandable?

miotislucifugis:
maybe something like this:

Default menu navigation is 2 click to enter/go down , 1 click to return/go up.   
This is defined by "EncoderClickIsConfirm=0" in minidexed.ini

Alternately, one can change the navigation behavior to 1 click to enter/go down and 2 clicks to return/go up  by setting "EncoderClickIsConfirm=1" in minidexed.ini

fp64lib:
I would describe it as follows:

If you do find it difficult to navigate using the default "double-click for down/click for up" behaviour, try setting "EncoderClickIsConfirm=1" in the config file minidexed.ini. This will swap the behaviour of click and double-click: click will now navigate you "down" in the menu tree, double click will navigate you "up". 

Furthermore, the option EncoderClickIsConfirm=1 will also help you setting several parameters with less clicks. When you are at a parameter, there is no more submenu and both double-click and click will bring you back to the previous menu. So, it's easier to memorize that single click confirms your selection, i.e. you either open up a submenu or you confirm the current/modified value of the parameter.

Small example for changing the voice, volume and pan of TG1: 
1. Triple-click to return to the top menu, "TG1" will appear. Click the knob to open the submenu for that tone generator. "Voice" will appear as submenu. Click again to open the submenu. Now turn the encoder knob clockwise or anti-clockwise until the voice you want to use is shown. 
2. Click to confirm the voice. This will confirm your parameter and bring you back to "Voice". Turn the encoder know until "Volume" is shown, click to select, adjust the volume by turning the encoder e.g. to 75.
3. Click to confirm the volume ("Volume" will reappear), turn the encoder until "Pan" is shown, click to select it, and finally adjust panning.

rcarmo:
I _really_ like this behavior. I find double-clicking to be completely non-intuitive for entering menus, but very suitable for backing out of them (and I recall similar behavior from watches and clocks).

Could we have this merged, or turned into an option?

miotislucifugis:
@fp64lib when you get a chance, please update this branch and hopefully it can get merged!  I too am eager for single click

probonopd:
Can we have two options in the `minidexed.ini` file for this?
* One option to just reverse double/single click (this would also be useful in case we implement using an additional button as the "back" button - users would never have to double click)
* Another option to also enable the shortcut ("both double-click and click will bring you back to the previous menu")

probonopd:
Somewhere else I saw click to enter menus and long-press to exit menus.
Wouldn't that be easier, wdyt?

probonopd:
@fp64lib thanks for this PR. In the meantime, additional requests for configurability have led to #274 which should allow users to configure the buttons quite flexibly. Please give it a test; I think we might close this PR in favor if #274 which gives users even greater flexibility in configuring buttons. With the following configuration you get single-click to enter/select and long-press to cancel/back:

```
ButtonPinPrev=0
ButtonActionPrev=
ButtonPinNext=0
ButtonActionNext=
ButtonPinBack=11
ButtonActionBack=longpress
ButtonPinSelect=11
ButtonActionSelect=click
ButtonPinHome=0
ButtonActionHome=

# KY-040 Rotary Encoder
EncoderEnabled=1
EncoderPinClock=10
EncoderPinData=9
```

But it is also possible to use, e.g., an extra button for cancel/back.

fp64lib:
@probonopd Thanks for the update. I'm currently on holidays without any chance testing it. I will do so after returning home in about 2 weeks.

probonopd:
Would like to close this in favor of https://github.com/probonopd/MiniDexed/pull/274 if no one objects. Thoughts?

PR #206, opened 2022-05-05
Support for Portamento and Pitch Bender parameters

arsamus:
Hi everybody, I bring you my proposal to include control over the Portamento and Pitch bender parameters. This enhancement is necessary in order to play more expressive solos. Musicians will be grateful.


probonopd:
Hello @arsamus. Welcome to the project and thanks for this very welcome contribution!

A build is available here:
https://github.com/probonopd/MiniDexed/suites/6393336169/artifacts/232528437

Please test and let me know whether it works as intended, at which point I'm happy to merge it.

arsamus:
Hi @probonopd everything works as expected!! 

Yes, of course, I will send you a draft to add on wiki.

miotislucifugis:
Ok- I found a bug:  
When changing portamento settings, it always show and gets gets applied to TG1, regardless of which TG was selected. 

to repeat:  Select a TG other than TG1, scroll and double click "portamento",  scroll to and double click "Mode, "Glissando" or "Time".   It always say TG1.  the Pressing and scrolling "shortcut" does not change  TG1  to another TG.   The shortcut TG change is registered by MiniDexed, because when I exit Portamento options, the TG is different than the one originally selected.

arsamus:
@miotislucifugis thank you for testing, i've just modify it with your comments: 1-replace numbers 0/1 with a description similar to DX7 menu, 2-Correct the bug.

@probonopd here are the draft for the Wiki extrated from DX7 manual: 
Portamento: When you play two consecutive notes, the second takes on its pitch gradually, equivalent to sliding your finger across a violin string joining two notes. This controls can be used to set portamento (glide) and glissando (stepped glide) effects.

	Mode: there are two modes, the first "Fingered" produces the portamento only when the second note is pressed before releasing the first (legatto); or "Full time", the portamento is always produced. (Note: This is not correctly implemented in synth_dexed yet, it seems that there is a bug on the code but it is left activated in advance)

	Glissando: When the Glissando feature is turned "ON", the glide in pitch occurs in discrete steps. This effect is best heard with a slower rate and when two widely separated notes are played one after the other. When Glissando is turned "OFF" then normal Portamento (continuous glide) is available.

	Time: This sets the time of the Portamento effect. A setting of 0 produces no effect, while a setting of 99 produces the longest (slowest) pitch changes.

Pitch Bender: sets what is happens when 
	Range: This lets you program the maximum amount of pitch bend from no effect (0) to plus or minus an octave (12 semitones). The numbers in the programming display indicate how many semitones the wheel can change the pitch in each direction. 
	
	Steps: This allows you to set the size of the increments by which the wheel changes the pitch. A setting of zero TIME (0) will cause perfectly smooth pitch bending. Each This sets the time of the Portamento or Glissando effect. A setting of 0 produces no effect, while a setting of 99 produces the longest (slowest) pitch unit above zero represents the number of semitones by which the pitch will "jump" as you move the wheel.

probonopd:
Thanks @arsamus. Works well for me. Are the defaults you chose the same defaults as on the DX7?

Also, could you move the new options after "Channel"? (I think the MIDI channel is changed more frequently than the new options, at least according to my experience.)

miotislucifugis:
Personally I think that its better and more logical to leave them at their current place in the menu, along with the other sound parameters.   Leaving midi Channel last makes it easier to find because I can quickly scroll to the end of the list without having to search around.

probonopd:
OK, then let's not move them.

arsamus:
Hi @probonopd my contribution is just a grain of sand!!! This is a nice project!!
Let me find on the internet about default values. I don't have a near DX7 to check it, (last time a play one was in early nineties..) but general rule on most synths is that portamento is set off (glissando off, and time set on 0) and pitch bender has a range of 2 semitones and with no discrete steps (step set on 0).

probonopd:
This is what the [DX7 Operating Manual](https://de.yamaha.com/files/download/other_assets/5/1197065/dx7e.zip) says:

Regarding Portamento

* "FINGERED PORTA." In this mode portamento is applied only to legato notes
* "FULL TIME PORTA." In this mode portamento is always applied
* The glissando function is turned either ON or OFF. When it is OFF a normal portamento effect is produced
* Adjusts the speed of the portamento/glissando effect from 0 to 99. A0 setting results in no effect, while a setting of 99 produces the longest (slowest) portamento or glissando.

Regarding Pitch Bend
* The range of pitch bend can be set from 0 to 12. 0 range is equivalent to no pitch bend. A setting of 12 permits pitch bend over a plus minus 1200 cent (2 octave) range, If the range is set at 7, then pitch bend will be possible over a plus minus 700 cent range (i.e. plus or minus one fifth).
* The step parameter can be set from 0 to 12. A setting of 0 corresponds to 0-cent steps, and a setting of 12 corresponds to 1200-cent (1 octave) steps. If STEP is set to 0, then a perfectly smooth pitch hand will result. IF STEP is set to 1, the pitch will bend in 100-cent (semitone) steps. 
* Pitch bend will not function if RANGE is set to 0.

miotislucifugis:
Is there a build of this available to test?

probonopd:
Yes. Each time you see a green checkmark there is an updated build to test.
To access it:
* Click on the green checkmark
* Click on "details"
* Click on "Summary"
* At the bottom of the page, under "Artifacts", there is the build

miotislucifugis:
Oh sorry, I was there but didnt scroll down to see "artifacts"- thanks!

probonopd:
It's all a bit hidden unfortunately.

miotislucifugis:
if i understand this correctly, "fingered" mode is not working properly in synth_dexed, right?   right now both it and "full-time" are producing the same effect.    is that the expected behavior?   
Other than that, everything else seems to be working well.   Ill have more time to glide and bend tomorrow to double check, but I think that its good.      Great work!

arsamus:
@miotislucifugis yes both are the same now because sinth_dexed don't handle it in the right way.

probonopd:
@arsamus can you describe in more detail what @dcoredump would need to change in Synth_Dexed for it to be able to work?

arsamus:
Look at the code below. Portamento_mode has a constrain that limit the value between 0 and 1, then controllers.​portamento_enable_cc can be true only when portamento_mode is greater than 63... Then some lines above can be true when time parameter (there are a mistake in the constrain of this variable) is greater than 0.

I don't thing the portamento_enable_cc is the same as portamento_mode.

Something seems not good here.

 ​void​ ​Dexed::setPortamentoMode​(​uint8_t​ portamento_mode, ​uint8_t​ portamento_glissando, ​uint8_t​ portamento_time) 
 ​{ 
 ​  portamento_mode = ​constrain​(portamento_mode, ​0​, ​1​); 
 ​  portamento_glissando = ​constrain​(portamento_glissando, ​0​, ​1​); 
 ​  portamento_mode = ​constrain​(portamento_mode, ​0​, ​99​); 
  
 ​  controllers.​portamento_cc​ = portamento_time; 
 ​  controllers.​portamento_enable_cc​ = portamento_mode > ​63​; 
  
 ​  ​if​ (portamento_time > ​0​) 
 ​    controllers.​portamento_enable_cc​ = ​true​; 
 ​  ​else 
 ​    controllers.​portamento_enable_cc​ = ​false​; 
  
 ​  controllers.​values_​[​kControllerPortamentoGlissando​] = portamento_glissando; 
  
 ​  controllers.​refresh​(); 
 ​}

dcoredump:
> @arsamus can you describe in more detail what would need to be changed in Synth_Dexed for it to be able to work? @dcoredump does it work in MicroDexed?

Yes, the portamento code has problems. But currently I am diving deep into the **missing SYSEX 0xf7** problem, so I can't fix Synth_Dexed. Best way would be to get a PR in codeberg. Would this be feadible?

miotislucifugis:
I did some more testing and everything seems to be working.   Pitchbend Steps is a little strange, everything over 6 steps causes the bend to jump to the highest note, regardless of how many notes are set in Range... but that may be the "normal" behavior.   I dont have a real DX7 anymore and when I had one, i dont think that is a function that I ever used.   i actually work at a synth museum and we have several Dx7s...ill  pull one out tomorrow and test to see if Minidexed is working the same.
One suggestion: the original DX7 had a dedicated portamento jack that allowed for activating/deactivating portamento with a footswitch.  This allows for quick and immediate switching on the fly w/o having to adjust the time level... which could be left at a desired setting.   Would it be possible to add an on/off feature possibly controlled by a midi CC command?  is this possible in miniDexed or would this have to be added to synth_dexed first?

arsamus:
As I've red on DX7 manual if pitch bender "step" is set on a value different than zero, then the range setting is ignored, maybe Dexed do the same. But if you access to a real one to check it, that would be better.

The portamento on/off is implemented as I've saw on the code so it'd have been controlled via MIDI (I don't test it), there are a controller (cc#65) that that do that. The code synth_Dexed needs some adjustments but it seems that everything was implemented. 

"Mono" mode also would seem to be implemented, there are a method to enable/disable it.

probonopd:
@arsamus it looks like the new sysex handling code that I merged recently makes some changes necessary. Could you please have a look? Thank you.

arsamus:
@probonopd yes the error was due to new version of dexed, I've just update the code. 

by the way, I've download the last build (with SysEx handlings) and something not works ok with midi, some note events are not received properly. I've saw the code for handling SysEx and there are somethig strange on it. You would consider the following code instead for the "for" loop. I've test it and handles Ok sysex, running status and rest of events: 
`	for (int i = 0; i < nResult; i++)
	{
		u8 uchData = Buffer[i];
		
		switch (m_nSerialState)
		{
		case 0:
		MIDIRestart:
			if (   (uchData & 0x80) == 0x80		// status byte, all channels
			    && (uchData & 0xF0) != 0xF0)	// ignore system messages
			{
				m_SerialMessage[m_nSerialState++] = uchData;
			}
			
			if (uchData == 0xF0) // SysEx status Byte jumps to m_nSerialState=5 and iniciate reading
			{
				m_nSerialState = 5;
				m_nSysEx = 0; 
				goto MIDISysEx;
			}
			
			break;

		case 1:
		case 2:
		DATABytes:
			if (uchData & 0x80)			// got status when parameter expected
			{
				m_nSerialState = 0;

				goto MIDIRestart;
			}

			m_SerialMessage[m_nSerialState++] = uchData;

			if (   (m_SerialMessage[0] & 0xE0) == 0xC0
			    || m_nSerialState == 3)		// message is complete
			{
				MIDIMessageHandler (m_SerialMessage, m_nSerialState);

				m_nSerialState = 4; // State 4 for test if 4th byte is a status byte or a data byte 
			}
			break;
		case 4: // Running Status evaluation
			
			if ((uchData & 0x80) == 0)  // true data byte, false status byte
			{
				m_nSerialState = 1; // Byte 0 not change on Running Status
				goto DATABytes;
			}
			else 
			{
				m_nSerialState = 0;
				goto MIDIRestart;  // This is necessary in order to not miss the first byte
				
			}
			break;
		case 5: // SyxEx reading
			MIDISysEx:
			m_SerialMessage[m_nSysEx++] = uchData;
			if (((uchData & 0x80) && m_nSysEx > 1 ) || m_nSysEx >= MAX_MIDI_MESSAGE)		
			{
				m_nSerialState = 0; //New Status byte ends SerialState 5 (SysEx reading)
				if (uchData == 0xF7)
				{
					MIDIMessageHandler (m_SerialMessage, m_nSysEx);
				}
				else
				{
					goto MIDIRestart; //other status byte abort SysEx process and jump to MIDIRestart in order to not miss the byte
				}
			}
			break;
		default:
			assert (0);
			break;
		}
	}`

probonopd:
Hi @arsamus, thanks for looking into it, could you please open another PR regarding the MIDI handling for loop? Thank you very much!

arsamus:
Hi @probonopd I've update the "for" loop on serialmididevice.cpp to handle properly Serial MIDI messages (SysEx, Running Status, etc. )

arsamus:
Everything works as expected related to this features.

It seems that the sound quality change a bit since the last upgrade of synth_dexed. The new versión is not so "cristal" and "pure", it sounds a little detuned (or chorused).

probonopd:
What surprises me a bit:
Are Pitch Bend and Portamento settings really performance setting?
Aren't they part of voices and voice .syx files?

craigyjp:
On the DX7 MK1 there was ony one performance for all 32/64 sounds. 
On the MK2 they had 32/64 performances selectable independent of the voice.

I don't know about later machines like the TX802, but it makes sense that each voice should have its own settings rather than a global for all voices.

miotislucifugis:
I check with one of our DX7s today and there is definitely a problem with the behavior of pitch bend steps.  Here is what _should_ be happing:
bend steps 12 = one step of 12 semitones
bend steps 11 = one step of 11 semitones
bend steps 10 = one step of 10 semitones
bend steps 9 = one step of 9 semitones
bend steps 8 = one step of 8 semitones
bend steps 7 = one step of 7 semitones
bend steps 6 = 2 steps, 6 &12 semitones 
bend steps 5 = 2 steps, 5 & 10 semitones 
bend steps 4 = 3 steps, 4, 8 &12 semitones
bend steps 3 = 4 steps  3,6,9,12 semitones
bend steps 2 = 6 steps 2,4,6,8,10,12 semitones
bend steps 1 = 12 steps, each one semitone

in miniDexed bend steps 1-4 , 6 and 12  are working properly, but bend steps 5 results in 2 steps, 6 &12 semitones ; and bend steps 7-11 are always one step at 12 semitones.

Im not certian, but I suspect that this problem lies in synth_Dexed, but i have no way of checking independent on MiniDexed.  Has anyone here built a microdexed to confirm?

miotislucifugis:
oh, one more thought...   "Pitch Bender" should be changed to just simply "Pitch Bend" to keep it consistent with both DX7 and general synth terminology.   No biggie tho'.

craigyjp:
Can you explain the last 6 steps in this configuration as they make no sense. If it is set to 1 I would expect a max pitchbend of 1 semitone, 2 would be a whole tone etc.

miotislucifugis:
That would be logical, but no how it works.  its always a number of steps evenly divisible by 12.   
here is an excerpt from the DX7 manual:
"(As soon as
you select a [STEP] size other than zero, the
[RANGE] is automatically reset to 12 semitones.)"
"Go back and experiment with different step sizes by
pressing [PITCH BEND-STEP] and moving the
slider and/or pressing the [+1] and [-1] buttons. The
bend was continuous with the step size at 0. A step
size of 1 enables you to hear each semitone as you
move the wheel, and, at the other extreme, a step
size of 12 means you'll hear the note or chord "jump"
a full octave.
Any step size above 0 sets the range to 12, but spe-
cific step sizes affects the maximum available range; it
may not always be as much as plus or minus an
octave. For example, if you choose a step size of 5
steps, and move the PITCH wheel slightly you'll get a
jump up or down of 5 semitones. Full deflection of
the wheel from its resting position will cause the pitch
to jump up or down a total of 10 semitones... and
that's it. A range of 6 steps, however, would allow for
an octave of pitch "bend" in 2 increments"

craigyjp:
The manual states if range is set to 0 then pitchbend will not function regardless of the step settings.

probonopd:
> On later synths (early nineties Roland JV80 for example), portamento and pitch bender settings are part of the voice structure, but that would not be the case on DX7.

For me personally it would be _logical_ if things like Pitch Bend Step, Pitch Bend Mode, and Portamento Step were voice parameters, would be inside the "Edit Voice" submenu, and they would not be saved in voice `.syx`. But indeed this is not how it works on the DX7. 

https://homepages.abdn.ac.uk/d.j.benson/pages/dx7/manuals/dx7-man.pdf pp. 15 makes clear that "function parameters" including portamento and pitch bend apply to all voices, and hence are not saved with the individual voices.

So the current implementation is correct.

sou7611:
To be more clear, function parameters are applied from the "controller"
side of things (in DX7 thinking), so it's a keyboard thing, and they apply
across the whole range of the keyboard.

That's why they're universally applied, and not tied specifically to each
voice.

craigyjp:
I think we have to be clear that this is a multitimbral DX7, so a global function page is not really practical. 

It has to be at a voice level and not higher. 

Maybe consider the minidexed akin to the TX802 and look at how that functions for parameters like portamento, pitchbend etc.

sou7611:
it you want to properly "mimic" the TX816-style hardware, it should apply
to each one of the 8 "tone modules" independently, I'd think.  And yes,
this should match your idea of voices, since each one of the 8 tone modules
can support only a single "voice" at a time.

Or, more accurately, any set of "bonded" modules (which can be a single, or
multiple, modules).

Essentially, the easiest way is to set by midi channel.  e.g., anything
"listening" on channel 1, 2, 3, etc.

arsamus:
Current implementation define individual setting for each TG (each like a DX7 module). The thing here is that as it's commented DX7 define this parameters as general parameters (as part of the keyboard setting) but not as a parameter of voice patches, as later synths do. The consequence of this is that no voice SysEx define this parameters, is necesary an specific SysEx message to modify it and apply for every patch in ram or roms. 

Other discusión is where is the best place to insert the menus that define this settings on MiniDexed, on a main TG menú level (level 2), as I defined on this proposal, or inside "Voice Edit" menú. ¿?

dcoredump:
> > On later synths (early nineties Roland JV80 for example), portamento and pitch bender settings are part of the voice structure, but that would not be the case on DX7.
> 
> For me personally it would be _logical_ if things like Pitch Bend Step, Pitch Bend Mode, and Portamento Step were voice parameters, would be inside the "Edit Voice" submenu, and they would not be saved in voice `.syx`. But indeed this is not how it works on the DX7.
> 
> https://homepages.abdn.ac.uk/d.j.benson/pages/dx7/manuals/dx7-man.pdf pp. 15 makes clear that "function parameters" including portamento and pitch bend apply to all voices, and hence are not saved with the individual voices.
> 
> So the current implementation is correct.

Yes, the current implementation is correct but problematic in live setups. For MicroDexed we used another way: The configuration holds a complete setup of all engines (including the functions) and refers only to the bank/voice which should be loaded with _this_ setup. That makes it easier to create live setups.

probonopd:
> Other discusión is where is the best place to insert the menus that define this settings on MiniDexed, on a main TG menú level (level 2), as I defined on this proposal, or inside "Voice Edit" menú. ¿?

I think that only voice parameters should go into "Edit Voice", and since Portamento and Pitch Bend are not voice parameters, they should not go there. 

> The configuration holds a complete setup of all engines (including the functions) and refers only to the bank/voice which should be loaded with this setup. 

Isn't that basically what a "performance" is for? (Yes, some day we want to have ability to have more than one `performance.ini`).

arsamus:
Load and save different performance.ini will be an important enhancement for live performance. May be the first menú item to quickly select the predefined configuration.

Other important feature would be the possibilty to save voice data on performance.ini, with this it would be possible to save edited parameters on "edit voice" or those params modified via SysEx. I've just have implemented it on one of my local cloned repos and works very well. I can share here if you feel confortable with this enhacement. Data is saved as string of space separated two digit hex numbers, so it can be edited easily with a text editor.

Both thing would convert MiniDexed on a near full stand alone instrument.

probonopd:
> Other important feature would be the possibilty to save voice data on performance.ini, with this it would be possible to save edited parameters on "edit voice" or those params modified via SysEx. I've just have implemented it on one of my local cloned repos and works very well. I can share here if you feel confortable with this enhacement. 

Cool! But why would one not save those to a `.syx` file?

arsamus:
> Cool! But why would one not save those to a `.syx` file?
From my point of view, both things are necessary. On one hand a future enhacement would have to allow to save any voice config on a custom "Rom". But, on the other hand, it's good that all parameters are saved on a performance, so, it doesn't mind what is happend with your roms, you can rename or order on another way, but you never lost your performance config. Many Synths do that, on my Roland JV80 you can save, via SysEx, the whole performance memory, including all voice parameter for each part. Performance.ini is like a curstom Minidexed SysEx file but in a human redeable way...

probonopd:
Do we all think this is ready to merge?

miotislucifugis:
Everything in this branch has been working well for me...(Aside from the 2 issues with bend-steps and portamento-mode mentioned above.)   I would like to have a portamento enable - yes/no option that is mapped to a midi CC and also portamento time but those features can be added later.

probonopd:
@arsamus would it be OK for you to change "Pitch Bender" to "Pitch Bend" in GUI and source code? That would make it more consistent with existing literature.

arsamus:
> @arsamus would it be OK for you to change "Pitch Bender" to "Pitch Bend" in GUI and source code? That would make it more consistent with existing literature.

Yes, I will tell you when all commits will be done.

arsamus:
> @arsamus would it be OK for you to change "Pitch Bender" to "Pitch Bend" in GUI and source code? That would make it more consistent with existing literature.


Done!

probonopd:
Excellent! Please test
https://github.com/probonopd/MiniDexed/suites/6507944348/artifacts/241967451
and let me know if it is ready for primetime :+1:

arsamus:
Tested it! I think it is ready.

JoaquinSainz:
I just tested as well nice implementation 🤘

probonopd:
Thank you very much. Looks like this is ready for being merged then.

Would anyone like to write the new entries for
https://github.com/probonopd/MiniDexed/wiki/Menu
(please follow the brevity and style of the other sections, and mention the defaults, min and max for parameters)?

probonopd:
@arsamus I have added brief descriptions to https://github.com/probonopd/MiniDexed/wiki/Menu, please check to ensure that they are correct. Thanks!

probonopd:
> Other important feature would be the possibilty to save voice data on performance.ini, with this it would be possible to save edited parameters on "edit voice" or those params modified via SysEx. I've just have implemented it on one of my local cloned repos and works very well. I can share here if you feel confortable with this enhacement.

It would be great if you could open a pull request @arsamus. Thank you very much.

arsamus:
> 
> It would be great if you could open a pull request @arsamus. Thank you very much.

Ok. I'll prepare a commit for that as soon as I can!

Issue #182, opened 2022-04-27
Internal mixer mixes linear and not logarithmic


dcoredump:
... as the subject says... I think panning is also coded linear (I am not sure if this is right). Should it be also logarithmic?

Will fix this soon for internal mixer class (if noone else wants to do this, because I am currently writing the SYSEX engine).

probonopd:
Great chatch! According to https://sound.stackexchange.com/a/41550,

> The MIDI specification itself does not define the exact response to the volume controller.
> 
> Nowadays, there is almost no device that does no implement the expression controller. Both the GM Level 1 Developer Guidelines and the DLS specification use L(dB) = 40 log (volume × expression / 127²). Furthermore, the guidelines mention that
> 
>> there was general agreement about how these two controllers interacted as well: In 9 cases, their values were combined (multiplied) to get the actual level. The recommended volume response curves for CC#7 (volume) and CC#11 (expression) used herein were provided to the General MIDI Working Group of the IASIG/MMA by Yamaha Corporation. Roland uses the same response curve, and other Japanese manufacturers who are members of the AMEI have agreed to do the same.
> 
> So this is how almost all devices actually implement it.

probonopd:
Merged. Thank you @dcoredump

PR #210, opened 2022-05-10
Fixing linear internal mixer to non-linear (x^4) / Fixing linear pan to non-linear (sin/cos) #182

dcoredump:
As described in #182:
The linear behavior of the internal mixer has now been changed to `x^4` as an approximation. Also, the behavior of the panorama controller has been changed from linear to using a `sin/cos` function.

Both should sound more natural to the human ear.


probonopd:
Thank you very much @dcoredump

Issue #169, opened 2022-04-25
I2C display support


Deftaudio:
Hi, just wondering, what's the current plan on this? I'm willing to further reduce the size of MiniDexed and just limited with parallel LCD options... In particular, OLED 128x32 would be the best.

luisfcorreia:
my current C++ skils are basicly non existing

user Interface should have two things detached, display and encoder handling

I tried, without success

probonopd:
i2c display support would need a volunteer to write it. The proper way would be to get those displays supported in [Circle](https://github.com/rsta2/circle/), e.g., by [upstreaming them from mt32-pi](https://github.com/dwhinham/mt32-pi/blob/63dc284de3f7e4926ea143cdd4bce6a091bc9614/include/config.h#L70). Once in Circle, we could think about using them in this project.

Deftaudio:
I see, thanks!

diyelectromusic:
For info, I've sent in support for the I2C version of the HD44780 so hopefully that will be in the next Circle release.

I have draft support for SSD1306 based displays too based on the driver in MT32-Pi (you can see it in my own branch of circle), but it looks like that is going to open a bit of a wider discussion on how to do graphical displays that will have to wait until after the next release of Circle.  To follow the conversation, see https://github.com/rsta2/circle/discussions/296

But in summary, if I can work out how to build MiniDexed myself with my own version of Circle, I might have a go at building a version using a 128x32 ssd1306, which would be nice... but I'm not sure it would be something releasable more widely at present...

Kevin

probonopd:
Wow, that's quite exciting @diyelectromusic!

> Your HD44780 I2C support is on the develop branch in the Circle repo now 

Since MiniDexed is using the develop branch of Circle, we should be good to go!
How can MiniDexed make use of it?

diyelectromusic:
Well to use the I2C hd44780 just use the same initialisation but with I2C instead.  There is a new addon/display/sample/hd44780i2c example that shows how (I haven't looked at how MiniDexed does it yet - I'll have a look and see).  The SSD1306 in theory would be similar, but as I say that isn't in Circle yet... :)

diyelectromusic:
So I think you need to first create a CI2CMaster object and then pass this into the new HD44780 constructor something like the following:

#define I2C_MASTER_DEVICE	(CMachineInfo::Get ()->GetDevice (DeviceI2CMaster))

[...]

PR #220, opened 2022-05-14
I2C display support

probonopd:
Closes https://github.com/probonopd/MiniDexed/issues/169
Thanks @diyelectromusic

TODO to make this proper:
 
- [ ] In https://github.com/probonopd/MiniDexed/blob/i2c/src/config.h change `LCDI2CAddress` to `GetLCDI2CAddress`, implement reading this value from the ini file in `config.c`
- [ ] In `userinterface.cpp` only use i2c for the display if `LCDDevice=44780i2c` in `minidexed.ini` and non-i2c if `LCDDevice=44780`
- [ ] Have reasonable defaults in the software and in the `minidexed.ini` file

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
       run: |
            set -ex
            cd circle-stdlib/libs/circle
-           git checkout ae22928 # develop
+           git checkout c9a4815 # develop
            cd -
     - name: Install toolchains
       run: |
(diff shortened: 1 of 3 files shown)


probonopd:
As @diyelectromusic said...

> It looks like it will need to be added to the interface for CUserInterface and then passed in from kernel.c when m_UI is setup in CMiniDexed...

Issue #125, opened 2022-04-20
Feature request - switchable performance.ini files


ebrown3k:
I think adding a way to have multiple performance.ini files and swap them through the UI/menu would add a ton of usefulness to this in a live performance setting.

probonopd:
Maybe a structure similar to sysex should be used, along the lines of `performances/001_PerformanceName.ini`, `performances/002_PerformanceName2.ini`

craigyjp:
The original DX7 only had one performance mode unfortunately, but the MK2 and S models had 32 or 64 performance modes. So it would be nice to have the extra performance patches.

arsamus:
> Maybe a structure similar to sysex should be used, along the lines of `performances/001_PerformanceName.ini`, `performances/002_PerformanceName2.ini`

This will be a good enhancement it is necessary for those that use minidexed as a multitumbral synth. Selection of saved performances is the way to change this multitumbral paches, otherwise you have to navigate on the menus and ajust the parameters to set the decired patch. 

Other important enhancement could be load and saving performances and sysex banks on an USB pendrive. this avoid to access to the SD to add new roms or to modify performance files.

probonopd:
Do you think you could do it @arsamus?

arsamus:
> Do you think you could do it @arsamus?

Let me see if I can do something.

PR #228, opened 2022-05-16
Switchable performance files and saving all voice data parameters.

arsamus:
This feature allows to save on Performance.ini all voice parameters edited via UI or SysEx.


probonopd:
Thanks @arsamus. What is supposed to happen if VoiceDataN is not there, e.g., in the default `performance.ini`? Strangely, that sounds different now than without this patch.

arsamus:
> Thanks @arsamus. What is supposed to happen if VoiceDataN is not there, e.g., in the default `performance.ini`? Strangely, that sounds different now than without this patch.

For this example, nothing, just random data is apply to parameters, then if you want to load a former performance.ini file you have to go to voice menu and move the encoder 1 step up and 1 step down then save Performance to fill VoiceDataN. On a final release it is possible to define that don't set anything if the value is not properly formated or missing. 

Sounds different maybe because for the example I paste the same voice parameter for all TG. But nothing would have change in relation to sound.

arsamus:
Hi @probonopd I have an improvement for this feature. If VoiceDataN is not there it will priorice BankNumber and VoiceNumber values, so no random data will be load. I will commit this soon. 

Moreover, I also have a solution for #125 switchable performance files (load, save and create), I can commit all in this PR. Maybe after #238 will be merged in order to get that improvement.

probonopd:
Looking forward to it @arsamus.

arsamus:
> Looking forward to it @arsamus.

Hi @probonopd I've just PR the code for manage switcheable performances and save all voice data parameters on performance, with this patch it's possible to edit all paramenters and saved on the actual performance file or in a new one. On the other hand the new Performance menu allows to switch from diferent saved performance files, very usefull on live performance. To load selected performance it is necessary to double click on it, performance is loaded when [Ld] flag apears on left side of the LCD. 

The enhacement is retrocompatible so, the only thing that it have to be done is create a folder named "performance" on root directory. performance.ini is the default performance file and has to be located on root directory, other performances files must be saved in a "performance" named folder. The format file name must be: XXXXXX_NAME.ini, XXXXXX is the incremental index and Name can't exced 14 characters. Example: 000001_Perf1.ini.

If VoiceData# parameter is not set on the file, voice and bank parameters are loaded.

Future improvements: 
-Edit name of new performance via UI (now is automatically generated and can only be edited via some file manager application on a computer)
-Edit voice name via UI
-read / save performance from / to USB Pendrive.

probonopd:
Thank you very much @arsamus.
Let's test [MiniDexed_2022-05-22-acfb6d4](https://github.com/probonopd/MiniDexed/suites/6607214458/artifacts/248435010).

miotislucifugis:
This is super!  Trying it now.

probonopd:
Works for me. :+1: This is really, really useful.

Couple of questions:

* Wouldn't it be more logical to have one menu entry "Performance" with submenus "Save" and "Load"?
* When loading performances, can we make it so that performances are already loaded when the name is displayed, without having to double-click it? Like it works with Voices?
* Are the menu entries ordered by number? I have

```
performance/000002_4xE.Piano+More.ini
performance/000001_4xE.Piano.ini
```

Would have expected the order in the menu to be: `Default`, `4xE.Piano`, `4xE.Piano+More` but it is `Default`, `4xE.Piano+More`, `4xE.Piano`.

arsamus:
> * Wouldn't it be more logical to have one menu entry "Performance" with submenus "Save" and "Load"?

It could be but from my point of view I think load performance needs to be very easy to get. So first level it could be the best option.  

> * When loading performances, can we make it so that performances are already loaded when the name is displayed, without having to double-click it? Like it works with Voices?

 I've asked myself this question at the begining, but there are two things here. In live performance if you  want to switch from one perf.. to another you would have to pass for all perf.. in between, with de current implementation you activate the new performance just when double click it. The way that Voices works ok for voices and banks because musicians  used it in an "experimental mode" when want to find a proper sound, but in performance, they would need to switch between different configs. The other thing here is that, in order to get a easy viable solution and to avoid to do agressive code interventions, performance is loaded from file at the moment you select it and it is not preloaded like voices and banks. Change this implies re-think Performanceconfig class and it dependents, creating arrays for all variables, etc.

> * If the `performance/` directory does not exist yet, can we create it? Currently it says "Complete" but it doesn't seem to save anything when that directory doesn't exist

Yes, I've saw that, I think this is easy to fix. 

> * Are the menu entries ordered by number? I have
> performance/000002_4xE.Piano+More.ini
> performance/000001_4xE.Piano.ini
> ``
> Would have expected the order in the menu to be: `Default`, `4xE.Piano`, `4xE.Piano+More` but it is `Default`, `4xE.Piano+More`, `4xE.Piano`.

This would have been easy to implement, let me see how to do it.

> * When one tries to save but there is no SD card in the device, it still says "Complete". Can we make it say "Card error" instead? 

Yes, same as 3rd question, this is related to the "saving performance fix" #238 where the function always returns a true value, but I thing it could be avoided in some way.

miotislucifugis:
This seems to be working well.  no problems here

arsamus:
I've just update some minimum changes in order to fix or improve the following things:

- if "performance" directory does not exist, attemp to create it. If it can't do it, saving new performance show "error" message.
- Performance files are listed on menu ordered them by ID number (first six numbers) and not by creation date.

arsamus:
> Before I butcher i

No problem @probonopd, done it! actually there was no real conflict

probonopd:
Let's test: [MiniDexed_2022-05-25-838edab](https://github.com/probonopd/MiniDexed/suites/6647819228/artifacts/251107148)

anycolorsmith:
Wow! It works!

JoaquinSainz:
I just tried this version it works very well, nice incorporation👍

arsamus:
It works as expected!

probonopd:
Thank you very mcuh @arsamus :+1:

Banana71:
MiniDexed cannot handle more than 40 performances

There are 39 Perf.ini files in my performance folder. When I try to save a new performance, the Raspberry Pi crashes.
I've tried a few things with no success. (ID sorted, more than 40 performances in the folder)

[performance org.zip](https://github.com/probonopd/MiniDexed/files/10089372/performance.org.zip)
[performance 68.zip](https://github.com/probonopd/MiniDexed/files/10089371/performance.68.zip)

Make a back up from your data before copying my performance to the SD card.

src/performanceconfig.h
```diff
@@ -113,6 +131,17 @@ class CPerformanceConfig	// Performance configuration
 	unsigned m_nPortamentoMode[CConfig::ToneGenerators];
 	unsigned m_nPortamentoGlissando[CConfig::ToneGenerators];
 	unsigned m_nPortamentoTime[CConfig::ToneGenerators];
+	std::string m_nVoiceDataTxt[CConfig::ToneGenerators]; 
+	
+	unsigned nLastPerformance;  
+	unsigned nLastFileIndex;
+	unsigned nActualPerformance = 0;  
+	unsigned nMenuSelectedPerformance = 0;
+	std::string m_nPerformanceFileName[40];
```
probonopd, line 140: Does this limit the number of performances to 40? https://github.com/probonopd/MiniDexed/issues/379

PR #247, opened 2022-05-22
Several SYSEX fixes, added master volume

dcoredump:
Now SYSEX supports MIDI channels, so there is no need for setting voices to OMNI. Also the master volume SYSEX code is now working.

Note: Newer version from Synth_Dexed is needed!


dcoredump:
Will send a new one. Just have trouble to resolve the conflict...

Issue #83, opened 2022-04-10
Some voices are not played


probonopd:
https://yamahablackboxes.com/patches/dx7/factory/rom3a.syx

* 17=SAX BC
* 24=VOICES BC

https://yamahablackboxes.com/patches/dx7/factory/rom4a.syx

* 10=BRASS BC

https://yamahablackboxes.com/patches/dx7/vrc/vrc106a.syx

* 17=SYN.BRA. 4

https://yamahablackboxes.com/patches/dx7/vrc/vrc106b.syx

* 25=SYN.GUITAR

https://yamahablackboxes.com/patches/dx7/vrc/vrc108a.syx

* 15=Good-Vibes

https://yamahablackboxes.com/patches/dx7/vrc/vrc108b.syx

* 20=BCTrumpet

TODO: Test with other software instruments

probonopd:
One this mystery is solved, we may consider to create nice performances using the TX816  voices.

For example, set TG1...8 to use `Tfi1...8.syx` from `Dexed_cart_1.0/Original Yamaha/TX816/`, each voice 11. (Possibly no additional detune; tbc.) The result is a nice rich piano

probonopd:
@dcoredump can you reproduce this issue, also on MicroDexed?

dcoredump:
I have to check this tomorrow.

dcoredump:
> @dcoredump can you reproduce this issue, also on MicroDexed?

Couldn't try, sorry.

Perhaps those sounds use the breath controller for the AMP envelope?

dcoredump:
Hm, have to check this... sorry, currently not much free time for this 😢

frank-que:
Hello, I've noticed a strange behaviour, sounds that use algorithm 16,17,18 are muted.

frank-que:
I've found where is the problem:
turn left one step "A MOD Sens" on OSC1 and the sound are back.
All setting with "A MOD Sens" on OSC1 at maximum value are muted.
There is some problem there...

dcoredump:
> Hello, I've found where is the problem: turn left one step "A MOD Sens" on OSC1 and the sounds are back. In some conditions with "A MOD Sens" on OSC1 at maximum value sounds are muted.

Hi @frank-que!

Thanks for finding THIS issue... I will try to fix it inside Synth_Dexed asap.

Regards, Holger

frank-que:
@dcoredump 
[...]

PR #267, opened 2022-06-02
Switchable performance files improvements + Poly/Mono  + Fix to #298 and #83 issues

arsamus:
Hi @probonopd , in order to complement the swichable Performance solution, I have added these additional features to improve the usability of the solution:
-Custom defined new performance name: predefined Performance name can be modified. Characters are input by the rotary encoder, to move the cursor, hold down the encoder and turn right or left. Then just double click to confirm.
-Delete Performance files: This allows to delete perfomance files from the list. To do that, just double click on the already loaded performance and confirm yes or no. The default performance.ini file in the root directory can't be deleted.
-Change voice name: Edited voice can have a custom name (and will be saved on performance file).
-Some minor improvements on related Methods.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -103,7 +103,7 @@ void CConfig::Load (void)
 
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
-	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 0) != 0;
+	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 1) != 0;
 }
 
 const char *CConfig::GetSoundDevice (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -88,6 +88,7 @@ void CConfig::Load (void)
 
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
+	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 0) != 0;
 }
 
 const char *CConfig::GetSoundDevice (void) const
@@ -208,4 +209,9 @@ bool CConfig::GetMIDIDumpEnabled (void) const
 bool CConfig::GetProfileEnabled (void) const
 {
 	return m_bProfileEnabled;
+}
+
+bool CConfig::GetPerformanceSelectToLoad (void) const
+{
+	return m_bPerformanceSelectToLoad;
 }
--- a/src/config.h
+++ b/src/config.h
@@ -98,6 +98,9 @@ public:
 	// Debug
 	bool GetMIDIDumpEnabled (void) const;
 	bool GetProfileEnabled (void) const;
+	
+	// Load performance mode. 0 for load just rotating encoder, 1 load just when Select is pushed
+	bool GetPerformanceSelectToLoad (void) const;
 
 private:
 	CPropertiesFatFsFile m_Properties;
@@ -130,6 +133,7 @@ private:
 
 	bool m_bMIDIDumpEnabled;
 	bool m_bProfileEnabled;
+	bool m_bPerformanceSelectToLoad;
 };
 
 #endif
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -33,9 +33,12 @@ LOGMODULE ("mididevice");
 #define MIDI_NOTE_OFF		0b1000
 #define MIDI_NOTE_ON		0b1001
 #define MIDI_AFTERTOUCH		0b1010			// TODO
+#define MIDI_CHANNEL_AFTERTOUCH 0b1101   // right now Synth_Dexed just manage Channel Aftertouch not Polyphonic AT -> 0b1010
 #define MIDI_CONTROL_CHANGE	0b1011
 	#define MIDI_CC_BANK_SELECT_MSB		0	// TODO
 	#define MIDI_CC_MODULATION			1
+	#define MIDI_CC_BREATH_CONTROLLER	2 
+	#define MIDI_CC_FOOT_PEDAL 		4
 	#define MIDI_CC_VOLUME				7
 	#define MIDI_CC_PAN_POSITION		10
 	#define MIDI_CC_BANK_SELECT_LSB		32
@@ -226,6 +229,12 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 						m_pSynthesizer->keyup (pMessage[1], nTG);
 						break;
 		
+					case MIDI_CHANNEL_AFTERTOUCH:
+						
+						m_pSynthesizer->setAftertouch (pMessage[1], nTG);
+						m_pSynthesizer->ControllersRefresh (nTG);
+						break;
+							
 					case MIDI_CONTROL_CHANGE:
 						if (nLength < 3)
 						{
@@ -238,7 +247,17 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 							m_pSynthesizer->setModWheel (pMessage[2], nTG);
 							m_pSynthesizer->ControllersRefresh (nTG);
 							break;
-		
+								
+						case MIDI_CC_FOOT_PEDAL:
+							m_pSynthesizer->setFootController (pMessage[2], nTG);
+							m_pSynthesizer->ControllersRefresh (nTG);
+							break;
+
+						case MIDI_CC_BREATH_CONTROLLER:
+							m_pSynthesizer->setBreathController (pMessage[2], nTG);
+							m_pSynthesizer->ControllersRefresh (nTG);
+							break;
+								
 						case MIDI_CC_VOLUME:
 							m_pSynthesizer->SetVolume (pMessage[2], nTG);
 							break;
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -52,7 +52,8 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	m_bProfileEnabled (m_pConfig->GetProfileEnabled ()),
 	m_bSavePerformance (false),
 	m_bSavePerformanceNewFile (false),
-	m_bSetNewPerformance (false) 
+	m_bSetNewPerformance (false),
+	m_bDeletePerformance (false)
 {
 	assert (m_pConfig);
 
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1283,8 +1283,7 @@ bool CMiniDexed::SavePerformanceNewFile ()
 
 bool CMiniDexed::DoSavePerformanceNewFile (void)
 {
-	std::string nPerformanceName=""; // for future enhacements: capability to write performance name - Not implemented, delete and modify CreateNewPerformanceFile(Name) to CreateNewPerformanceFile(void)
-	if (m_PerformanceConfig.CreateNewPerformanceFile(nPerformanceName))
+	if (m_PerformanceConfig.CreateNewPerformanceFile())
 	{
 		if(SavePerformance())
 		{
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1261,7 +1261,7 @@ void CMiniDexed::SetMenuSelectedPerformanceID(unsigned nID)
 {
 	m_PerformanceConfig.SetMenuSelectedPerformanceID(nID);
 }
-/* 
+*/ 
 //  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%5
 
 bool CMiniDexed::SetNewPerformance(unsigned nID)
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -53,7 +53,8 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	m_bSavePerformance (false),
 	m_bSavePerformanceNewFile (false),
 	m_bSetNewPerformance (false),
-	m_bDeletePerformance (false)
+	m_bDeletePerformance (false),
+	m_bLoadPerformanceBusy(false)
 {
 	assert (m_pConfig);
 
@@ -226,7 +227,7 @@ bool CMiniDexed::Initialize (void)
 		return false;
 	}
 #endif
-
+	
 	return true;
 }
 
@@ -264,10 +265,14 @@ void CMiniDexed::Process (bool bPlugAndPlayUpdated)
 		m_bSavePerformanceNewFile = false;
 	}
 	
-	if (m_bSetNewPerformance)
+	if (m_bSetNewPerformance && !m_bLoadPerformanceBusy)
 	{
 		DoSetNewPerformance ();
-		m_bSetNewPerformance = false;
+		if (m_nSetNewPerformanceID == GetActualPerformanceID())
+		{
+			m_bSetNewPerformance = false;
+		}
+		
 	}
 	
 	if(m_bDeletePerformance)
@@ -954,9 +959,10 @@ void CMiniDexed::ProcessSound (void)
 
 #endif
 
-bool CMiniDexed::SavePerformance (void)
+bool CMiniDexed::SavePerformance (bool bSaveAsDeault)
 {
 	m_bSavePerformance = true;
+	m_bSaveAsDeault=bSaveAsDeault;
 
 	return true;
 }
@@ -998,6 +1004,11 @@ bool CMiniDexed::DoSavePerformance (void)
 	m_PerformanceConfig.SetReverbDiffusion (m_nParameter[ParameterReverbDiffusion]);
 	m_PerformanceConfig.SetReverbLevel (m_nParameter[ParameterReverbLevel]);
 
+	if(m_bSaveAsDeault)
+	{
+		m_PerformanceConfig.SetNewPerformance(0);
+		
+	}
 	return m_PerformanceConfig.Save ();
 }
 
@@ -1263,17 +1274,21 @@ bool CMiniDexed::SetNewPerformance(unsigned nID)
 
 bool CMiniDexed::DoSetNewPerformance (void)
 {
+	m_bLoadPerformanceBusy = true;
+	
 	unsigned nID = m_nSetNewPerformanceID;
 	m_PerformanceConfig.SetNewPerformance(nID);
 	
 	if (m_PerformanceConfig.Load ())
 	{
 		LoadPerformanceParameters();
+		m_bLoadPerformanceBusy = false;
 		return true;
 	}
 	else
 	{
 		SetMIDIChannel (CMIDIDevice::OmniMode, 0);
+		m_bLoadPerformanceBusy = false;
 		return false;
 	}
 }
@@ -1288,7 +1303,7 @@ bool CMiniDexed::DoSavePerformanceNewFile (void)
 {
 	if (m_PerformanceConfig.CreateNewPerformanceFile())
 	{
-		if(SavePerformance())
+		if(SavePerformance(false))
 		{
 			return true;
 		}
@@ -1395,4 +1410,7 @@ bool CMiniDexed::DoDeletePerformance(void)
 	return false;
 }
 
-
+bool CMiniDexed::GetPerformanceSelectToLoad(void)
+{
+	return m_pConfig->GetPerformanceSelectToLoad();
+}
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -72,7 +72,7 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 		m_nPortamentoMode[i] = 0;
 		m_nPortamentoGlissando[i] = 0;
 		m_nPortamentoTime[i] = 0;
-
+		m_bMonoMode[i]=0; 
 		m_nNoteLimitLow[i] = 0;
 		m_nNoteLimitHigh[i] = 127;
 		m_nNoteShift[i] = 0;
@@ -676,6 +676,7 @@ void CMiniDexed::SetTGParameter (TTGParameter Parameter, int nValue, unsigned nT
 	case TGParameterPortamentoMode:		setPortamentoMode (nValue, nTG);	break;
 	case TGParameterPortamentoGlissando:	setPortamentoGlissando (nValue, nTG);	break;
 	case TGParameterPortamentoTime:		setPortamentoTime (nValue, nTG);	break;
+	case TGParameterMonoMode:		setMonoMode (nValue , nTG);	break; 
 
 	case TGParameterMIDIChannel:
 		assert (0 <= nValue && nValue <= 255);
@@ -710,7 +711,7 @@ int CMiniDexed::GetTGParameter (TTGParameter Parameter, unsigned nTG)
 	case TGParameterPortamentoMode:		return m_nPortamentoMode[nTG];
 	case TGParameterPortamentoGlissando:	return m_nPortamentoGlissando[nTG];
 	case TGParameterPortamentoTime:		return m_nPortamentoTime[nTG];
-
+	case TGParameterMonoMode:		return m_bMonoMode[nTG] ? 1 : 0; 
 	default:
 		assert (0);
 		return 0;
@@ -983,6 +984,7 @@ bool CMiniDexed::DoSavePerformance (void)
 		m_PerformanceConfig.SetNoteShift (m_nNoteShift[nTG], nTG);
 		m_pTG[nTG]->getVoiceData(m_nRawVoiceData);  
  		m_PerformanceConfig.SetVoiceDataToTxt (m_nRawVoiceData, nTG); 
+		m_PerformanceConfig.SetMonoMode (m_bMonoMode[nTG], nTG); 
 		
 		m_PerformanceConfig.SetReverbSend (m_nReverbSend[nTG], nTG);
 	}
@@ -1003,7 +1005,7 @@ void CMiniDexed::setMonoMode(uint8_t mono, uint8_t nTG)
 {
 	assert (nTG < CConfig::ToneGenerators);
 	assert (m_pTG[nTG]);
-
+	m_bMonoMode[nTG]= mono != 0; 
 	m_pTG[nTG]->setMonoMode(constrain(mono, 0, 1));
 	m_pTG[nTG]->doRefreshVoice();
 	m_UI.ParameterChanged ();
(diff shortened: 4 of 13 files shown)


miotislucifugis:
Unfortunately, I cant get this build to work- using a pi3.     MiniDexed appears to boot, LCD shows "home screen"= TG1... but it does not respond to the encoder/switch  or make sound when I play the keyboard.

arsamus:
> Unfortunately, I cant get this build to work- using a pi3. MiniDexed appears to boot, LCD shows "home screen"= TG1... but it does not respond to the encoder/switch or make sound when I play the keyboard.

I just realized that I had missed to initialize a variable when I've copy the code here, may be that was the reason of the bug. Unfortunatly, I don't have a RPI3 to test it and I have no problems on a RPI2, so let me know if on this last build the issue has been resolved.

miotislucifugis:
That fixed it.  Just did a quick check and everything seems to be working.  great work

arsamus:
Great! 😃

BobanSpasic:
@arsamus - did you take a look at this: https://github.com/probonopd/MiniDexed/discussions/261
Thx for all your work

arsamus:
> @arsamus - did you take a look at this: https://github.com/probonopd/MiniDexed/discussions/261
> Thx for all your work

Yes. I've got it. I upload asap with this pr.

probonopd:
Great @arsamus. Please give me a ping once you think this should be merged. Thanks!

arsamus:
Hi, @BobanSpasic, I've just added on this PR a Poly/Mono switch via UI to set Poly/Mono mode (#261 ). You can test it with this build (https://github.com/probonopd/MiniDexed/suites/6801306556/artifacts/261228550). You found it after "portamento" sub menu on each TG sub menu.

miotislucifugis:
No problems here.   The mono mode really begs for a key priority option...not sure how it was on the dx7 and/or if it is supported in synth_dexed.   Currently it is "highest note" ...which definitely is not my preferred behavior.

BobanSpasic:
@arsamus - I can test next week. I am on vacation this week, away from home.

@miotislucifugis - DX has Mono with "latest note" priority. Some other synths has also "highest" and "lowest" as options.

miotislucifugis:
> @miotislucifugis - DX has Mono with "latest note" priority. Some other synths has also "highest" and "lowest" as options.

it only has "latest note" priority?  It cannot be changed?   If i can only have one option,  latest is definitely the way to go.  Changing from "highest", as it is currently to "last", which is the original/preferred behavior, would have to be done in synth_dexed, right?

arsamus:
> Great @arsamus. Please give me a ping once you think this should be merged. Thanks!

Hi, @probonopd everything works as expeted, so it could be merged. This is for WIKI: 

Main menu->Performance>allows to select a Performance file. Saved configuration will be load when double click on desire file. Once loaded, "[LD]" mark will apears on upper left corner of the display. Double clicking on already loaded Performance file, allows to delete it (after Yes/No confirmation). Default performance file (Performance.ini on root directory) can't be deleted.
Main Menu->Save->Overwrite: All parameters edited will be saved overwritten the already loaded performance file.
Main Menu->Save->New: A new performance file will be created on "performance" directory. The default name can be edited vía rotary encoder, press it and move right or left to set the cursor position and then move the encoder to select the desired character. Then just double click to confirm. 
Main Menu->TG1->Poly/Mono (after portamento menu): Set and unset "mono" mode. (right now, Synth Dexed allows only hi-note proirity mono mode)
TG1->Edit Voice>Name (after "Transpose"): Edit voice name. New name can be saved, as part of voice data, on Performance files. This procedure not modifies bank files saved on sysex/voice directory.  Edition can be done vía rotary encoder, press it and move right or left to set the cursor position and then move the encoder to select the desired character. Then just click to return to parent menu.

probonopd:
Hello @arsamus, thanks for the ping.
You are really squeezing a lot of functionality out of our poor little rotary encoder ;-)

Users have asked for more flexibility in configuring the input devices (rotary encoders, buttons, etc.). For example, there seems a pretty broad consensus in the user base that double clicking for opening submenus is not ideal (hence, the possibility to use single clicks for select/confirm/OK with `ButtonActionSelect=click`), and some users would like an extra button for back/cancel. So work is underway to have more abstraction, see #274. I am now a bit concerned that the logic for saving and renaming performances might clash with the input device abstraction introduced there...

> Custom defined new performance name: predefined Performance name can be modified. Characters are input by the rotary encoder, to move the cursor, hold down the encoder and turn right or left. Then just double click to confirm.

Can we make use of the variables introduced by #274 once it is merged, e.g., "confirm by using the configured (ButtonPinSelect and ButtonActionSelect)[https://github.com/probonopd/MiniDexed/pull/274/files#diff-525537093a4f031fba1efead66adec4a99230d119edd545cc83ba25fa4ae1c41R38-R39] instead of "double click to confirm"? (Same for editing voice names)

> Delete Performance files: This allows to delete perfomance files from the list. To do that, just double click on the already loaded performance and confirm yes or no. The default performance.ini file in the root directory can't be deleted.

Let's not hide functionality in unexpected ways like this. Logical and discoverable menu structures would be:

* Performance -> Load -> (select a performance by turning the knob)
* Performance -> Save -> (select a performance)
* Performance -> Rename -> (select a performance)
* Performance -> Delete -> (select a performance)

or:

* Performance -> (select a performance) -> Load
* Performance -> (select a performance) -> Save
* Performance -> (select a performance) -> Rename
* Performance -> (select a performance) -> Delete

I think it's worth the small extra step in loading performances to have a logical menu that makes it easy to find functionality. Loading a performance file would not even require more clicks (if you configure `ButtonActionSelect=click`).

Wdyt?

arsamus:
Hi @probonopd, sounds good. This could be a posible menu tree in order to be align with the idea:
Performance
--->Load-->(select performance)(select to confirm)*

--->Save-->Overwrite(select to confirm)
              ->New-->[input name](selecto to confirm)
              ->Set as default(select to confirm)

-->Delete->Yes/No(select to confirm)

-->Rename->[edit name](select to confirm). Not implemented yet

*Right now performances are loaded from file system at the time they are selected, preload them on RAM, like voice banks, implies a more agresive redesign of performance.cpp class. That is why it is necessary to push select (by double click like now or by one click/push a button like future features ) in order to avoid unnecessary file system access. Besides this technical issue, for live performance, is it better to go to the desired performance just rotating the knob, found it through the list, or preselec it and just doble click (or click) at the time the change is needed?  The second I think is more near to a "full buttons and knobs" synth.

Respect to button changes. Maybe we have to wait until the final solution, but as I could see on a fast reading, the proposed changes affects things that trigger the events but not the events, so we may replace in language words "click" and "double click" for "select" and "cancel/back". Am I right?

arsamus:
Hi @probonopd I've just update the menu tree according to what was discussed above. On minidexed.ini there are a new property to set load behavior. If  `PerformanceSelectToLoad=1` then to load a the performance, the user has to push Select, otherwise, the performance can be selected just turning the knob (for testing only, but maybe this feature has to be oculted before merge because it needs some imrpovements)

arsamus:
Hi @probonopd, I've add to this PR a new menu in order to fix #298 and #83 issues. 

The whole PR include: 
-An inprovement of switchable Performance files according to the discussion of some weeks ago. 
-The addition of Poly/Mono menu
-The addition of Modulation menu according to DX7 / Dexed modulation settings (this settings and the following fix the issues #298 and #83). 
-A new default Mod Wheel settings. Now default modulation wheel target is only Pitch (not EG Bias and not Amplitude). 
-Voice name edition

All of this settings can be loaded and saved on Performance files.

miotislucifugis:
Wow Super!   I just tried it.   Mod wheel is working well - I can turn on and off the mod destinations... unfortunately  no luck with  aftertouch, breath, and foot control (cc#2&4, respectively.)   Has anyone else tested these other mod sources?

arsamus:
> Wow Super!   I just tried it.   Mod wheel is working well - I can turn on and off the mod destinations... unfortunately  no luck with  aftertouch, breath, and foot control (cc#2&4, respectively.)  Even when enabled, there does not seem to be any effect.    Has anyone else tested these other mod sources?

Hi @miotislucifugis.  I've just test it and I have the same issue. Let me see why because the code is very simple and identical for the 4 controllers.... I have an hypothesis related to one parameter called "mode" that synth dexed has for them. Besides Mod Wheel, other 3 controllers are not been initialized by MiniDexed at the begining so this parameter is not setted. I am not sure but his could be the cause of the issue, let me check it.

BobanSpasic:
What's the function of the BankNumber and VoiceNumber in performance.ini when the voice parameters are already included in VoiceData?

arsamus:
> What's the function of the BankNumber and VoiceNumber in performance.ini when the voice parameters are already included in VoiceData?

If you leave blank VoiceData on performance.ini (or in custom files), Bank and Voice number parameters are considered, could be useful for a quick perferomance file editing, but ones voicedata is present both parameters are ignored by Synth Dexed and them just been used for setting Voice and Bank navigation menu. If voice has not been edited, redundant data is stored, but it allow to interchange performance files between different dispositives and users. Otherwise a standard bank order would have to be defined.

arsamus:
> > Wow Super!   I just tried it.   Mod wheel is working well - I can turn on and off the mod destinations... unfortunately  no luck with  aftertouch, breath, and foot control (cc#2&4, respectively.)  Even when enabled, there does not seem to be any effect.    Has anyone else tested these other mod sources?
> 
> Hi @miotislucifugis. I've just test it and I have the same issue. Let me see why because the code is very simple and identical for the 4 controllers.... I have an hypothesis related to one parameter called "mode" that synth dexed has for them. Besides Mod Wheel, other 3 controllers are not been initialized by MiniDexed at the begining so this parameter is not setted. I am not sure but his could be the cause of the issue, let me check it.

I've got it. My hypothesis was wrong, the issue is related to midi messages handler (mididevice.cpp). The existing code just handle mod wheel control but it doesn't catch foot, breath or aftertouch messages. I can add those features on this days as soon as I have sometime to do it.

arsamus:
Hi everybody and @miotislucifugis, I've just submit the PR that fixed Foot Controller, Breath Controller and Aftertouch message handling. Right now all modulation paths are working as expected (like Dx7 is supposed to do it, unfortunely I don't have a real one to test it) and also complement #298 and #83 solution. New build is ready to test it: https://github.com/probonopd/MiniDexed/suites/7282549823/artifacts/293357205

BobanSpasic:
@arsamus - I would have one more request - could you add to the performance.ini one more key (PerformanceName=char[1..30])?
This way, we could get all the parameters of a DX/TX performance dump into performance.ini

arsamus:
Hi @BobanSpasic. Yes, it could be added,  but take in mind that performance.ini (and the rest of switchable files) save in one file the eight performance parameters of each TG of MiniDexed so we will have to face with this things:
1) we will have two "performance" meanings, one for the whole instrument (the eight TG as a whole) and one for each TG (for each TX/DX instance). Right now when we talk about performance we  refer to the first one. So, when we will talk about performance in TG (DX instance) context we have to use another name in order to avoid user confusion.
2) we have to think how to handle 30 chars on 16 chars displays (we may need to add some code for scrolling animation or something like that)
3) this name it doesn't take any utility (it only is my point of view, I could be wrong) on MiniDexed context since performance.ini save all TG in one file. On other hand, it is supposed that each TG mimics a basic DX7 (of course I think that this project have the potencial to be more than that on a future, adding more features backward compatible with DX FM synthesis, for examaple having other waveforms for Ops, adding extra algorithms, adding multi effect send channels, etc )

Other approach could be having the name just for inport/export TX Sysex tasks, but hidden on UI. Names could be stored on performance.ini (users could be edited there if they want) and can be added on future feature that export sysex messages on TX format.
Performance created on MiniDexed could have a default names, for example TG1, TF2, etc. 

@probonopd what it's your point of view?

BobanSpasic:
DX7II can play 2-voice performance, either as solo (just Voice A), split or dual. There is a defined SysEx bulk dump for such Performance. The performances have names and there are 64 storage positions on DX7II for the performances (32 on TX7) according to the documentation.
I still must search for TX816, how is the situation there. I got the manual, but I can't find any definition of performance bulk dumps for TX816.
MiniDexed have much more possibilities than any of DX/TX. At some point, velocity layering could be implemented (e.g. TG1 - TG7 plays all the notes from the whole velocity range and TG8 plays some explosive brass section if the velocity is over 90).
No DX/TX has this implemented (as far as I know).
I would use the name "performance" just like it is - the setup of the whole MiniDexed in one file. This would follow the logic of DX7mkII, DX11 etc. where Performance sets up all the voice generators.

1. "Performance" per TG can be called Addition, Supplement etc. Yamaha did it so for all the extra parameters the new generations have, but original DX7 do not have. DX7mkII uses SysEx with original DX7 voice + supplement. If you sent the same SysEx to original DX7 - it will use just the Voice part and the Supplement part is ignored. Do MiniDexed really needs supplement per TG?

2. We don't need to show 30 chars on 16 chars display, we can edit the names. Let it be 30 defined just for compatibility. Or make it 16, no problem here either.

3. There is no use at the moment, but at some point we will or could have a SysEx dump for performances (PC to MiniDexed), and MiniDexed could take this 30 char name for file name (for storage on the SD Card).

4. Forget about importing TX performance SysEx directly, without conversion software on a PC. The performance dump from a DX7mkII or TX alone, without the same voice banks setups, means nothing. Here comes the advantage of performance.ini files, where the voices are integrated into the performance files. I just love this idea. That is also the reason for my question earlier, if we really need the VoiceNumber and BankNumber in the performance.ini files - with just these two parameters, the performance gets lost when you load another bank in the same place (just like it is lost on a real DX). Another problem is incompatibility between DX/TX series itself - one series have some parameter in the range 0-15 where the same parameter on other series is in the range 0-99. It needs to be converted.

BobanSpasic:
Got some info:
TX816 = 8 x independent DX7 (mkI, alias original DX7), no global performance file.
TX802 = 8 x DX7mkII (alias DX7II) - performance files are something like MiniDexed's performance.

arsamus:
,>I would use the name "performance" just like it is - the setup of the whole MiniDexed in one file. This would follow the logic of DX7mkII, DX11 etc. where Performance sets up all the voice generators.

The last build of this PR (https://github.com/probonopd/MiniDexed/suites/7282549823/artifacts/293357205) allows to save the whole MiniDexed parameters in one file and named with a 14 chars name that can be editable before save it (renaming saved files is pending). This covers what you think about performance management? Why do you think that it is necessary to add a name parameter inside the file? The loaded performance name can be checked on Perfrormance->load menu. Since the file is loaded at the time the user wants, there are no limit for performance files (just de SD card capacity limit)
 

>TX816 = 8 x independent DX7 (mkI, alias original DX7), no global performance file.
TX802 = 8 x DX7mkII (alias DX7II) - performance files are something like MiniDexed's performance.

Respect to tx802. Performance files save the whole unit as a whole (8 TF modules on the same file (or sysex dump) like MiniDexed) or each module has it own performance file?

>MiniDexed have much more possibilities than any of DX/TX. At some point, velocity layering could be implemented (e.g. TG1 - TG7 plays all the notes from the whole velocity range and TG8 plays some explosive brass section if the velocity is over 90).

Velocity layering, global transpose and key range are relatively easy to implement with a few lines of code, but of course, this is not part of dx7 features. Maybe on other PR I can suggest some of this features, this PR has a lot of new features and covers, as I know, the DX7 performance parameters respect that are missing on the Productive build.

BobanSpasic:
> > I would use the name "performance" just like it is - the setup of the whole MiniDexed in one file. This would follow the logic of DX7mkII, DX11 etc. where Performance sets up all the voice generators.
> 
> The last build of this PR (https://github.com/probonopd/MiniDexed/suites/7282549823/artifacts/293357205) allows to save the whole MiniDexed parameters in one file and named with a 14 chars name that can be editable before save it (renaming saved files is pending). This covers what you think about performance management? Why do you think that it is necessary to add a name parameter inside the file? The loaded performance name can be checked on Perfrormance->load menu. Since the file is loaded at the time the user wants, there are no limit for performance files (just de SD card capacity limit)
> 
Because one day we will want to SysEx dump the performance file, and in such case the name must be written inside the file.
I am writing a Librarian app for MiniDexed, and my points are going in this direction - how to transfer the performance from a PC app to MiniDexed.
> > TX816 = 8 x independent DX7 (mkI, alias original DX7), no global performance file.
> > TX802 = 8 x DX7mkII (alias DX7II) - performance files are something like MiniDexed's performance.
> 
> Respect to tx802. Performance files save the whole unit as a whole (8 TF modules on the same file (or sysex dump) like MiniDexed) or each module has it own performance file?
> 
One file for performance:
tx802
The other parameters from MiniDexed's performance.ini (portamento parameters, poly/mono etc.) are part of Voice (this is the difference between DX7/TX816 and DX7II/TX802) as Supplement:
DX7II

For TX802, the performance name is 20 char long, for TX816 is 30, for V50 is 10 etc...
> > MiniDexed have much more possibilities than any of DX/TX. At some point, velocity layering could be implemented (e.g. TG1 - TG7 plays all the notes from the whole velocity range and TG8 plays some explosive brass section if the velocity is over 90).
> 
> Velocity layering, global transpose and key range are relatively easy to implement with a few lines of code, but of course, this is not part of dx7 features. Maybe on other PR I can suggest some of this features, this PR has a lot of new features and covers, as I know, the DX7 performance parameters that are missing on the Productive build.
I agree.

probonopd:
Dear all, thanks for your hard work on this PR. Please test the new build [MiniDexed_2022-07-15-144ef8d](https://github.com/probonopd/MiniDexed/suites/7378043811/artifacts/299938527) which is based on top of the now merged https://github.com/probonopd/MiniDexed/pull/274.

arsamus:
>Please test the new build [MiniDexed_2022-07-15-144ef8d](https://github.com/probonopd/MiniDexed/suites/7378043811/artifacts/299938527) which is based on top of the now merged https://github.com/probonopd/MiniDexed/pull/274.

Hi @probonopd I've tested it and everything works as expected and compatible with #274

probonopd:
>> @arsamus - I would have one more request - could you add to the performance.ini one more key (PerformanceName=char[1..30])?
This way, we could get all the parameters of a DX/TX performance dump into performance.ini
>
> we will have two "performance" meanings, one for the whole instrument (the eight TG as a whole) and one for each TG (for each TX/DX instance)

I was under the impression that on the TX, a "performance" describes the settings how the various TG instances work together (and possibly which effects are added). So I would not use the word "performance" to describe the individual TGs, unless I am overlooking something.

> So, when we will talk about performance in TG (DX instance) context we have to use another name in order to avoid user confusion.

How is Yamaha calling those things in their TX documentation?

> "Performance" per TG can be called Addition, Supplement etc. Yamaha did it so for all the extra parameters the new generations have, but original DX7 do not have. DX7mkII uses SysEx with original DX7 voice + supplement.

So, I'd be fine with "Supplement".

> I would use the name "performance" just like it is - the setup of the whole MiniDexed in one file. 

I tend to agree, unless this is confusing because Yamaha users expect something entirely different.

> we have to think how to handle 30 chars on 16 chars displays

Frankly, I would not care too much at this point and just show the first characters. This is purely cosmetic and maybe people will use larger displays with more characters soon.

> We don't need to show 30 chars on 16 chars display, we can edit the names. 

Yes. Allow 30 chars to be stored but only show what fits on the respective display.

> import/export TX Sysex tasks

Once we get to the point that MiniDexed can do this, one _could_ consider to use the TX sysex data format for storing things locally (instead of `performance.ini` files) as I had originally assumed we would do. But then, by now it has become clear that `performance.ini` files are easier to work with than `.syx` files, and - most importantly - they contain the voice data so that everything ends up together in just one file that can easily be moved from machine to machine, so probably using the TX sysex data format for storing things locally should not be our goal anymore. Wdyt?

> Performance created on MiniDexed could have a default names, for example TG1, TF2, etc.

Agree.

Testing [MiniDexed_2022-07-15-144ef8d](https://github.com/probonopd/MiniDexed/suites/7378043811/artifacts/299938527) now. @arsamus do you think this PR is ready for merging? Additional features and improvements to existing features could always be brought in through additional PRs later. I'd like to have major changes (like this one) in main rather sooner than later, so that other PRs won't run into merge conflicts eventually.

Thanks for your great work!

arsamus:
>@arsamus do you think this PR is ready for merging? Additional features and improvements to existing features could always be brought in through additional PRs later. 

Hi @probonopd, yes I think is ready to merge. This PR adds a lot of new of important features. We can continue discussing about other features, improvements and ideas that can be included on future PRs.

probonopd:
Great @arsamus. Could you suggest edits for https://github.com/probonopd/MiniDexed/wiki/Menu?

arsamus:
Yes, I will post it here as soon as I can.

probonopd:
Which defaults do we have do change so that the "silent" BC voices https://github.com/probonopd/MiniDexed/issues/83 play by default?

BobanSpasic:
Explanation about the performances on real DX7
DX7 - single TG, performance is global (and just for one TG). Performance parameters are these that we have in MiniDexed performance.
TX816 - 8xTG, performance per TG. No global performance file. Every TG in TX816 is a separate device.

DX7mkII - 2xTG. Most of the parameters from performance is moved to voice parameters (supplement data). The remaining parameters in performance are for setting both the TGs (keyboard split, dual etc.)
TX802 - 8xTG. same logic like DX7mkII

MiniDexed uses DX7 (not mkII) parameters and performance.ini is DX7-alike, but for 8 TGs.
Importing DX7mkII voices would mean splitting it at DX7 Sysex + parameters that are going into performance.ini.

arsamus:
I don't know which are factory default values for this controllers. Intuition tell me that just Pitch would be setted 'on' and amplitude and EG Bias setted 'off'. I don't think Yamaha default roms have voices that are muted by default. But maybe someone who have a real DX7 can confirm. Maybe @SurvivalHacking can (if this doesn't bother him), I saw some of his YT videos and he have one.

arsamus:
Hi, @probonopd here are the insertions and changes for Wiki->Menu. Default values for Mod Wheel is Range=99, Pitch=ON, Amplitude=OFF, EG BIAS=OFF; rest of controllers: Range=99, Pitch=OFF, Amplitude=OFF, EG BIAS=OFF to avoid unpredictive behavior (like muted voices);

[insert Poly/Mono and Modulation between Portamento and Channel inside TG menu]
Poly/Mono: Enable and Disable monophonic (single note synth with upper note prioroty) or polyphonic mode of each TG. 

[Modulation:
Allows to set (On and Off) modulation wheel, foot controllers, breath controller and after touch targets. For any of this controllers there are three main components of voice to witch LFO modulation can be applied: to the PITCH, to the AMPLITUDE and to EG (Envelope Generator) BIAS. Actually each Operator may or may not be progremmed to be sensitive to modulation sources. Range parameter allows to set the amount of effect that will be apply to targets. Some settings could result in little or no audible sound, especially when EG BIAS is On.

[on main menu replace "save" with this "performance" menu]
Performance:
->Load: allows to switch between different performance files. Right now move the rotary encoder or increment/decrement buttons in order to select the desired Performance file, then press select to load it. The flag "[L]" at the upper left of the display indicate the current loaded file

->Save:
     -->Overwrite: write changes to the current performance file 
     -->New: create a new performance file. By default a suggested name will be generated, but it can be edited before the new file will be created. To do that, Push Select button and move rotary encoder or increment/decrement buttons to move cursor. Select to confirm and back to abort 
     -->Set as default: Save the current performance parameters to default file (performance.ini file on root directory). Default file contains the parameters that are going to be loaded when MiniDexed is turned on.  Take in mind that performance.ini file will be overwritten after this action.

->Delete: this function allows to delete custom performance files. Default file (performance.ini at root directory) can't be deleted. 
 
[After Transpose and before TG2]
Voice name: Change current voice name. Right now, this changes, like other voice changes, will not be saved on rom banks. Any voice parameter change can be saved on performance files. See Performance menu. Shortcut to switch between TGs in disabled on this menu because the functionality is used for change the cursor position

probonopd:
> EG BIAS=OFF to avoid unpredictive behavior (like muted voices)

Are you saying that by default you are not getting muted voices anymore? Then in my last test I must have done something wrong, because I still got muted voices as in #83. Maybe because I had pre-existing Performance files. Will delete those and try again.

miotislucifugis:
By leaving EG Bias "off" on all mod sources, you will have "muted" voices.   The patch expects EG bias to be controlled by a modulation source (usually BC, but can be the mod wheel or aftertouch;)   as it is set to a minimum.   Only when "modulated" will it open up and produce audible sound.   I dont think that there is such a thing as "default" mod setting on the DX7... unless you mean what the setting is after changing the internal battery or what it was set to right out of the box.    Unless I am mistaken, the DX7 remembers the last user settings for stuff like modulation, pitchbend, and potamento, etc.    I would need to test to confirm 100%( perhaps tomorrow), but i suspect that is is totally normal for the DX7 to produce no sound on these patches if no EG Bias modulation source is connected and value wiggled.   often these patches are labeled with the letters "BC"   to indicated that they are expecting this to be set up.  for example voice #14 on rom 2A= "Brass 6 BC" ... tho' others like  "train" (i forget which rom this was on)   make sound by default(chug chug), but rely on something to modulate the EG bias to animate the sound and add the train whistle effect.    All the original yamaha patch that require this seem to be on the supplementary rom carts... perhaps there was a note in the original paperwork/ documentation that made note of the modulation requirements?

arsamus:
>By leaving EG Bias "off" on all mod sources, you will have "muted" voices.

I don't know what happens on a real DX7 but on Desktop Dexed and on Minidexed, setting off EG Bias unmute those "muted" voices. 

>The patch expects EG bias to be controlled by a modulation source (usually BC, but can be the mod wheel or aftertouch;) as it is set to a minimum. Only when "modulated" will it open up and produce audible sound.[...]often these patches are labeled with the letters "BC" to indicated that they are expecting this to be set up.

I didn't notice that BC means Breath Controller... And what you say sounds convincent.

>unless you mean what the setting is after changing the internal battery or what it was set to right out of the box

Yes. This is what I mean, but since DX7 remembers the last user settings until someone disconnect the battery, because DX7 doesn't have a factory reset, and these controllers parameters are not part of voice parameters, I'd think that it doesn't matters these default values. Eventhough, we can define as default values the most "standard" behavior, for example pitch on and amplitude and eg bias off, or set the factory values if some who have a DX7 can check them.

miotislucifugis:
I tested the some of "quiet" patches today on a DX7.   Patches were sent via desktop dexed.  the patches were a few of the "BC"  patches from yamaha roms 3a and 4a  = "brass BC", "sax BC" and "voice BC".    As i suspected, when EQ Bias is turn off for all the mod sources (or mod source level is at 0)   the patch remains quiet.     (by the way,  for me, these patch are also quiet on desktop dexed:  As i sent the patches to the DX7, I tested the sound in Dexed with the little on screen keyboard, and there was no sound.     What patches were quiet in Minidexed, but had sound in Dexed?)

One small difference between dexed/minidexed and the real DX7-   on both dexed variants, when the EQ Bias is down or not being modulated, there is no sound from the patch... on the dx7,   while it is quiet and muted, you _can_ hear the sound... it seems like the filter effect it isnt fully being closed... if that makes sense.    I know that the DX7 uses a range of 100 for the velocity.   just a guess, but perhaps it also uses this range internally for other variables like modulation level...while Dexed uses 0-127.    however I would think that the range would be higher, not lower?  ....shrug?

 regardless,   its not really an issue... i kinda prefer how it behaves in dexed vs the original anyway.      That aside, the behavior in this merge request seems constant with both the DX7 and Desktop dexed.

probonopd:
Sorry, but it is still not working as I think it should. To reproduce:
* Delete/rename the `performance` folder on the SD card if you had one
* Install the latest build from this PR
* In TG1 to TG4, set voice to __SAX BC__
* Play some notes. You hear nothing. I would like the defaults to be set up so that you hear something. Or is that not how an original DX/TX behaves?

probonopd:
Change voice name: How do I move to the next character? (I only have one button: the one in the rotary encoder).

arsamus:
>Change voice name: How do I move to the next character? (I only have one button: the one in the rotary encoder).

You have to mantain the button pushed and move the rotary encoder left or right. Same as shortcut, that is not activated on this menu.

probonopd:
Thank you very much @arsamus. You are a true master of the subject! Sorry it took a bit longer, but I have now tested the PR and updated the documentation. Merging.

Thanks for this very valuable contribution! :+1:

PR #271, opened 2022-06-05
Support for a 'back' button, closes #172

probonopd:
As mentioned by @diyelectromusic in

https://github.com/probonopd/MiniDexed/discussions/172#discussioncomment-2820541

> Buttons now renamed; 2-button mode now possible; button logic and debouncing seems to work ok (for me) without interrupts. So it would be really good if someone who knows C++ better than me could take a look and see what they think...

```
# GPIO Button Navigation
#  There are two suggested button schemes:
#    5 Buttons based on an "arrow keypad" with a "home"
#    2 Buttons "back" and "select" for use in addition to a rotary encoder
#
#  Although actually any buttons set to 255 will be ignored
#
ButtonEnabled=1
ButtonPinPrev=16
ButtonPinNext=20
ButtonPinBack=5
ButtonPinSelect=6
ButtonPinHome=12
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,7 +9,7 @@ CMSIS_DIR = ../CMSIS_5/CMSIS
 OBJS = main.o kernel.o minidexed.o config.o userinterface.o uimenu.o \
        mididevice.o midikeyboard.o serialmididevice.o pckeyboard.o \
        sysexfileloader.o performanceconfig.o perftimer.o \
-       effect_compressor.o effect_platervbstereo.o
+       effect_compressor.o effect_platervbstereo.o uibuttons.o
 
 OPTIMIZE = -O3
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -81,6 +81,13 @@ void CConfig::Load (void)
 	m_nLCDPinData7 = m_Properties.GetNumber ("LCDPinData7", 25);
 	m_nLCDI2CAddress = m_Properties.GetNumber ("LCDI2CAddress", 0);
 
+	m_bBTNEnabled = m_Properties.GetNumber ("BTNEnabled", 0) != 0;
+	m_nBTNPinLeft = m_Properties.GetNumber ("BTNPinLeft", 5);
+	m_nBTNPinRight = m_Properties.GetNumber ("BTNPinRight", 6);
+	m_nBTNPinUp = m_Properties.GetNumber ("BTNPinUp", 10);
+	m_nBTNPinDown = m_Properties.GetNumber ("BTNPinDown", 9);
+	m_nBTNPinSelect = m_Properties.GetNumber ("BTNPinSelect", 11);
+
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 9);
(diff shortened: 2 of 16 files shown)


probonopd:
[MiniDexed_2022-06-05-fea86e4](https://github.com/probonopd/MiniDexed/suites/6798325995/artifacts/261011564)

probonopd:
it would be nice if this was integrated more fully,
* `ButtonEnabled` should not be needed as this should never be off
* *Unused buttons should be set to `0` (as we do for other pins)
* `ButtonPinSelect` and `EncoderPinSwitch` seem to be about the same; hence `EncoderPinSwitch` should be no longer used in favor of `ButtonPinSelect`
* It should be possible to also define long press and double click actions

This configuration crashed for me:

```
EncoderPinSwitch=0

ButtonEnabled=1
ButtonPinPrev=0
ButtonPinNext=0
ButtonPinBack=5
ButtonPinSelect=11
ButtonPinHome=0
```

stevelittlefish:
> Thanks @diyelectromusic. Quickly tested it, and I think it would be nice if this was integrated more fully:
> 
> * `ButtonEnabled` should not be needed as this should never be off
> * *Unused buttons should be set to `0` (as we do for other pins)
> * `ButtonPinSelect` and `EncoderPinSwitch` seem to be about the same; hence `EncoderPinSwitch` should be no longer used in favor of `ButtonPinSelect`
> * It should be possible to also define long press and double click actions
> 
> This configuration resulted in a red line in the HDMI log for me:
> 
> ```
> EncoderPinSwitch=0
> 
> ButtonEnabled=1
> ButtonPinPrev=0
> ButtonPinNext=0
> ButtonPinBack=5
> ButtonPinSelect=11
> ButtonPinHome=0
> ```
> 
> Also, while we are at it, maybe we could make each action configurable depending on whether we got a click, double-click,or long press? Something like:
> 
> ```
> ButtonPinSelect=11
> ButtonActionSelect=click
> ButtonPinBack=11
> ButtonActionBack=longpress
> ```
> 
> Wdyt?

I really like these suggestions!  It would let me assign home to long press of the back button.

I also much prefer getting rid of the encoder click and replacing it with the buttons.

@diyelectromusic do you want to make these changes?  If not I'm happy to do it.

diyelectromusic:
> Wdyt?

Happy to go with whatever makes sense with the UI.  It was just a starting point :)

> do you want to make these changes? If not I'm happy to do it.

@stevelittlefish  I won't get to look at it for a while, so go for it!

stevelittlefish:
I have forked here to add my commits: https://github.com/stevelittlefish/MiniDexed/tree/gpio_buttons_v2

probonopd:
Thanks @stevelittlefish, I opened a PR at https://github.com/probonopd/MiniDexed/pull/274 so that we get a binary build that can be tested. You can continue to push to your branch and it will automatically trigger new builds each time you push.

diyelectromusic:
> * *Unused buttons should be set to `0` (as we do for other pins)

By the way, the reason I didn't use 0 is that 0 is a valid GPIO pin (this says 0 to 53 are valid in Circle: https://circle-rpi.readthedocs.io/en/44.4/basic-system-services/gpio-access.html)

But I guess as as it is "reserved for HATs" on a RPi as far as I can see (looks like it is used for the identification stuff), I guess in this context, it shouldn't be configurable in the system anyhow.

Anyway, that's fine if that is the convention :)

stevelittlefish:
> > * *Unused buttons should be set to `0` (as we do for other pins)
> 
> By the way, the reason I didn't use 0 is that 0 is a valid GPIO pin (this says 0 to 53 are valid in Circle: https://circle-rpi.readthedocs.io/en/44.4/basic-system-services/gpio-access.html)
> 
> But I guess as as it is "reserved for HATs" on a RPi as far as I can see (looks like it is used for the identification stuff), I guess in this context, it shouldn't be configurable in the system anyhow.
> 
> Anyway, that's fine if that is the convention :)

That's perfectly logical!  Don't worry, nothing you've done has been wrong.  I think most of the feedback has been either trying to cater to a wider range of use cases or keep things consistent with existing code and convention.

PR #274, opened 2022-06-05
Configurable GPIO buttons

probonopd:
Thanks @diyelectromusic and @stevelittlefish

(This is a continuation of #271 and includes that PR, so only one of the two should be merged eventually.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,7 @@ MiniDexed*
 sdcard
 *.zip
 *.img
+
+# Editor related files
+*.swp
+*.swo
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,7 +9,7 @@ CMSIS_DIR = ../CMSIS_5/CMSIS
 OBJS = main.o kernel.o minidexed.o config.o userinterface.o uimenu.o \
        mididevice.o midikeyboard.o serialmididevice.o pckeyboard.o \
        sysexfileloader.o performanceconfig.o perftimer.o \
-       effect_compressor.o effect_platervbstereo.o
+       effect_compressor.o effect_platervbstereo.o uibuttons.o
 
 OPTIMIZE = -O3
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -81,6 +81,13 @@ void CConfig::Load (void)
 	m_nLCDPinData7 = m_Properties.GetNumber ("LCDPinData7", 25);
 	m_nLCDI2CAddress = m_Properties.GetNumber ("LCDI2CAddress", 0);
 
+	m_bBTNEnabled = m_Properties.GetNumber ("BTNEnabled", 0) != 0;
+	m_nBTNPinLeft = m_Properties.GetNumber ("BTNPinLeft", 5);
+	m_nBTNPinRight = m_Properties.GetNumber ("BTNPinRight", 6);
+	m_nBTNPinUp = m_Properties.GetNumber ("BTNPinUp", 10);
+	m_nBTNPinDown = m_Properties.GetNumber ("BTNPinDown", 9);
+	m_nBTNPinSelect = m_Properties.GetNumber ("BTNPinSelect", 11);
+
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 9);
@@ -178,6 +185,36 @@ unsigned CConfig::GetLCDPinData7 (void) const
 unsigned CConfig::GetLCDI2CAddress (void) const
 {
 	return m_nLCDI2CAddress;
+}
+
+bool CConfig::GetBTNEnabled (void) const
+{
+	return m_bBTNEnabled;
+}
+
+unsigned CConfig::GetBTNPinLeft (void) const
+{
+	return m_nBTNPinLeft;
+}
+
+unsigned CConfig::GetBTNPinRight (void) const
+{
+	return m_nBTNPinRight;
+}
+
+unsigned CConfig::GetBTNPinUp (void) const
+{
+	return m_nBTNPinUp;
+}
+
+unsigned CConfig::GetBTNPinDown (void) const
+{
+	return m_nBTNPinDown;
+}
+
+unsigned CConfig::GetBTNPinSelect (void) const
+{
+	return m_nBTNPinSelect;
 }
 
 bool CConfig::GetEncoderEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -96,6 +96,7 @@ void CConfig::Load (void)
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 9);
+	m_nEncoderPinSwitch = m_Properties.GetNumber ("EncoderPinSwitch", 11);
 
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
@@ -254,6 +255,11 @@ unsigned CConfig::GetEncoderPinClock (void) const
 unsigned CConfig::GetEncoderPinData (void) const
 {
 	return m_nEncoderPinData;
+}
+
+unsigned CConfig::GetEncoderPinSwitch (void) const
+{
+	return m_nEncoderPinSwitch;
 }
 
 bool CConfig::GetMIDIDumpEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -81,12 +81,12 @@ void CConfig::Load (void)
 	m_nLCDPinData7 = m_Properties.GetNumber ("LCDPinData7", 25);
 	m_nLCDI2CAddress = m_Properties.GetNumber ("LCDI2CAddress", 0);
 
-	m_bBTNEnabled = m_Properties.GetNumber ("BTNEnabled", 0) != 0;
-	m_nBTNPinLeft = m_Properties.GetNumber ("BTNPinLeft", 5);
-	m_nBTNPinRight = m_Properties.GetNumber ("BTNPinRight", 6);
-	m_nBTNPinUp = m_Properties.GetNumber ("BTNPinUp", 10);
-	m_nBTNPinDown = m_Properties.GetNumber ("BTNPinDown", 9);
-	m_nBTNPinSelect = m_Properties.GetNumber ("BTNPinSelect", 11);
+	m_bButtonEnabled = m_Properties.GetNumber ("ButtonEnabled", 0) != 0;
+	m_nButtonPinPrev = m_Properties.GetNumber ("ButtonPinPrev", NOPIN);
+	m_nButtonPinNext = m_Properties.GetNumber ("ButtonPinNext", NOPIN);
+	m_nButtonPinBack = m_Properties.GetNumber ("ButtonPinBack", NOPIN);
+	m_nButtonPinSelect = m_Properties.GetNumber ("ButtonPinSelect", NOPIN);
+	m_nButtonPinHome = m_Properties.GetNumber ("ButtonPinHome", NOPIN);
 
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
@@ -187,34 +187,34 @@ unsigned CConfig::GetLCDI2CAddress (void) const
 	return m_nLCDI2CAddress;
 }
 
-bool CConfig::GetBTNEnabled (void) const
+bool CConfig::GetButtonEnabled (void) const
 {
-	return m_bBTNEnabled;
+	return m_bButtonEnabled;
 }
 
-unsigned CConfig::GetBTNPinLeft (void) const
+unsigned CConfig::GetButtonPinPrev (void) const
 {
-	return m_nBTNPinLeft;
+	return m_nButtonPinPrev;
 }
 
-unsigned CConfig::GetBTNPinRight (void) const
+unsigned CConfig::GetButtonPinNext (void) const
 {
-	return m_nBTNPinRight;
+	return m_nButtonPinNext;
 }
 
-unsigned CConfig::GetBTNPinUp (void) const
+unsigned CConfig::GetButtonPinBack (void) const
 {
-	return m_nBTNPinUp;
+	return m_nButtonPinBack;
 }
 
-unsigned CConfig::GetBTNPinDown (void) const
+unsigned CConfig::GetButtonPinSelect (void) const
 {
-	return m_nBTNPinDown;
+	return m_nButtonPinSelect;
 }
 
-unsigned CConfig::GetBTNPinSelect (void) const
+unsigned CConfig::GetButtonPinHome (void) const
 {
-	return m_nBTNPinSelect;
+	return m_nButtonPinHome;
 }
 
 bool CConfig::GetEncoderEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -96,7 +96,6 @@ void CConfig::Load (void)
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 9);
-	m_nEncoderPinSwitch = m_Properties.GetNumber ("EncoderPinSwitch", 11);
 
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
@@ -255,11 +254,6 @@ unsigned CConfig::GetEncoderPinClock (void) const
 unsigned CConfig::GetEncoderPinData (void) const
 {
 	return m_nEncoderPinData;
-}
-
-unsigned CConfig::GetEncoderPinSwitch (void) const
-{
-	return m_nEncoderPinSwitch;
 }
 
 bool CConfig::GetMIDIDumpEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -81,7 +81,6 @@ void CConfig::Load (void)
 	m_nLCDPinData7 = m_Properties.GetNumber ("LCDPinData7", 25);
 	m_nLCDI2CAddress = m_Properties.GetNumber ("LCDI2CAddress", 0);
 
-	m_bButtonEnabled = m_Properties.GetNumber ("ButtonEnabled", 0) != 0;
 	m_nButtonPinPrev = m_Properties.GetNumber ("ButtonPinPrev", NOPIN);
 	m_nButtonPinNext = m_Properties.GetNumber ("ButtonPinNext", NOPIN);
 	m_nButtonPinBack = m_Properties.GetNumber ("ButtonPinBack", NOPIN);
@@ -185,11 +184,6 @@ unsigned CConfig::GetLCDPinData7 (void) const
 unsigned CConfig::GetLCDI2CAddress (void) const
 {
 	return m_nLCDI2CAddress;
-}
-
-bool CConfig::GetButtonEnabled (void) const
-{
-	return m_bButtonEnabled;
 }
 
 unsigned CConfig::GetButtonPinPrev (void) const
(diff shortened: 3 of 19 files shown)


probonopd:
A build is available for testing:
[MiniDexed_2022-06-05-a9e4268](https://github.com/probonopd/MiniDexed/suites/6801783943/artifacts/261263317)

stevelittlefish:
@probonopd I have pushed a working version with all of the suggested features.  It would be good to get any feedback soon as I need to get back to my much more boring (but pays the bills) actual job!

probonopd:
Thank you very much @diyelectromusic and @stevelittlefish. I think many MiniDexed users will be very happy!

There is one issue I found: It breaks [shortcuts](https://github.com/probonopd/MiniDexed/wiki/Menu#shortcuts)

> While you are editing a parameter that applies to one of the Tone Generators (TG1...8) or one of the Operators (OP1...6), you can go from TG1...8 or from OP1...6 quickly by keeping the knob pressed and rotating it at the same time. This allows you to quickly edit the same parameter for multiple tone generators or operators.

To reproduce:

* Go to TG1 -> Voice
* Select a voice
* While staying in the voice selection submenu, press the knob and rotate it to the right while it is pressed. You should be able to select TG2, 2,... TG8

My configuration:

```
ButtonPinPrev=0
ButtonActionPrev=
ButtonPinNext=0
ButtonActionNext=
ButtonPinBack=11
ButtonActionBack=longpress
ButtonPinSelect=11
ButtonActionSelect=click
ButtonPinHome=0
ButtonActionHome=

# KY-040 Rotary Encoder
EncoderEnabled=1
EncoderPinClock=10
EncoderPinData=9
```

stevelittlefish:
The easiest fix would be to completely remove the select button, and reinstate the encoder switch pin.  Then the encoder button would do the shortcut and also act as the select button.

diyelectromusic:
I must admit that I wasn't quite sure why we took out the encoder switch... it seems much simpler to have an encoder mode and a switch mode for most "trivial" use cases but then allowing people to mix and match as they require if they are a little more willing to experiment.

Another minor point is that removing the "enabled" flag also means that the code has to run and check all buttons even when disabled, but at the clock speed of a Pi, that isn't likely to be a problem really I guess.

Just out of interest, did you measure the update period?  I didn't get a chance to.  I see you've added code to only scan if 100uS have passed... this will make the call to update quite "lumpy" but that might be fine (as I said elsewhere, I've not seen any documenting of the scheduler algorithms in play).  If we're worried about updating every time, then different option (one I use in Arduino code quite a lot) would be to spread out the button scanning so it only checks one per call or something... but without measurements, I have no idea what would make a difference or not.  Either way, if you're slowing it down, then I guess that answers the previous polling vs interrupt question! :)

I do worry slightly that the whole single-double-long press thing has made things quite a bit more complicated for users in what was originally quite a simple thing to grasp... but I don't have a strong view either way :)  I do think the default configuration ought to err on the side of simple though.

Aside: why did you rename all the member variables?  I thought they were following the naming convention used elsewhere? Again I don't really have a view, as long as it is all consistent, but I thought I was following this: https://github.com/probonopd/MiniDexed/wiki/Development#naming-conventions - so if I got that wrong, do let me know so I know for next time.

Thanks,
Kevin

sou7611:
For my 2 cents on this, I'd be happy with encoder to change values, encoder
click for enter, and either encoder double click for back, or a separate
button for back (or both)

stevelittlefish:
> I must admit that I wasn't quite sure why we took out the encoder switch... it seems much simpler to have an encoder mode and a switch mode for most "trivial" use cases but then allowing people to mix and match as they require if they are a little more willing to experiment.

It wasn't my suggestion, but it seems nice to only have 1 way of specifying buttons.  I think we should probably add it back in though to support shortcuts, and also because the encoder library has no way to disable it - you have to bind it to a pin.  (I bound it to GPIO0 as this pin is reserved.)

> Another minor point is that removing the "enabled" flag also means that the code has to run and check all buttons even when disabled, but at the clock speed of a Pi, that isn't likely to be a problem really I guess.

If all of the button pins are set to 0 it basically does nothing.  The time to loop through all buttons and check that they are 0 is not going to be an issue, but if it was we could just internally disable the buttons without having to have a separate config option for it.

> Just out of interest, did you measure the update period? I didn't get a chance to. I see you've added code to only scan if 100uS have passed... this will make the call to update quite "lumpy" but that might be fine (as I said elsewhere, I've not seen any documenting of the scheduler algorithms in play). If we're worried about updating every time, then different option (one I use in Arduino code quite a lot) would be to spread out the button scanning so it only checks one per call or something... but without measurements, I have no idea what would make a difference or not. Either way, if you're slowing it down, then I guess that answers the previous polling vs interrupt question! :)

I haven't measured it.  If performance becomes an issue we could spread out the button scanning as you describe, but I'm not convinced it will make a noticeable difference.

The reason I put in the limit was because I had to set the timers to extremely high values to handle long press.  Also, as more stuff gets added to the code base and the update cycle takes longer the numbers may need to be adjusted.  I'm assuming the timer value will not change so will remain consistent.

> I do worry slightly that the whole single-double-long press thing has made things quite a bit more complicated for users in what was originally quite a simple thing to grasp... but I don't have a strong view either way :) I do think the default configuration ought to err on the side of simple though.

I think most of the button options should be commented out so as not to make the initial config more confusing.  Having long press on the back button to return to home is nice though!

> Aside: why did you rename all the member variables? I thought they were following the naming convention used elsewhere? Again I don't really have a view, as long as it is all consistent, but I thought I was following this: https://github.com/probonopd/MiniDexed/wiki/Development#naming-conventions - so if I got that wrong, do let me know so I know for next time.

I actually never saw that page.  I added quite a few variables and then I started getting confused and just changed all of the ones that I kept forgetting how they were spelt!  It was a long day!

Like nPin and pPin - I changed them to pinNumber and pin so I could remember which one was which!

probonopd:
@stevelittlefish that approach you described in the post above sounds good to me as it would combine the best of both worlds if I understand it correctly. Do you think you could implement it? That would be trememdous. I know from other discussions that many users will appreciate this kind of configuration flexibility.

stevelittlefish:
@probonopd I will try to get it done, but my wife literally just gave birth so I'm pretty busy all of a sudden!

probonopd:
Congratulations!

stevelittlefish:
@probonopd I managed to sneak off and implement this.  I did some quick tests but the encoder I'm using is not wired up reliably so it was a little hard to test.  I'm pretty sure it works.

Would you be able to test that it works well for you?

Once everyone's happy with the functionality I will do a bit of tidying up.  We could rename the ini file variable from `EncoderPinSwitch` - just let me know what you think it should be called.

miotislucifugis:
I just tested this using a pi3 and the default minidexed.ini GPIO navigation settings.    "Click= select" and "long press=home" seem to work but "Double click = back" does not work.    Also the press and turn "shortcut" for changing the TG while staying at the same menu level is no longer working.   I havent really been following the development of this branch,   but if the shortcut is no longer supported, it is a real shame because it is really useful.

miotislucifugis:
this is what i have, copied directly from this release:
`# GPIO Button Navigation
ButtonPinPrev=0
ButtonActionPrev=
ButtonPinNext=0
ButtonActionNext=
ButtonPinBack=11
ButtonActionBack=doubleclick
ButtonPinSelect=11
ButtonActionSelect=click
ButtonPinHome=11
ButtonActionHome=longpress
`

stevelittlefish:
@miotislucifugis can you confirm you also have this in your config please?

`EncoderPinSwitch=11`

stevelittlefish:
@probonopd How do I download the build?  I've tested this with my own local build but would just like to double check with the build on github to eliminate any doubt that I'm testing the right thing!

miotislucifugis:
The "long press" seems to have a rather short timer.   I could see myself accidentally triggering a return to the home screen if there is any hesitation when using the press and turn TG short cut.    From a user perspective (this user ;) the 3-click-to-return-home seemed better as it took a more deliberate action and could be preformed just as quickly as a "long press" 
is "three clicks" still a usable config setting?

stevelittlefish:
I redid my hardware setup tonight.  I have added some hardware debounce to the encoder with a low pass RC filter, and also rewired to actually use the encoder button instead of a separate tactile button.  I was unable to double click fast enough with the encoder as it's nowhere near as good as the tactile button I was using previously, so I'm thinking all of my values were too low.  I have therefore adjusted the timing variables and I think this might be the problem.

I've increased the double click timeout from 170ms to 400ms and the long press timeout from 450ms to 1s.

@miotislucifugis would you be able to test again with the latest version and see if it solves the problem for you?  I think it's quite possible that it was timing out too quickly, especially if the encoder wasn't being detected with 100% accuracy.  I don't think the downloadable build will be triggered automatically, maybe @probonopd needs to do something first?

> The "long press" seems to have a rather short timer. I could see myself accidentally triggering an unwanted return to the home screen if there is any hesitation when using the press and turn TG short cut. From a user perspective (this user ;) the 3-click-to-return-home seemed better as it took a more deliberate action and could be preformed just as quickly as a "long press"
is "three clicks" still a usable config setting?

I'd prefer not to add triple click for 2 reasons.  Firstly it complicates the code a little and the timing code is difficult to understand (actually easier to write it than understand it from reading it!) and secondly, it would make the single click less responsive.  The single click has to have a delay long enough to process the double click, and this would be made longer if we introduced a triple click.

Can you see if increasing the long click time makes it more usable for you?

miotislucifugis:
I just gave it a quick test and everything seems much better.  Double click is working... also the "press 'n turn" shortcut.   The response time, especially for the double click, seems excessively "snappy" and quick ( the return function has already happened before I even finish the second click,)   - but thats not not really a problem, just a little disconcerting as it seems like the minidexed is predicting my moves.    
I still think i prefer 3 clicks to long press, but that isnt too big of a deal.   I tried to "accidentally" trigger a return to home page while performing the TG shortcut and I couldnt,  so  "long press" is something i can live with.   As long as I can have single click to "select" (like god intended ;) Im a happy man.  Nice work

stevelittlefish:
@probonopd What do we need to do next to get this merged?

If you also think the double click doesn't feel right, let me know and I'll tweak it.  We could make it so that it doesn't take effect until you release the button after the double click if you think that would be better.

Also if you want me to rename the config variable for the encoder switch please let me know what you want me to rename it to.

probonopd:
Thanks @stevelittlefish. Haven't come around to testing it yet, will do asap.
@all, please post your test results/obervations. Thanks!

miotislucifugis:
Everything seems to be working well for me.  Maybe I would prefer that the double click doesnt register until after the second click is released ( more consistent w/ original behavior and other devices with encoder clicks), but it is really no biggie.   It works fine either way.

stevelittlefish:
> Thanks @stevelittlefish, looking very good now in my tests. Shortcuts are working. What do you think about the following suggestions?
> 
> * longpress event should occur much earlier (currently one has to press much too long; just pressing a tad longer than clicking should trigger a longpress event)

We actually tested it with the longpress being half the current length, but kept triggering it by accident when trying to do the shortcut.  This is especially problematic if the encoder sometimes doesn't register being turned.

> * doubleclick event should only be after the button is released the second time (regarding "It doesn't wait for you to release the button because there are no other possible events at that point", we might want to invent a click-and-longpress in the future...)

Yeah we can definitely do this.  I was starting to think this is better anyway.  As @miotislucifugis it feels like it's predicting the double click which is kind of creepy...

> * Now that we have `ButtonPinBack`, `ButtonPinSelect`, `ButtonPinHome` (which all can be set to the same pin as `EncoderPinSwitch`), why do we still need `EncoderPinSwitch` and what does it even do? Can we remove the setting for `EncoderPinSwitch`?

EncoderPinSwitch defines which pin will do the shortcut!  We can't really remove this but it does need to be renamed.  Do you know what you want this to be called?

My intention was always to rename it  (I suggested it earlier).  I'm not really sure what to call it though.  Something along the lines of ButtonPinShortcut.  I'm not familiar enough with this functionality to come up with a really good name!

> * Would it be possible to have the number of milliseconds configurable for `longpress`? This would allow one to use different length longpresses for different actions, Like so:
> 
> ```
> ButtonPinSelect=11
> ButtonActionSelect=click
> ButtonPinBack=11
> ButtonActionBack=longpress,150
> ButtonPinHome=11
> ButtonActionHome=longpress,1000
> ```

Given enough effort anything is possible :-D

I think it would be a little complicated to do your example above as you could potentially have up to 5 different long presses on a button (I know nobody would realistically do this!) and the event generation code would get difficult to write / understand.

Could we instead just have a single new config variable like `LongPressTimeout=1000`?

probonopd:
> Could we instead just have a single new config variable like LongPressTimeout=1000?

Yes, let's try that.

> Something along the lines of ButtonPinShortcut. 

Let's go with `ButtonPinShortcut`, although I'd have assumed that it is always the same value as `ButtonPinSelect`. Probably this assumption was short-sighted.

> We actually tested it with the longpress being half the current length, but kept triggering it by accident when trying to do the shortcut. 

I would guess that when checking for a longpress, we are currently not taking any rotary encoder movements into consideration. Could we make it so that a longpress is only triggered when there was no movement on the rotary encoder since the button was pressed? That should avoid the issue you describe and allow for very short `LongPressTimeout` values, or am I overlooking something?

stevelittlefish:
> > Something along the lines of ButtonPinShortcut.
> 
> Let's go with `ButtonPinShortcut`, although I'd have assumed that it is always the same value as `ButtonPinSelect`. Probably this assumption was short-sighted.

Actually, thinking about it it would certainly be possible to just always bind it to whatever ButtonPinSelect is.  It just felt a little weird to me if you'd bound select to a double click.  There's not reason why it can't work like that it just felt unintuitive to me.

I think this way is the most flexible, but realistically it could be done either way.  I also like the fact that it's very obviously defined in the config (i.e. it tells you the shortcut is on that pin instead of it being a hidden behaviour you'd have to find in the documentation).

stevelittlefish:
@probonopd I've pushed all of these changes now.  Once you've tested and are happy with everything, let me know and I will remove all of the debug code and tidy some stuff up ready for merging.

probonopd:
Thank you very much @stevelittlefish. Will test now.

stevelittlefish:
> ```
> # Timeouts in milliseconds for double click and long press
> DoubleClickTimeout=400
> LongPressTimeout=200
> 
> # KY-040 Rotary Encoder
> EncoderEnabled=1
> EncoderPinClock=10
> EncoderPinData=9
> ```

One thing to note - LongPressTimeout must be longer than DoubleClickTimeout.  If you set it shorter than DoubleClickTimeout, it will print an error to the HDMI and increase LongPressTimeout to match DoubleClickTimeout.

So in your above example, it will have used 400ms for the Long Press.

probonopd:
Makes sense...
I think functionality-wise this PR is definitely a candidate for merging now.
@fp64lib @craigyjp @diyelectromusic what do you think?

Would anyone volunteer to do a code review? Should be someone more versed in C++ than myself ;-)

stevelittlefish:
Can we get this merged?  It's been sitting here for a while.

probonopd:
Indeed. Looks like in the process of experimenting, I fried the input pins on my Raspberry Pi due to electrical issues on my side, so can't test right now. But it was looking good in my last tests, so if no one objects in the last second, I'll go ahead and merge it.

Would you agree to making this the default?

```
# GPIO Button Navigation
ButtonPinPrev=0
ButtonActionPrev=
ButtonPinNext=0
ButtonActionNext=
ButtonPinBack=11
ButtonActionBack=longpress
ButtonPinSelect=11
ButtonActionSelect=click
ButtonPinHome=11
ButtonActionHome=doubleclick
ButtonPinShortcut=11
# (Shortcut doesn't have an action)

# Timeouts in milliseconds for double click and long press
DoubleClickTimeout=400
LongPressTimeout=400

# KY-040 Rotary Encoder
EncoderEnabled=1
EncoderPinClock=10
EncoderPinData=9
```

stevelittlefish:
> Would you agree to making this the default?
> 
> ```
> # GPIO Button Navigation
> #  Any buttons set to 0 will be ignored
> ButtonPinPrev=0
> ButtonActionPrev=
> ButtonPinNext=0
> ButtonActionNext=
> ButtonPinBack=11
> ButtonActionBack=longpress
> ButtonPinSelect=11
> ButtonActionSelect=click
> ButtonPinHome=11
> ButtonActionHome=doubleclick
> ButtonPinShortcut=11
> # (Shortcut doesn't have an action)
> 
> # Timeouts in milliseconds for double click and long press
> DoubleClickTimeout=400
> LongPressTimeout=400
> 
> # KY-040 Rotary Encoder
> EncoderEnabled=1
> EncoderPinClock=10
> EncoderPinData=9
> ```

I think this is fine - but I haven't really played with it with this configuration.

The way I was using it was to have LongPressTimeout really long (around 800ms), and to use double click for back, and long press for home.

I don't have a strong opinion though - the whole point of this set of changes is that I wanted to have a separate back button! :-)

stevelittlefish:
@probonopd I've pushed a commit with the changes you suggested.

probonopd:
Note to everyone updating to the new version:

__PLEASE NOTE:__ `minidexed.ini` has changed. Please update yours accordingly when updating MiniDexed.
 
* __Default behavior changed__: Click to enter menus, long-press to leave menus, double-click to go to the home screen (configurable)
* Buttons can have click, double click and long press events
* Changed default config, added new options and renamed `EncoderPinSwitch` to `ButtonPinShortcut`

stevelittlefish:
Thanks all for your help getting this tested / ready for merge.

Kevin @diyelectromusic - your first batch of changes saved me a lot of time and head scratching and the discussion around polling vs interrupts was very helpful - definitely changed the way I will think about this going forwards!

probonopd:
Look Ma! No rotary encoder!

diyelectromusic:
Oh, very nice :)  Is that using a ssd1306 display too?

probonopd:
Yes.

probonopd:
What I think would be nice is if the prev/next buttons would continue to decrement/increment values while being hold down. E.g., when you change a numeric value, you would not have to press up to 127 times. (No long-press for these.)

stevelittlefish:
> What I think would be nice is if the left/right = prev/next buttons would continue to decrement/increment values while being hold down. E.g., when you change a numeric value, you would not have to press up to 127 times. (No long-press for these.)

Technically, this is very easy to achieve.

The thing to think about, however, is that it means those buttons behave differently to the other buttons.  This causes 2 issues - first of all there is going to be a different handler for these buttons which means there are now 3 different places in the code where the buttons are handled (encoder, current buttons solution, and a special handler for these buttons).  But perhaps more importantly, the ini file could get a little more complicated as there will be inconsistencies between the different buttons.

Another thing to consider is would you want the (press and hold) shortcut to work with these buttons too, so they can switch between the different channels?

I think it might be worth thinking about where you want to go with the user interface a bit longer term, considering the following:

- This change that you've suggested - holding the buttons to increment the values
- MIDI controller buttons triggering the UI
- The current encoder library is handling one of the buttons and is not ideal (see below)

I personally am not very impressed with the ky040 encoder library - even with hardware debounce on the encoder pins it still doesn't work 100% accurately for me, even though with the exact same circuit with an arduino nano works flawlessly.  The other issue I have with it is that there is no way to unmap the encoder button.  If you set it to 0 in the config it actually binds it to GPIO0.  It's not a huge deal as this pin is not likely to ever be used for anything, but I'd like to be able to turn off the encoder button!

I also find calling the encoder pins "clock" and "data" pretty confusing.

Those are just a few points to think about.  If you want to go ahead with your suggested change I'm happy to do the development work as I'm familiar with the button handling - let me know!

probonopd:
Thanks @stevelittlefish. By now you know this code much, much better than me, so you are in the best position to make technical choices. I will add my 2 cents about the intended behavior:

> it means those buttons behave differently to the other buttons

That would be acceptable for me, as we use only the left/right = prev/next buttons for decrementing/incrementing values, for which we need the press-and-hold behavior.

> would you want the (press and hold) shortcut to work with these buttons too

I would not need the prese-and-hold behavior for the shortcut (while selecting the different TGs), because it only goes through 8 values (whereas where we need the press-and-hold behavior there are 127 values most of the time). But of course the shortcut should still be usable with the left/right = prev/next buttons (for selecting the different TGs). I hope my description is clear, if not please ask.

> I think it might be worth thinking about where you want to go with the user interface a bit longer term

I fear I cannot add much usable information to this discussion, but:

> I personally am not very impressed with the ky040 encoder library - even with hardware debounce on the encoder pins it still doesn't work 100% accurately for me

For me neither - but I don't know how to improve the situation.

> I also find calling the encoder pins "clock" and "data" pretty confusing.

That's actually how the rotary encoder PCBs are labeled, hence the library is using those terms, even though they are technically incorrect.

Sorry I can't be more helpful with the technical choices here, but I believe who implements this is best positioned to make technical decisions. I'd be very happy if you could give it a go @stevelittlefish.

stevelittlefish:
@probonopd I never implemented the shortcut for the buttons at all - it only works on the encoder.  Do we need to add it for the buttons too?

probonopd:
The button in the encoder is also just a button... nothing prevented me from using the same pin for just a plain button - which I did, and it worked for me iirc. Will resume my tests as soon as I have soldered my v2 prototype modules :)

stevelittlefish:
@probonopd I've been super busy with some work / life stuff and not got onto this.  Can we open a new discussion for this set of features?  Do you need me to create my branch first?

diyelectromusic:
For what it's worth I'm having a look at the feasibility of adding MIDI buttons, so in the interests of not having the code diverge too much at the same time, maybe there is no rush for an update right now? :)

probonopd:
Thanks @stevelittlefish, no worries. We are not in a hurry. If you have some time, it'd be super cool if you could send a pull request for "shortcut for buttons" and/or "decrement/increment values while left/right = prev/next are being hold down". 

If you would like to discuss more regarding these features first, feel free to open new issues for them first. 

Thank you for your highly valuable contributions.

sou7611:
Brings whole new meaning to "mini" Dexed

PR #307, opened 2022-07-13
getsysex for mac

sergioska:
wget doesn't work on mac by default, so I have created a simple "decorator" sh command to convert wget in curl. 
it works by applying a simple sed
I hope that can be useful

### DIFF
--- a/getsysex-mac.sh
+++ b/getsysex-mac.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+GETSYSEX_SCRIPT=getsysex.sh
+
+cat $GETSYSEX_SCRIPT | sed 's/wget -c \(.*\) -O \(.*\)/curl -o \2 \1/' | bash


probonopd:
Good idea, but to make it even simpler for users (needing just one script file rather than two) we could just check for the availability of the `wget` and `curl` commands and use whichever is available as the command to download the files with. Do you think you could code that?

sergioska:
Yep @probonopd, curl version should work fine both on linux and mac. let me check it and if it's so I'll convert the main getsysex.sh to use curl instead wget

probonopd:
Not every system has curl installed, and not every system has wget installed. So we should check which one is present and use that one. Ideally all in just one script file.

donluca:
Most Unix distributions now ship with cUrl by default, but implementing a simple check between wget and cUrl would be trivial.

When I'm done with PR #443 I'll tackle this and rewrite the script from scratch to make it more robust and reliable (and easily expandable, so people can add links they wish to include more easily).

This time I know what I'm doing, I've been scripting in bash for years ;D

donluca:
I've created #444 to address this.

Issue #282, opened 2022-06-14
Midi Out will fail when a 0x0A character is in the output.


BenZonneveld:
In the file circle-stdlib/libs/circle/include/circle/serial.h the defaults translate a NL to NL+CR thereby changing the size of voice dumps and transmitting a panning ctrl which also is 0x0A (NL)

// serial options
#define SERIAL_OPTION_ONLCR     (1 << 0)        ///< Translate NL to NL+CR on output (default)

diyelectromusic:
Interesting!  Good spot :)

So it looks like there ought to be a call to m_Serial.SetOptions with a value of 0 (there doesn't seem to be a definition for "clearing options") in serialmididevice.cpp for CSerialMIDIDevice::Initialize do you think?

I can see where the default setting of options is to the above in circle/serial.cpp... but there doesn't seem to be a way to user-set this without using SetOptions that I can see?

BenZonneveld:
I fixed it by changing the default in that file as I saw no other options.
At least my sysex out is working correctly now

probonopd:
Could you please send a pull request/diff? Thanks @BenZonneveld

diyelectromusic:
@BenZonneveld does the proposed fix fix this for you now?

Many thanks,
Kevin

probonopd:
Merged #311 for now so that the fix is in the latest continous builds for easy testing; please let me know if the merge causes any trouble for you. Thanks!

diyelectromusic:
I propose that this issue is now closed as this has now be implemented and as far as I know has solved the problem unless told otherwise.

PR #287, opened 2022-06-19
Sys ex control

BenZonneveld:
Several improvement in sysex control.
All instances get a fixed sysex channel as not to cause any conflicts.

PR #311, opened 2022-07-16
Add call to set Serial options to prevent CR to CRLF conversion

diyelectromusic:
Initial attempt at a fix for https://github.com/probonopd/MiniDexed/issues/282

Grateful if someone could review, test and advise...

Kevin


probonopd:
Ping @BenZonneveld, does this improve things for you?

probonopd:
I see that as part of #287 @BenZonneveld would like to merge a private copy of Circle's `serial.h`, which I would like to avoid if possible, in order not to have the burden of maintenance. Does this PR the same thing that the patched `serial.h` does in #287?

probonopd:
Merged this for now; please let me know if the merge causes any trouble for you. Thanks!

PR #313, opened 2022-07-17
Support more MIDI messages

probonopd:
Port from https://github.com/BenZonneveld/MiniDexed/blob/main/src/mididevice.cpp
Thanks @BenZonneveld


probonopd:
Closing in favor of https://github.com/probonopd/MiniDexed/pull/287

Issue #338, opened 2022-08-23
Might be nice if scrolling through voices automatically moves to next bank


diyelectromusic:
When scrolling through voices on one voice bank, then it might be useful to be able to configure the system so that the next voice bank is automatically selected when then end of one bank is reached, with wrap-around for the last bank back to the first one.  If you see what I mean...

arsamus:
Yes. It would be a good functionality since it is a bit disgusting go back to select bank over and over in order to find the desired voice... 

I could do it, maybe as an optional function setted on Minidexed.ini I'll push a PR soon.

probonopd:
Indeed. It would be welcome.

If I am not mistaken, then @abscisys either had implemented or wanted to implement this in a fork. I think it _might_ even be hidden somewhere in https://github.com/probonopd/MiniDexed/pull/293/files... not sure though. 

@abscisys had [let us know](https://github.com/probonopd/MiniDexed/pull/293#issuecomment-1167732539) that
> Now that I am a Project Manager or even program director of IT projects, I do no longer work on Conf Management system.

diyelectromusic:
Here is my suggested fix for this: https://github.com/diyelectromusic/MiniDexed/tree/Issue338Fix

Grateful if anyone free could have a quick look and see what they think.

Kevin

probonopd:
Never forget to send pull requests :+1: 

https://github.com/probonopd/MiniDexed/pull/392

probonopd:
The advantage of doing a PR is that we get a testable build. Here it is:
[MiniDexed_2022-12-10-11472af](https://github.com/probonopd/MiniDexed/suites/9789395492/artifacts/470798087)

Banana71:
Works perfectly, even the jump from the last bank to the first. Well done

PR #341, opened 2022-08-27
Automatic bank jumping from voice menu

arsamus:
Hi everybody, this PR allows to jump to the next or previous Bank from Voice menu when move beyond the last (32) or the first (1) voice in a continues way. This avoid to go back and forth between Voice and Bank menus when exploring different patches.

This solve #338 posted by @diyelectromusic

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -118,6 +118,7 @@ void CConfig::Load (void)
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
 	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 1) != 0;
+	m_bAutoBankSkip = m_Properties.GetNumber ("AutoBankSkip", 1) != 0;
 }
 
 const char *CConfig::GetSoundDevice (void) const
@@ -358,4 +359,9 @@ bool CConfig::GetProfileEnabled (void) const
 bool CConfig::GetPerformanceSelectToLoad (void) const
 {
 	return m_bPerformanceSelectToLoad;
+}
+
+bool CConfig::GetAutoBankSkip (void) const
+{
+	return m_bAutoBankSkip;
 }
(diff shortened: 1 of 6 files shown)


probonopd:
This seems to work when using buttons to change voices, but seems not to work when using MIDI CCs to change voices. For example, MIDI controllers like the M-AUDIO Keyststation Mini 32 NK3 can be set up so that its + and - keys change voices. This presents somewhat of a dilemma: The CCs don't say "increase/decrease voice", but they send particular voice numbers. So the question is whether we want to do some trickery to interpret e.g., voice 32/"33" on bank 0/"1" as voice 0/"1" on bank 1/"2", and so on... Wdyt?

arsamus:
> The CCs don't say "increase/decrease voice", but they send particular voice numbers. So the question is whether we want to do some trickery to interpret e.g., voice 32/"33" on bank 0/"1" as voice 0/"1" on bank 1/"2", and so on... Wdyt?

Hi @probonopd, yes that is an issue. I've noticed that Program Change signal just send a value between 0 and 127 instead an up or down order. There are a posible solution. It is posible to modify "programchange" method on Minidexed class that instead of set the value sent by the midi controller, jump up one position if the value is increased or jump down one in case the value is decresed. When voice gets the upper value (>32), jump to the next bank and when get 0 jump to the previous one. In this case the controller acts as a "rotary encoder". 

In my synth when you hit "PROG-UP" and the voice number is 127, the counter goes to 0, and when hit "PROG-DOWN" on 0, the counter goes to 127, so the solution may work. I don't if all controllers have the same comportamnt. 

What do you think?

probonopd:
I think it might be worth a try... if it doesn't overly complicate the code and makes it unmaintainable. Maybe it should be possible to disable the behavior?

arsamus:
Ok. Let's try. All feature could be optionally activated via minidexed.ini.

arsamus:
I've just added a Minidexed.ini variable "AutoBankSkip" that switch between original Voice and Bank behavior and Auto Bank jumping behavior.  With this user can choose and set it preferred feature.

probonopd:
If I am not mistaken, then https://github.com/probonopd/MiniDexed/commit/4a8241226d31c749b9e7aee2a2fae0096dd06741 is implementing this as well. Hence closing here. (No particular reason why I merged the other PR instead of this one, so no offense intended.) Thanks @arsamus .

PR #366, opened 2022-10-03
Implement IgnoreAllNotesOff=1

### LINKED ISSUE #356: Some trouble with Classic MIDI In and old Roland Piano
Hello,

i have some troubles with the classic midi input

after releasing a single note, the Roland RD-300 or the master keyboard Roland MKB-300 sends the command "CC: All Note off". MiniDexed then promptly cuts off all tones. The sustain pedal also has no effect. Other devices such as Waldorf Blofeld or the Roland JX-08, JD08 can handle it and can be played well with the RD-300.
Is there a way to filter the "CC All Note Off" command in MiniDexed.

Logfile vom Roland RD-300:
Roland RD-300 Midi In

If it ever works I'll put the MiniDexed in the RD-300 and MKB-300.

probonopd:
Implement `IgnoreAllNotesOff=1` in `minidexed.ini`, closes #356

According to https://community.cantabilesoftware.com/t/roland-d-50-midi-cc-123-all-notes-off/2157/6, certain Roland devices send CC 123 ("All Notes Off") when the last key is released. This is probably a bug or a misinterpretation of the MIDI standard by Roland at the time.

Apparently Roland owners just filter out CC 123.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -69,7 +69,7 @@ void CConfig::Load (void)
 	}
 	
 	m_bMIDIRXProgramChange = m_Properties.GetNumber ("MIDIRXProgramChange", 1) != 0;
-
+	m_bIgnoreAllNotesOff = m_Properties.GetNumber ("IgnoreAllNotesOff", 0) != 0;
 
 	m_bLCDEnabled = m_Properties.GetNumber ("LCDEnabled", 0) != 0;
 	m_nLCDPinEnable = m_Properties.GetNumber ("LCDPinEnable", 4);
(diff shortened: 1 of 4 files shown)


diyelectromusic:
Did you see my note that actually, we ought to be ignoring AllNotesOff when in Omni mode?  If we do that, the filtering probably isn't required, as it would only apply to single channels anyway?

probonopd:
Not sure I understand what you are saying. Let's assume the Roland keyboard is set to Channel 1. Wouldn't we stil have to filter out the AllNotesOff?

diyelectromusic:
> Not sure I understand what you are saying. Let's assume the Roland keyboard is set to Channel 1. Wouldn't we stil have to filter out the AllNotesOff?

Yes, but I think in that case it would work ok as the Roland, although it sends it more than it should, still only sends it when all notes on that channel are really off (as I understand things)?  The problem is when you have several TGs listening to the messages so they all process AllNotesOff even if not actually part of their stream of notes?  I could be mistaken, but that was how I interpreted the problem and what the Roland seems to be doing...?

diyelectromusic:
Ok, I see this has now been merged in. I still think we have a bug here though in that in Omni mode, AllNotesOff should be ignored regardless of any config setting...?  But I'd like someone to check my reading of the MIDI spec!? :)

(and I still wonder then if that would mean this issue wouldn't apply - but if you want a config setting to ignore it too, then ok I guess?)

Kevin

PR #388, opened 2022-12-09
Fix #386

### LINKED ISSUE #386: On power up there are 16 MIDI SysEx Voice dumps sent out to "channel 0x11"
When a ProgramChange is initiated (either over MIDI or via the UI) then the following happens:
* MiniDexed::ProgramChange is called
* which calls MidiDevice::SendSystemExclusiveVoice
* which calls MiniDexed::getSysExVoiceDump and then calls send() for each attached MIDI Device with the resulting data.

getSysExVoiceDump builds a Yamaha specific systems exclusive message that starts with the bytes: 0xF0 0x43 0x0c 0x00 0x01 0x1B where c is the MIDI channel number (in 0 to 15 format).

This means that every time the program is changed, the current voice data for the selected TG is sent out on all MIDI devices.

However, on power up, the initialisation in MiniDexed.cpp calls ProgramChange (0, i) and then later on calls LoadPerformanceParameters which also calls ProgramChange() again.  Both of these calls happen before LoadPerformanceParameters calls SetMIDIChannel to set the MIDI channel for the TG from the performance.ini file.

This means that there are two occasions on power up, for each tone generator, where the MIDI channel is still set to its initialised (i.e. prior to reading the performance.ini file) value of "disabled".  As the Enum for MIDI channel is the following (mididevice.h):
enum TChannel
{
   Channels = 16,
   OmniMode = Channels,
   Disabled,
   ChannelUnknown
};

Then "Disabled" - 1 is 17 or 0x11.  Which is indeed the value we see in the 16 instances of the SysEx voice dump sent out on startup.

Not quite sure what the fix is... something in this chain should probably check for a disabled MIDI channel before dumping out voice data maybe?  I'm not sure.  Also I guess if the MIDI channel for a TG is set to Omni then the voice data will always show "channel 16" ("OmniMode=17" - 1).

I get why ProgramChange always dumps the voice data too, but I can't help think that an editor, if it wants to refresh its data, ought to ask for it rather than "spamming" all MIDI channels ever time a voice is changed...?

And what happens if a sequencer/MIDI file player/etc sends a load of PC messages on all channels at the start? The worst case might be sending PCs on all 16 channels when all 8 TGs are in Omni mode... Won't that cause 128 voice dumps to channel 16 over every MIDI OUT device!?

Thoughts?

probonopd:
Fixes #386
Thanks @diyelectromusic


probonopd:
Please test [MiniDexed_2022-12-09-1460571](https://github.com/probonopd/MiniDexed/suites/9779814170/artifacts/470161023).

diyelectromusic:
Having spent some time today trying to work out how SysEx works for patch dumps, I think this is all ok now.  It would be good for someone else to check over my findings (see the original issue) but I'm happy that the code matches my understanding of how I think it is supposed to work.  Of course, if someone would like to check my understanding too that would be great! :)

probonopd:
Thanks @diyelectromusic. What would be a good description of `MIDIAutoVoiceDumpOnPC` for the documentation?

diyelectromusic:
> Thanks @diyelectromusic. What would be a good description of `MIDIAutoVoiceDumpOnPC` for the documentation?

Yeah... Good question :)

Are all the config options described somewhere (I didn't spot them on the wiki otherwise I'd have updated it myself)?  I found the original point where this was added and there were some words there, but something like:

"When this option is 1 (the default) every program change, either via MIDI or the UI, results in all voice parameter data for the new voice being automatically dumped out as a SysEx message over all MIDI interfaces.  This can be quite useful if using a MIDI linked patch editor to keep the editor display synced up with what is going on with MiniDexed.  If you don't need this or your patch editor will re-request the voice data when a patch is changed, or you'd rather this was manually triggered than automatic, then this can be set to 0 to disable this behaviour."

Personally I would have preferred 0 to be the default, but I thought it more important to maintain compatibility with the previous version of the code, which was doing it automatically....

probonopd:
Why do I see this:

image

Without a button to update this pull request, despite this is set:

image

diyelectromusic:
I see this a the end of this thread of conversation - is that what you are wanting?  Do I push the button?  Is it because this is somehow tied to the branch in my own repository?

As you know I'm always reluctant to work directly on the main repository as git and me rarely seem to get on :)

But if there is something you need me to do, just say so and I'll get on to it.

Kevin

image

probonopd:
Interesting. So you see the "Update branch" button whereas I don't.
Can you please click it?

diyelectromusic:
BTW - did that sort this out?  Can this one move on now?

If we can get some of these accepted, then I'll do some work on some of the other issues related to voices/etc.

Thanks,
Kevin

probonopd:
Unfortunately I still don't have the button... and "This branch is out-of-date with the base branch" again, since the main branch has changed in the meantime.

probonopd:
```
In file included from kernel.h:23,
                 from kernel.cpp:20:
circle_stdlib_app.h: In member function 'virtual bool CStdlibAppStdio::Initialize()':
circle_stdlib_app.h:218:33: error: invalid initialization of reference of type 'CConsole&' from expression of type 'FATFS'
  218 |                 CGlueStdioInit (mFileSystem, mConsole);
      |                                 ^~~~~~~~~~~
```

diyelectromusic:
Consider the button duly pushed once again :)

probonopd:
Thank you very much. Quickly merged it before main goes out of sync again.

PR #392, opened 2022-12-10
Fix Issue #338

probonopd:
Proposed fix for https://github.com/probonopd/MiniDexed/issues/338

### DIFF
--- a/src/sysexfileloader.cpp
+++ b/src/sysexfileloader.cpp
@@ -65,6 +65,7 @@ CSysExFileLoader::CSysExFileLoader (const char *pDirName)
 :	m_DirName (pDirName)
 {
 	m_DirName += "/voice";
+	m_nNumLoadedBanks = 0;
 
 	for (unsigned i = 0; i <= MaxVoiceBankID; i++)
 	{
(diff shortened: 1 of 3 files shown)


probonopd:
Please test:
[MiniDexed_2022-12-10-11472af](https://github.com/probonopd/MiniDexed/suites/9789395492/artifacts/470798087)

Banana71:
Works perfectly, even the jump from the last bank to the first. Well done

probonopd:
Works like a charm.

Should we make it configurable (on by default)?

diyelectromusic:
> Works like a charm.
> 
> Should we make it configurable (on by default)?

I don't think this needs to be configurable?  But happy to be told otherwise.

This feels like the "least surprise" behaviour to me so I imagined this is how people would be expecting the rotary encoder voice selecting to work?

probonopd:
Thank you very much @diyelectromusic

diyelectromusic:
> I found one small issue. When navigating through voices, banks change as expected. However when navigating through banks, the selected voice number is not reloaded from the new selected bank. You then have to exit the bank menu and go back to preset menu and move the voice selection to have the voice changing bank. Not sure I make myself clear so let's imagine:
> 
> * Initial state: MiniDexed is playing Bank 1 / Voice 1
> * Go to voice and move to 2 --> MiniDexed is playing Bank 1 / Voice 2 as expected
> * Go to bank and move to Bank 2 --> MiniDexed is still playing Bank 1 / Voice 2 while it is expected to be playing Bank 2 / Voice 2, don't you agree?

Yes, I agree this is might be slightly odd, but it is also what the original did, so I try to not change additional things as "side effects" of another change.  Basically this is a function of the original BankSelect code.  I don't know what the DX7 does on reception of bank select, so didn't fiddle with it.

But on moving from patch 31 to patch 0 of the next bank, obvs it has to select a new patch at that point.

Kevin

PR #395, opened 2022-12-11
Automatically changing cartridge (bank) when preset <0 or >32

abscisys:
Support MIDI Bank Slect for MSB and LSB
Reload Preset of new bank when bank is changed
Adding much more cartridges in the getsysex.sh script
--> Only tested on RPI 4b


abscisys:
Hi @probonopd, yes most likely, my proposed code also include the support of up to 2^14 instead of 127. It also prepare to respond to MIDI BankSelect MSB and LSB events. Finally it proposes an update to the getsysex.sh script so it prepare a set of 600+ cartridges (= banks)

probonopd:
The points addressed by this PR should now be covered by https://github.com/probonopd/MiniDexed/pull/461. Thank you very much @abscisys

Issue #396, opened 2022-12-11
Ad a "All TG" menu


probonopd:
Add a "All TG" menu in addition to TG 1...8.

Rationale:
Sometimes we want to change an aspect of all TGs at once (e.g., the voice).

Banana71:
like this idea.

What do you think of this menu?
I/O for changing certain TG's

Multi TG Edit 1-8
X X O O X O O X

The settings would change TG 1, 2, 5 and 8

probonopd:
Would be good for power users but probably also be more complicated to use and to code.

@diyelectromusic wdyt?

diyelectromusic:
> @diyelectromusic wdyt?

I must admit, I wasn't entirely sure of this idea at all.  Is this related to the discussion elsewhere about a "unison mode"?  And what do you do if you select "all TG" but the parameter you are changing is different for each one - do they all converge on the same parameter?  If numerical, do they all increment from their own starting point by 1?  This only really makes sense for certain options I'd say, like changing voice.  But if you are wanting all TGs to play the same voice at the same time, then they'd have the same MIDI channel, and so would all respond to a PC message over MIDI?

I guess what I'm saying is that this is probably possible with the right MIDI configuration already, so do we need to try to mimic it in the UI too?  Or if we do, then the idea of configuring a single "unison mode" instrument would probably make more sense to me - so all TGs become "one" as part of some top-level "global setting".  But then I guess even then there would have to be some parameters that wouldn't be the same - otherwise what is the point (e.g. detune)?

I don't know.  I certainly think that "all TG" is probably complicated enough to understand the implications of, so any kind of subset of selected TG editing is getting really, really complicated so personally think I'd recommend we don't go there...

I'm inclined, in general, to treat MiniDexed a bit like the original TX816 - you'd really need to drive it over MIDI for the full power.  MiniDexed already has a much more sophisticated UI compared to three buttons and a 7-segment display, but I think if it gets too complicated then it will become unusable...

Summary: I can see a "make the whole thing act like one 8-way TG" unison mode being useful, but we'd need to think about certain parameters that might need to remain unique, but I'm still not sure about "all TG" but might be convinced... but I really think "some selectable subset of TGs" is too far.

That's my current thinking, but very happy to hear opinions :)

abscisys:
Hello,

[...]

Issue #315, opened 2022-07-18
Unison mode


probonopd:
Looks like, e.g., the Korg opsix has a __Unison__ mode, where you set up one voice, and then just enable Unison with three parameters: 
* How many voices to use (2...8), 
* Spread (how much Stereo Pan to use), 
* Detune (how large the overall range for is for Detune across all voices). 

* See the Unison, Detune, Spread sections in the Korg opsix manual. 

WIth more powerful Raspberry Pis, maybe we can have more than just 8 voices in total? 8 TGs wich can have Unison would theoretically give us up to 64...

probonopd:
"Additional Parameters (ACED Format)" in the Yamaha DX7 II D Supplement Booklet specifies the parameter the DXII uses for Unison, especially PMOD (unison off/on) and UDTN (unison detune depth). No way to set the number of unison voices and the amount of stereo panning?

probonopd:
fwiw, the Korg [opsix native](https://www.korg.com/us/products/software/opsix_native/) VSTi also implements this as described in the original post. Makes a big difference for many sounds!

probonopd:
I can think of two ways of achieving this.

Way 1: Per TG, have one or multiple additional Synth_Dexed instances for the unison effect. Each of the Synth_Dexed instances would run slightly detuned, and the stereo spread would be applied outside of Synth_Dexed.
Pros: Synth_Dexed stays unchanged. Cons: We need many Synth_Dexed instances.

Way 2: Per TG, have one instance of Synth_Dexed, which gets changed to play each note multiple times, slightly detuned and stereo panned. This would mean that Synth_Dexed would need to become Stereo.
Pros: We don't need more Synth_Dexed instances. Cons: Every instance would play more notes in parallel, and would have to be Stereo.

Thoughts?

PR #405, opened 2022-12-20
Implementation proposition for: #315 and #396

abscisys:
- Unison mode #315
- Add a All TG menu #396

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -54,7 +54,8 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	m_bSavePerformanceNewFile (false),
 	m_bSetNewPerformance (false),
 	m_bDeletePerformance (false),
-	m_bLoadPerformanceBusy(false)
+	m_bLoadPerformanceBusy(false),
+	m_bSaveAsDefault(false)
 {
 	assert (m_pConfig);
 
@@ -157,6 +158,14 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	// END setup reverb
 
 	SetParameter (ParameterCompressorEnable, 1);
+
+	#ifdef ARM_ALLOW_MULTI_CORE
+	/* Unison START */
+	SetParameter (ParameterUnisonEnable, 0);
+	SetParameter (ParameterUnisonPanSpread, 60);
+	SetParameter (ParameterUnisonDetuneSpread, 12);
+	/* Unison END */
+	#endif
 };
 
 bool CMiniDexed::Initialize (void)
@@ -412,8 +421,9 @@ void CMiniDexed::SetPan (unsigned nPan, unsigned nTG)
 	assert (nTG < CConfig::ToneGenerators);
 	m_nPan[nTG] = nPan;
 	
-	tg_mixer->pan(nTG,mapfloat(nPan,0,127,0.0f,1.0f));
-	reverb_send_mixer->pan(nTG,mapfloat(nPan,0,127,0.0f,1.0f));
+	float normPan = mapfloat(nPan,0,127,0.0f,1.0f);
+	tg_mixer->pan(nTG,normPan);
+	reverb_send_mixer->pan(nTG,normPan);
 
 	m_UI.ParameterChanged ();
 }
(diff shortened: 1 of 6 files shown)


abscisys:
Addition of a "Unison" menu entry right after the Effects one that contains:

Enable: true/false
All TG: for a global edit of all parameter of all TGs. 2 exceptions: Pan and Detune that both represent the spread of respectively the Detune and Pan params. Note that when enabling Unison, this is TG1 that serves as a template for the other TGs 2 to 8 so that appart from Detune and Pan all the other params of TG1 are copied over to other TGs 2 to 8

abscisys:
Any interest on this one: Unison + AllTGs menu? Feedback would be appreciated.

probonopd:
Sorry for the late replay due to travel.

Currently have 8 TGs that can play different instruments each, OR can be used to play multiple instances of the same instrument, slightly detuned and stereo-panned. Each of those TGs can play 16 notes (RPi1: 8 notes) at the same time (polyphony).
Which means:
8 TGs * 16 notes polyphony = 128 notes in parallel

This PR essentially configures multiple TGs to work together to create the Unison effect more easily (instead of the user having to configure the TGs detune and stereo-pan).

What I'd like to have instead is for each of the TGs to be able to have their own unison effect, possibly at the cost of reducing the number of notes that can be played at the same time, and/or at the cost of having fewer TGs.

For example:
Would it be possible for TG1-4 to have "4-way unison" with with 8 notes polyphony?
4 TGs * 4-way unison * 8 notes polyphony = 128 notes in parallel. And since not all TGs would probably be maxed out all the time, we might possibly not even have to limit the polyphony in practice.

Reading https://www.native-instruments.com/forum/threads/simulating-dx-7ii-unison-poly-mode-closely.70366/ it seems like the DX7II "unison poly" feature was essentially doing something like this, adding the union effect at the expense of limiting the polyphony (instead of using up more than one TG).

Also, modern Yamaha synths are [doing it like this](https://www.yamahasynth.com/learn/yc-series/mastering-yc-menus-vs-settings):

> FM Unison is a cool setting for lead sounds found in many classic FM Synthesizers. Unison adds a thickness and spaciousness to the sound by duplicating notes when a single note is played. This setting is available only FM Voices.
> __Mode__: Selects the number of notes played when a single note is struck. When set to “Off” only one note is played. When set to “2 Unison” the note is duplicated twice; when set to “4 Unison” the note is duplicated 4 times. Default: Off
> __Detune__ Determines the difference in pitch between the two or four notes, which affects the degree of thickness of the sound. Detune adds a cool chorusing effect. Default: 0
> __Spread__: Determines the difference in panning between the two or four notes, which adds a nice spaciousness to the sound. Default: 0

Do you think we could keep our 8 (on lower power RPis maybe 4) TGs, and have 2-way and 4-way unison available for each of them?

abscisys:
Hello @probonopd 
Thanks for your reply! No problem, and ahappy to have your feedback.
I understand the intention with the unison at TG level using polyphony of the TG rather than spreading it across TGs. Only difference with this approach is that the unison will therefore be monophonic while in the proposition it is 16 voices of polyphony. But point taken. Will rework on that.

probonopd:
Hi @abscisys what do you mean by "unison will therefore be monophonic"? It should still be possible to play more than one note at a time?

abscisys:
Ah ok your question makes me understanding your example
> Would it be possible for TG1-4 to have "4-way unison" with with 8 notes polyphony?
4 TGs * 4-way unison * 8 notes polyphony = 128 notes in parallel. And since not all TGs would probably be maxed out all the time, we might possibly not even have to limit the polyphony in practice.

probonopd:
HI @abscisys do you think you could make it happen like in the example? Thanks for your help.

abscisys:
Hi @probonopb,

I need to allocate more time to minidexed I know but I have in a rush on my usual work for now. It should be better after Easter! However I would like to first finish of the work on the fx I started. I have been fixing few bugs but there are 2 that are difficult to find especially as I cannot reproduce what was reported to me!
Right after this I will take care of it!
Except if you would prefer that I implement the unison first!

One side note, the work I do on fx and the unison raises the question related to presets. These capabilities are not existing in dx7 hence the presets require additional parameters that are not in the sysex. For now I save that as part of performance. As it could be ok the fx, it is not for unison. So I was thinking about converting loading Sysex en converting them into new preset format that would support additional parameters. What do you think about that?

Thanks for your feedback

Kind regards

Vincent

Envoyé de mon iPhone

Le 1 avr. 2023 à 12:53, probonopd ***@***.***> a écrit :

﻿

HI @abscisys<https://github.com/abscisys> do you think you could make it happen like in the example? Thanks for your help.

probonopd:
Hi @abscisys. No hurries! Glad that you are still interested in the project.

I think that using performance data for unison and similar effects is ok, like we are doing for reverb.

diyelectromusic:
Hi there.  I've just been having a look at this change.  As I understand things this is I think what it is currently doing:
* Adds a Unison Menu with two items: Enable/Disable Unison mode; All TG Parameter Edit; unison pan/detune settings.
* When Unison mode is enabled it copies all parameters from TG0 over to TG1-7.
* When a parameter is changed under the all TG edit option, it uses a special TG number (100) that is caught in all the appropriate places to "send" the parameter update to all TGs at the same time.
* Unison pan/detune are only partially implemented right now I think?
* Unison settings (on/off, pan, detune) are stored in the performance system.
* Unison and "all TGs" are not recognised as such via MIDI, although enabling Unison mode will result in all TGs having the same channel number, so in principle anything coming in over MIDI should now go to all TGs anyway?

Is that a correct reading of the updates?

A few things I notice:
* I think there is a minor bug when enabling unison mode, the loop includes TG0 in the copy :)
* Am I right in thinking that any original, non "all TG edit", editing will change individual TGs regardless of the Unison setting?  So enabling Unison is more of a "at this point in time" rather than "instrument now in Unison mode"?
* In Unison mode, editing TG0 doesn't replicate to other TGs, you have to use the "all TG edit" menu?

In terms of some of the discussion in this thread - I don't think we should be changing the SysEx format as that is very much "within a single TG voice" remaining compatible with the DX7.  If Unison mode is used, then I think that is ok to be considered a special type of performance as it affects the whole system.

With regards to the idea of increasing TGs at the expense of polyphony, I think that will be quite a complicated thing to manage and implement.  This proposal is a really good first step as it allows the effect at a system level for the MiniDexed to be a whole playable instrument.

In the future, it might be possible to consider "sub TGs" for each "main TG" at the expense of polyphony, but that would be a TG specific setting not a whole instrument setting. It might be something interesting to experiment with.  But I'm not sure why that would be called a "unison mode".  And a good part of me wonders if that could be achieved with some kind of chorus effect, as I think that is kind of what we're talking about I think?

Either way, I think the Unison/all TG changes ought to be completed first, as they are useful and good changes in their own right.  Then we can consider any sub-TG type tradeoffs against polyphony as a new set of changes?

Kevin

diyelectromusic:
Aside: trading polyphony for additional (sub)TGs is probably quite possible, but at the moment this is all fixed at boot time here: https://github.com/probonopd/MiniDexed/blob/582c7407412130fa9fb2f7686c2dd9d98f1e7b47/src/minidexed.cpp#L94

So whilst I think it could be done, at present this would have to be a "reboot to take effect" type of setting I think...  Then there will be lots of places where anything acting on a single TG would have to perform the same action on sub-TGs too.  It might all mostly be constrained to MiniDexed.cpp, but it does do it rather a lot :)

Kevin

probonopd:
Thanks for your review @diyelectromusic.
What I was trying to say is that when I was thinking of "unison", I intended it to work a bit more like on the DX7II - that is, when you enable Unison on TG1 it would not affect TG2-8 (except maybe reducing their polyphony). Maybe we should have a closer look at the DX7II and how it handled Unison, also MIDI-wise?

diyelectromusic:
> Thanks for your review @diyelectromusic. What I was trying to say is that when I was thinking of "unison", I intended it to work a bit more like on the DX7II

Yes I think I know what you mean, hence my comments:
* This should probably be closed off "as is" here, as useful in its own right, but not quite what you were after.
* What you are after I think will be a bit more complicated (as per my second comment).
* There might be a performance hit from what you'd like.
* I'm still not quite clear how it is different to a chorus effect :)

I agree it would be a cool thing to look at, but there might be limitations given the current architecture of MiniDexed if it is possible :)

Kevin

probonopd:
Thanks for the clarifications @diyelectromusic.

> I'm still not quite clear how it is different to a chorus effect :)

[Making Mono Synths Stereo - Chorus vs Double Tracking with Oberheim OB-X8](https://www.youtube.com/watch?v=Z4lbSoUHRlg)

diyelectromusic:
Of course, when availability issues for the Pi eventually subside, there is always the option of several MiniDexeds working together :)  I think one could get quite far with a couple of 3A+s...

(I'm now wondering about the MiniDexed UI generating MIDI CC/SysEx messages that could pass through to another MiniDexed....)

Kevin

Issue #78, opened 2022-04-08
Adding delay FX


dcoredump:
Should I add a delay for each instance?

arsamus:
That would be nice!

Issue #77, opened 2022-04-08
Adding Chorus fx


dcoredump:
Should I add a chorus instance for each TG?

probonopd:
Having a chorus would be great.

I am not sure whether it should be per TG or "global" across everything (like the reverb currently is).
* What did one actually do in the 80s?
* Would it be significantly more CPU load to do it per TG?

rsta2:
Every additional stage in the audio signal chain adds some load. It can only be tested, when it is too much, especially on the RPi 2. We could decide to use `fast=true`, but this only works with a heat sink or fan on some RPi models.

probonopd:
So right now I am slightly leaning toward having effects global across all TGs (like reverb right now), unless someone thinks that is a really bad idea. After all, aren't effects mainly about simulating a room? All instruments are in the same room.

dcoredump:
I am a child of the 80s and I bought in 1987 a DX7II - and still have this nice synth. We used it for several years in our band, but the DX inside its flightcase is very heavy, so I decided to create a lighter version.

> * What did one actually do in the 80s?

I used a seperate FX (Alesis Quadraverb+) for my DX. But I think there is not much change for the FX chain:

Some FX are better used globally (like reverb, master limiter/compressor) and some are better located for each voice (like chorus, flanger, delay). This is not a law of nature but makes sense for most cases.




> Would it be significantly more CPU load to do it per TG?

Yes, sure. But the most CPU intensive FX are currently the compressor and the reverb. A chorus or a delay are more memory consuming but don't eat as much CPU time.

I have started with Synth_Dexed on a Teensy-3.6, which is a 180 MHz ARM Cortex M4 with 256 KB RAM. There was really a problem with CPU cycles and memory. This was solved with the Teensy-4.1 (600 Mhz ARM Cortex M7, 2 * 512 KB RAM). On the T_4.1 I can use 2 * Dexed (16 voices, chorus, delay per instance), 1 * MDA-EPiano (ROMpler, chorus, distortion), 8 Rompler-Drumsounds at a time (from 70 ROM samples) and a global reverb.

So I think: RPi3/4 should manage a delay and chorus for each TG without problems. Perhaps also RPi2. Only a real world test will show if it would do. IF not, it is not very problematic to use the FX as master FX instead of instance FX.

dcoredump:
Forgot to say :)

I have added some more NEON optimization for Synth_Dexed - hopefully this will give some more free CPU cycles.

[...]

PR #409, opened 2023-01-05
FX Rack proposition

abscisys:
Proposition of FX Rack providing a new set of Audio FX:
- Tube saturation simulator
- Chorus #77 
- Flanger
- Orbitone ( = Ensemble)
- Phaser
- Delay #78 
- Shimmer Reverb

All parameters are global and saved as Performance data


abscisys:
Hello and happy new year. I did a bit of work to add this FX Rack during the Christmas break. I would be please to hear what you think about this proposition. Thanks

BobanSpasic:
Is this sending the audio output of **every** TG through the FX?
Shouldn't there be some FXSend parameter (for every FX) in every TG that would set the dry/wet level? Something like the ReverbSend in the already existing configuration.

abscisys:
Yes indeed. That was a big question to me at the start.
The diagram below is the proposed layout and the way it is implement in the present Pull Request
image

Initially I envisaged the below layout that goes in your direction.
image
Then I realized:
- The big number of adds and mults
- The mess to implement in the existing UI -> ~140 new parameters to tweak -> new UI will be required I guess 
- Where shall I put the Compressor -> it should be one per TG -> impact on the Synth structure (or the compressor could also be considered as the other FX unit considering that fact that the feedback is possible)
- The enormous amount of feedbacks will certainly create divergent filter
- Best is the ennemy of good s I decided to give it a try like it is now meaning ultra simplified I agree

What you guys think? Any other simplified proposition?

abscisys:
Boban, I have 2 questions for you:
1) I do not understand why the GPT Commits summarizer / summarize (pull_request) is failing any idea?
2) I made some rapid research to emulate Raspberry PI 0 to 4 on a PC. QEMU seamed promising to me but did not manage to make it work. Do you have a special recipe to emulate Raspberry PI on PC? Or do you test on real HW?

BobanSpasic:
Make 2 FX Slots for every TG (with two FX Send sliders). User can choose which effect goes into slots FX1 and FX2. Make a switch if the Slots are in parallel or in serial connection to each other (e.g. send to chorus and delay separately, or send to chorus and then this wet signal to delay). Include the Reverb to the rest of FX. Leave Compressor as Global FX, just against the distortion.
The wet signals from all the TGs should be added to the dry signals before the compressor.

I think, this way, we would have it like in any common synth.

If this is too much for the RPi, then make a global selection of two FXs (FX1 and FX2) which are then used for all the TGs. TGs should then have two sliders, FX1Send and FX2Send.

BobanSpasic:
> Boban, I have 2 questions for you:
> 
>     1. I do not understand why the GPT Commits summarizer / summarize (pull_request) is failing any idea?
> 
>     2. I made some rapid research to emulate Raspberry PI 0 to 4 on a PC. QEMU seamed promising to me but did not manage to make it work. Do you have a special recipe to emulate Raspberry PI on PC? Or do you test on real HW?

Sorry, I'm no developer here. I just help with the theory as much as I can. My C/C++ is too bad for helping with coding here.
I am the developer for PC software editor/librarian for MiniDexed, coded in Lazarus/FreePascal

abscisys:
Is this what you have in mind?
image

BobanSpasic:
My idea was that FX1 and FX2 can be any of the single effects from your FXRack, PlateReverb inclusive.

abscisys:
> My idea was that FX1 and FX2 can be any of the single effects from your FXRack, PlateReverb inclusive. Example 1: FX1=Flanger, FX2=Delay Example 2: FX1=Delay, FX2=PlateReverb

I see! I need to think about it!!

abscisys:
> btw. which app you use for the diagrams? It would be a good thing to have the diagrams in repository, for documentation, and possibly for Wiki.

I used PowerPoint
[MiniDexed - FXRack layout.pptx](https://github.com/probonopd/MiniDexed/files/10354575/MiniDexed.-.FXRack.layout.pptx)

probonopd:
Very exciting @abscisys. And great remarks @BobanSpasic.
A build for testing is available at [MiniDexed_2023-01-05-d72e204](https://github.com/probonopd/MiniDexed/suites/10191703654/artifacts/498822552). Let's play with it a bit.

Since I am not an audio pro myself, I had to look this up:

image

Source: https://cecm.indiana.edu/361/rsn-sendeffects.html

probonopd:
@abscisys, during testing the following questions came up:

* Which RPi models did you test this on, and with which kind of audio output? On a RPI 3 with i2s DAC, I get lots of noise, distortion, and what seems like audio clipping. I tested this in the default configuration, without changing any parameters. Can you reproduce this? Switching off FXChain makes the noise, distortion, and clipping go away
* Are FXChain settings saved in performance.ini?
* Reverb should be "just another" effect, not a separate menu entry

BobanSpasic:
@probonopd 
Insert effects would be ideal, but that would need an extreme amount of processing power from the RPi.
You need a separate instance of every insertion. Every instance with its own settings. If you decide to have two FX slots for every TG, you need 16 instances of the various effects.

Using Send effects needs less power. If you decide to have 2 Global FX slots in MiniDexed (where you can select any of the effects from the palette and put it in a slot) and just two FXSends for every TG - it is less complicated, and you need just two effect instances.
In the code, FXSend need to be before any Volume attenuation (per TG). That way, you can have dry-only, and wet-only signals.

abscisys:
> @abscisys, during testing the following questions came up:
> 
> * Which RPi models did you test this on, and with which kind of audio output? On a RPI 3 with i2s DAC, I get lots of noise, distortion, and what seems like audio clipping. I tested this in the default configuration, without changing any parameters. Can you reproduce this? Switching off FXChain makes the noise, distortion, and clipping go away
> * Are FXChain settings saved in performance.ini?
> * Reverb should be "just another" effect, not a separate menu entry

@probonopd happy to answer:
* I used RPI4b which is the only model I have unfortunately. I should buy a 3 but they are even more expensive than 4 are nowadays! I quickly recorded a short audio under Cubase with different FX activated that I post here (the sound is pure from Minidexed -> no additional effect applied except the DAC from my Steinberg UR44 soundcard). If you listen you'll hear (Preset 1=Brass   1 with the plate reverb deactivated and with the following FX arrangement (as you can hear I am not a performer. link to the audio file: https://drive.google.com/file/d/1-r0dvei2NfwFTlTkjIudeUeTQ-BgpHL-/view?usp=sharing)
  * Dry
  * Shimmer Reverb
  * Tube > Shimmer Reverb
  * Orbitone > Shimmer Reverb
  *  Tube > Chorus > Delay
* Yes FX params are loaded & saved as part of the PerformanceConfig.
* I fully agree, I just wanted to avoid to be too much impactful as all the FX follow the same structure
image
However I have just completed the migration of the AudioEffectPlateReverb so it implements the reset and processSample methods while keeping the doReverb one so it enables the following design (I am almost almost done the audio part, the UI integration is a pain)
image

abscisys:
> Send need to be before any Volume attenuation (

@BobanSpasic the design I propose allows:
* to send a part of the signal to each indivual FX in parallel
* each FX returns and can send a part of the return to another FX (except it self)
* all TGs and FX return can send part of the signal to the main output

I implement a Dry FX that is the actual MainOutput and that allows to have a pretty simple code for processing this logic of course there will be 1 sample delay between the TG inputs and the FX returns:
```CPP
template<size_t nb_inputs>
void MixingConsole<nb_inputs>::process(float32_t& outL, float32_t& outR)
{
    float32_t fx_inputs_[Ouputs::kNbFX][StereoChannels::kNumChannels];
    float32_t fx_outputs_[Ouputs::kNbFX][StereoChannels::kNumChannels];

    for(size_t i = 0; i < Ouputs::kNbFX; ++i)
    {
        // Compute the samples that will feed the Ouputs and process Ouputs
        fx_inputs_[i][StereoChannels::Left ] = arm_weighted_sum_f32(this->inputs_[StereoChannels::Left ], this->levels_[i], nb_inputs + Ouputs::kNbFX - 1);
        fx_inputs_[i][StereoChannels::Right] = arm_weighted_sum_f32(this->inputs_[StereoChannels::Right], this->levels_[i], nb_inputs + Ouputs::kNbFX - 1);

        // Process the FX
        this->fx_[i]->processSample(
            fx_inputs_[i][StereoChannels::Left],
            fx_inputs_[i][StereoChannels::Right],
            fx_outputs_[i][StereoChannels::Left],
            fx_outputs_[i][StereoChannels::Right]
        );

        if(i != Ouputs::MainOutput)
        {
            // Feedback the resulting samples except for the main output
            this->setReturnSample(
                static_cast<Ouputs>(i), 
                fx_outputs_[i][StereoChannels::Left],
                fx_outputs_[i][StereoChannels::Right]
            );
        }
    }

    // Return this main output sample
    outL = fx_inputs_[Ouputs::MainOutput][StereoChannels::Left];
    outR = fx_inputs_[Ouputs::MainOutput][StereoChannels::Right];
}
```

abscisys:
> 

@probonopd I forgot to mention that my RPI4 is equiped with the RaspiAudio+ v2 interface https://raspiaudio.com/produit/audio

May I ask you how you debug on RPI? Do you have a mean to attach a debugger? Did you manage to virtualize the RPI with Qemu for instance. My attempts do not let boot the VM. I try to pass most of the code through valgrind to chase memory leaks.

BobanSpasic:
Man, that is a lot of sliders.
Each TG has 9 FXSends (incl. Dry output). Each FX has 8 ReturnSends. Every FX is active.
How does this impact the CPU? Does the polyphony for the TGs on the same RPI core suffers?

BobanSpasic:
>     * If you listen you'll hear (Preset 1=Brass   1 with the plate reverb deactivated and with the following FX arrangement (as you can hear I am not a performer. link to the audio file: https://drive.google.com/file/d/1-r0dvei2NfwFTlTkjIudeUeTQ-BgpHL-/view?usp=sharing)
>       
>       * Dry
>       * Shimmer Reverb
>       * Tube > Shimmer Reverb
>       * Orbitone > Shimmer Reverb
>       * Tube > Chorus > Delay
This is evil, pure evil, buhahahaha
I love it 🥇

abscisys:
Thanks @Banana71 !
There are few optimization opportunities that I am looking. Almost done but it will have to wait for the week end I am afraid. The problem I have is that I only have a RPI4b to test and the current version without an audio crack. I agree that there are level differences depending on the FX.

You right about the phaser effects (2 to 24 stages) on ep or strings renders quite nice sounds but my taste goes to the orbitone and the shimmer like reverb.

probonopd:
@rsta2 we have a great contribution here which greatly increases the value of MiniDexed; however the author only has access to a RPi 4. On that model it is working nicely but not on other models. Hence we would like to find out:
* Is this PR indeed requiring so much computational power that it is not possible to use this on lower models?
* Or do we need to do something to utilize the different cores more evenly?
How could we go about testing/experimenting/debugging this?

(RPi 1 and 0 are out of the equation, they certainly don't have enough horsepower, which is OK. We are not offering effects on those models.)

PR #419, opened 2023-01-18
SSD1306 display rotation support

probonopd:
https://github.com/probonopd/MiniDexed/discussions/412

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,11 +20,11 @@ jobs:
       run: |
            set -ex
            git submodule update --init --recursive
-    - name: Use Circle develop branch for WM8960 and i2c display support until it is merged upstream
+    - name: Use Circle develop branch for SSD1306 display rotation support until it is merged upstream
       run: |
            set -ex
            cd circle-stdlib/libs/circle
-           git checkout 646c362 # develop
+           git checkout ec09d7e # develop
            cd -
     - name: Install toolchains
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,9 @@ jobs:
     - name: Use Circle develop branch for SSD1306 display rotation support until it is merged upstream
       run: |
            set -ex
+           cd circle-stdlib/
+           git checkout e318f89 # Needed to support Circle develop?
+           cd -
            cd circle-stdlib/libs/circle
            git checkout ec09d7e # develop
            cd -
--- a/src/Rules.mk
+++ b/src/Rules.mk
@@ -23,6 +23,7 @@ LIBS += \
 	$(CIRCLEHOME)/addon/SDCard/libsdcard.a \
   	$(CIRCLEHOME)/lib/usb/libusb.a \
  	$(CIRCLEHOME)/lib/input/libinput.a \
+	$(CIRCLEHOME)/lib/sound/libsound.a \
  	$(CIRCLEHOME)/addon/fatfs/libfatfs.a \
  	$(CIRCLEHOME)/lib/fs/libfs.a \
   	$(CIRCLEHOME)/lib/sched/libsched.a \
--- a/src/circle_stdlib_app.h
+++ b/src/circle_stdlib_app.h
@@ -1,7 +1,4 @@
 /**
- * This file has been taken from the circle-stdlib project:
- *	https://github.com/smuehlst/circle-stdlib
- *
  * Convenience classes that package different levels
  * of functionality of Circle applications.
  *
@@ -22,7 +19,6 @@
 #include <circle/interrupt.h>
 #include <circle/screen.h>
 #include <circle/serial.h>
-#include <circle/writebuffer.h>
 #include <circle/timer.h>
 #include <circle/logger.h>
 #include <circle/usb/usbhcidevice.h>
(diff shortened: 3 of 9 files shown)


diyelectromusic:
Hmm.  Looks like that file has moved to circle/sound/soundbasedevice.h... not sure why my build didn't have an issue with that... Should be able to change it in minidexed.h

Kevin

probonopd:
Do we need a newer circle-stdlib?

diyelectromusic:
Tbh, I'm not sure how circle becomes circle-stdlib...

But that move happened last April, so its weird we've not picked it up! (see https://github.com/rsta2/circle/commit/7470c48854c54404d04a8e242aac9c4bcf4e0d5b)

Kevin

probonopd:
Well, I am pinning all versions of everything...

diyelectromusic:
I think you'll need to change https://github.com/probonopd/MiniDexed/blob/main/src/minidexed.h#L40

To be "circle/sound/soundbasedevice.h"

probonopd:
Still getting `make: *** No rule to make target 'circle/soundbasedevice.h', needed by 'main.o'.  Stop.`. Not clear how I can fix this.

diyelectromusic:
You've changed that line in minidexed.h and done a full clean build?

diyelectromusic:
Ok, I've fixed the includes, but am getting a whole pile of link errors. Looks like the sound support in circle has changed rather a lot recently!  Still looking...

diyelectromusic:
Right, I've submitted a PR for the update to support the new structure in circle.  Once that is accepted and builds ok, this one should be able to be added on top!

Fingers crossed that will do it.

probonopd:
```
+ make clean
  CLEAN  /home/runner/work/MiniDexed/MiniDexed/src
+ make -j
make: *** No rule to make target 'circle/soundbasedevice.h', needed by 'main.o'.  Stop.
```

probonopd:
@diyelectromusic can you please send one PR that includes both the SSD1306 rotation support and the update to support the new structure in circle? So that the PR builds green. I tried to achieve this by merging some things in the PRs, but I wasn't successful...

diyelectromusic:
I'm really not sure it is advisable to put several changes into a single PR...?  That is why I separated out the fix for issue 421.  Until that change is successfully applied, I don't think we can add other functionality?

It looks like there is some incompatibility between the version of circle-stdlib and circle being pulled into the build, but I don't know how the build setup you have works well enough to know which versions are being used.  When I've tried to run the "git checkout ABCDEF" command locally with the branches you're using, I always get an error...

So whilst I can look at, and fix, the code, this is a library version/github integration thing going on here and I don't really know how to look at it - sorry!

As I say, until this compatibility issue between circle versions is sorted out, I'm not sure it is advisable to do other changes at the same time...?

Can you see if you can just add the 421 fix on its own and find the right versions of circle-stdlib and circle (it should be both develop branches surely?) that work with it?

probonopd:
> I'm really not sure it is advisable to put several changes into a single PR...?

I'd like to have one PR for the rotation feature. And since this feature requires some deps to be updated, those updates should imho be part of the same PR so that we can test the resulting binary before everything is merged.
I don't want to commit the updates to main separately, only to find out that this feature still can't be built.

Right now I am a bit at a loss why I can't this PR to build "green".

diyelectromusic:
Give me a moment - I'm getting there.

My logic for the two PRs is that the "update to the latest develop" version is not related to the rotation, it is something we ought to do regardless - in fact I don't know how we were getting away with it before, as the change happened last April!?

If we decide there is an issue with the rotation, the "update to develop" is still required for any future changes...

But tbh, it is largely overtaken by events now :)

diyelectromusic:
Aside from all this we really need a better way to track build failures against the various versions of the libraries in use so that "updating to new version of library" becomes an independent update in its own right, not something that has to be done as a consequence of a change we want to include for other reasons...

probonopd:
Well, I never update dependencies if there is no concrete need to.

After updating `circle_stdlib_app.h]` we are now getting

```
kernel.cpp: In member function 'virtual CStdlibApp::TShutdownMode CKernel::Run()':
kernel.cpp:89:7: error: 'mbScreenAvailable' was not declared in this scope
   89 |   if (mbScreenAvailable)
      |       ^~~~~~~~~~~~~~~~~
kernel.cpp:91:12: error: 'class CScreenDevice' has no member named 'Update'
   91 |    mScreen.Update ();
      |            ^~~~~~
kernel.cpp: In static member function 'static void CKernel::PanicHandler()':
kernel.cpp:104:15: error: 'class CKernel' has no member named 'mbScreenAvailable'
  104 |  if (s_pThis->mbScreenAvailable)
      |               ^~~~~~~~~~~~~~~~~
kernel.cpp:106:20: error: 'class CScreenDevice' has no member named 'Update'
  106 |   s_pThis->mScreen.Update (4096);
      |                    ^~~~~~
make: *** [../circle-stdlib/libs/circle/Rules.mk:189: kernel.o] Error 1
make: *** Waiting for unfinished jobs....
Error: Process completed with exit code 2.
```

diyelectromusic:
Yes, I don't think you've ported over the changes from our updated circle_stdlib_app.h to use "unbuffered" screens... but as you say, #422 seems to have got all the required bits now!

But as I say, we /really/ need a process for updating the libraries so this doesn't happen again!  And ideally one that can ensure contributors have the required versions in their local repositories too (which I think was my problem in the first place).

probonopd:
I'd say, whenever a PR requires to bump versions of something, then the PR should contain that version bump as part of that PR.
I agree that the `git checkout` we are using is not super clean, but I found no other pragmatic way to update the commits of the dependencies using GitHub.

diyelectromusic:
The problem then though is that someone wants to add one piece of functionality, but then ends up having to regression test a whole pile of other updates and get tangled up in the kind of mess we're having today :)

I think there is a lot to be said for maintaining the linkages to the develop branches outside of other changes so we have a much clearer idea about what is changing and when.  If there is a future issue with circle_stdlib_app for example as a consequence of the changes we're doing right now, it will be very, very confusing trying to work out what the changes had to do with rotating the screen!

But it's your project :)

PR #446, opened 2023-03-03
Add settings to set velocity scale

donluca:


probonopd:
Thank you very much @donluca. Let's test:
[MiniDexed_2023-03-03-5fb24d3](https://github.com/probonopd/MiniDexed/suites/11338518586/artifacts/582615477)

probonopd:
Based on what @Banana71 says, I think we should
* Make this a per-TG setting
* Put it into the menu
* Save it with a performance

donluca:
I have zero clue about how all that works, so I'll prefer to leave it to someone else to expand in the future.

Tomorrow I'll repeat the tests for both my PRs on my RPi2 (it's the only one I have).

donluca:
Tested and working properly on RPi2.

Proof in the picture below.
Upper side is with VelocityScale=1 (DX7), lower side is VelocityScale=0 (OFF).
You can see that with DX7 the quiet notes are louder because the range doesn't start from 0.

Schermata 2023-03-04 alle 15 45 53

probonopd:
@arsamus if we make this configurable per TG and save it with teh Performance in the style of https://github.com/probonopd/MiniDexed/pull/228, will the existing performance files continue to work or will they all be broken?

Banana71:
Hi, 
**test, test,... one, two...., test**
It works, but...
I have already tested a few things but the results are not reproducible. Comparing VelocityScale=0 and VelocityScale=1 you can hear differences, but the effect is not as strong as expected. The best comparison can be made with the TX816 performances that have been uploaded here. https://github.com/probonopd/MiniDexed/discussions/414#discussioncomment-4952011
Left: Full Velocity-----------Right: DX7 Velocity
FULL Velo vs DX7 Velo Elec  Piano small

And as life goes, you test one thing and you find new problems again: The analog inputs of my Behringer UMC1820 have different latencies. After taking that into account, I found that my RPi 2 has shorter latency than the RPi 3. Unusual, how can that be?
More results will follow

donluca:
Please elaborate "the results are not reproducible", because all you have to do is play a note as quiet as you can (midi velocity output = 0) and measure the difference in loudness.
Same with the highest velocity.
If you're using a keyboard, you can hook HDMI up and enable debugging to make sure that the lowest note has velocity 0 and the highest 127.

Or even better, instead of using a keyboard, you can just set up a sequencer to play the same note at various velocity, hook it up to the RPi and record the result.

That way the test is 100% reproducible.

Don't trust your hearing, trust the measurements.

probonopd:
Build for testing:
[MiniDexed_2023-03-29-2d7d1b7](https://github.com/probonopd/MiniDexed/suites/11892734126/artifacts/622346935)

@donluca do you think this is ready for being merged?

@Banana71 are you able to reproduce the results as suggested by @donluca?

donluca:
Yes, this can be definitely merged, @Banana71 did some further testing in a discussion and found out that the change is working properly.

probonopd:
Would you like to suggest some text for the documentation @donluca?
Thanks.

donluca:
The reasoning behind setting a different velocity scale is not really simple or straightforward as discussed in #352 : the FM engine in the DX7 was capable of 0-127, but in reality the keyboard allowed you to only play from 16 to 109 (and DX7II from 6 to 119).
A Youtuber even managed to "glitch" the DX7 by hammering the notes extra-hard and sometimes he could read a note with velocity 120.

One could argue that 0-127 would be the correct setting (and, as a matter of fact, we have left it as default, for everyone's peace of mind), but 16-109 would be the more "authentic" or "accurate" if we're trying to emulate how a real DX7 sounds.

There is one thing which has not yet been implemented and Banana71 requested, which was to make the Velocity Scale different for each TG.
This has not been done and would require a complete rewrite to set velocity scale into performances instead of having it as a global option.
It's up to you whether we want to merge this as it is or wait for someone to rewrite this so that it works inside performances.

Banana71's point was that "recent" patches were most likely made with software like Dexed or other modern synths which expect the full 0-127 velocity, while older patches were obviously made with the original DX7 and so expect 16-109, which means that if you use more than 1 TG with patches made in "different era", than you would have to use 16-109 for the "old" patches and 0-127 for the "new" ones.

My personal take on this: if you wish, we can push this as it is and write just the bare minimum documentation, like "Option to make velocity behave more like a DX7, DX7II or full 0-127", without going into deeper details because people who just want to play around won't probably be interested in this and people who are more deeply involved in FM synths will already know what the option does.

probonopd:
I tend to agree, but eventually it would probably be nice to be able to save this setting for each TG as part of performances. Right?

donluca:
Yes, that was exactly what I was saying and echoes Banana71's thoughts.

Long term this should be rewritten or, if you will, expanded, so that people can set different velocity scales per TG, probably using performances.

probonopd:
Thinking more about it, it might be good if we had a way to say "this bank (`.syx` file) is from the era of DX7, DX7II,...". Maybe we could do this by having additional `.ini` files for each `.syx`. In those, we could store which velocity scale to use for the bank. Overkill?

donluca:
The "proper" way to do this would be to read the header of the sysex file because IIRC it has information on the machine it was made on, but sadly I have no idea on how to implement this.

Banana71:
> The "proper" way to do this would be to read the header of the sysex file because IIRC it has information on the machine it was made on, but sadly I have no idea on how to implement this.

The header is only available once per sysex file, isn't it? If so, it won't work. In many of the banks is a colorful mess and then it would not help. We probably can't get around an additional parameter in the Performance.ini. Or am I wrong?

probonopd:
Well, we could introduce the rule that within one bank you should not mix and match voices that were designed for different machines.

donluca:
Remember that Yamaha might not have made any distinction between the various DX7 revisions/variants (ie: DX7, DX7II, DX7S, etc.) which would result in the same header.

And then it would be up to the various banks' authors to provide a proper header.

I've been thinking about this and unfortunately I don't think it's something we can rely on.

probonopd:
Looks like a setting like this is needed in order not to fall into [this](https://www.youtube.com/watch?v=iHycGwuWcn0) trap.

Build for testing:
[MiniDexed_2023-11-02-3180d78](https://github.com/probonopd/MiniDexed/suites/17865811293/artifacts/1025378357)

probonopd:
@donluca, @Banana71 do you think this is ready for being merged?

probonopd:
We should document

|If the voice was made assuming|and if the master keyboard is sending|then you should set `VelocityScale` to|
|---|---|---|
|0-99|0-99|tbd|
|0-127|0-99|tbd|
|0-99|0-127|tbd|
|0-127|0-127||tbd|

Banana71:
>@donluca, @Banana71 do you think this is ready for being merged?

No, that was just a test.
As Donluca himself wrote (Apr 1): 
>"There is one thing which has not yet been implemented and Banana71 requested, which was to make the Velocity Scale different for each TG.
This has not been done and would require a complete rewrite to set velocity scale into performances instead of having it as a global option.
It's up to you whether we want to merge this as it is or wait for someone to rewrite this so that it works inside performances."

The result was good but not really outstanding.  I suspect that the velocity curve of the DX7 is not linear. Your own keyboard also has an influence on behavior.
The most flexibility would be with 2 adjustable values per TG, velocity max and velocity offset. please refer: https://github.com/probonopd/MiniDexed/discussions/352#discussioncomment-4948552
In addition, you could also create negative touch behavior with two parameters. This means that the harder you hit the keyboard, the quieter the sound.

​<html>
<body>
<p dir="auto" style="box-sizing: border-box; margin-top: 0px !important; margin-bottom: 16px; color: rgb(230, 237, 243); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"></p>


For this miniDexed version I would fill out your table as follows:
If the voice was made for | and if the master keyboard (controller) is | then you should set VelocityScale to
-- | -- | --
DX1/DX5/DX7/TX7/TX816 | DX1/DX5/DX7 | 0
DX7II/TX802 or later | A modern controller sending 0-127 | 2
DX1/DX5/DX7/TX7/TX816 | DX1/DX5/DX7 | 0
DX7II/TX802 or later | A modern controller sending 0-127 | 2


</body>
</html>

`# VelocityScale ( 0=OFF ; 1=DX7 ; 2=DX7II )`

probonopd:
Sorry there was a mistake in my table above @Banana71 (there were duplicated rows), I have updated my table; maybe you can update yours accordingly. Thank you very much!

Banana71:
@probonopd, the table is now updated

probonopd:
I wonder how we could put this into the menu in a halfway intuitive way. Maybe something along the lines of "V: Old, C: Old", "V: Old, C: New", "V: New, C: Old", "V: New, C: New"? (V = Voice, C = Controller)?

probonopd:
@Banana71 I see that this PR is not the final solution for this topic, but it seems like it would be an improvement over not being able to influence this setting at all. So maybe we should merge it but not put it into the menu for now, and note in the documentation that this may change to a per-TG setting in the future. Wdyt?

Banana71:
Yes of course.  This way, someone will be more likely to take up the topic again.  I would like to find out the optimal velocity values, but unfortunately compiling doesn't work for me.  It would be great if you showed me this sometime when we both have time.  🤩

donluca:
@Banana71 brought up a very important point which I completely missed: each MIDI keyboard will have its own MIDI velocity curve and they are all going to be different so it will be nearly impossible to replicate the exact curve of the DX7/DX7II/DX7S, not only because those velocity curves are not currently known, but because, as said, each MIDI controller will have their own.

Still, despite this, I think there's still merit in trying at least to define the lower and upper limit as this will get the user closer to the experience of a real DX7, even if the values in the middle will be all over the place due to different velocity curves.

In a perfect world, in the future there will be a dynamic table which will map the velocity values of various MIDI controllers to the ones of a DX7 to try and emulate as best as it can the velocity curve of the DX7.

Until then, this will do.

I agree not to put this in the menu, but only in the option file and the documentation for people who want to delve deeper.
Default value should be whatever you feel more appropriate. Maybe "OFF".

Banana71:
Hi @donluca, I'm glad it's that you're here again.  

I also believe that Yamaha intentionally tuned the keyboard so that it outputs velocity values between 16 and 109.  The keyboard has been adapted to the sound engine.  Patches with high K.Vel.Sens values cannot be heard at all if you press the keys very quietly.  That's why, in my opinion, velocity values are always at least 16.

donluca:
Unfortunately I'm just passing by, I'm busy with other things but hopefully one day I'll return to this project to contribute more.

I have no idea about what Yamaha engineers had in mind and, if you remember, there was a Youtube video were a guy managed to "glitch" his DX7 by pressing really really hard on his keys and squeeze a slightly higher velocity, so it's really anyone's guess.

I'd still stand by the very in depth research he did and stay our course with the values he found.

PR #449, opened 2023-03-04
Multi-FX update 1

abscisys:
I propose this update of the initial Multi-FX pull Request I posted.

This update includes :
- fixes for Tube: revisit the algorythm based on vacum tube response and saturation
- fixes for Delay: stereo issue fixed and noises
- fixes Shimmer: noise due to bad variable initialization and stereo issues
- Many FXEngine simplification and method addition to reduce CPU usage drives to impact on Chorus, Flanger, Orbitone and Shimmer
- optimizations to reduce resources consumption: reduction of multiplications and change the implementation  of LFOs
- introduction of a unit test suite running on Linux / WSL using gtest library. Test results are posted [here](https://drive.google.com/drive/folders/1Ex1EVszDvnpYR3r8DrCnj0CW6O3EWJ3h?usp=sharing)


probonopd:
Thank you very much. On which Raspberry Pi models could you test this so far?

abscisys:
Only on RPI4b. Considering that it runs on bare metal do you have a specific process to measure the resource consumption or emulate the RPI board / debug remotely?

probonopd:
Hi @abscisys I am not aware of RPi emulators. My method so far has been trial and error. We can only merge this once it is either working properly on all RPi models or we disable it (or some parts of it) on those models that can't handle (all of) it. Thanks for your understanding.

Banana71:
Hi,
thanks a lot @abscisys 
I tested with RPi 3 model B+ and RPi 2 model B, it basically works on both devices.

But be careful, the flanger effect produced **very loud** clipping on both devices.
@abscisys, can you please deactivate the effect if there are no parameters for the FXChain in the Performance.ini. Otherwise, the flanger sounds very good, but I disabled it to protect my hearing.

- chorus: nice stereo effect, parameter depth 0 - 10 effective, 11 - 99 like position 10. The function of parameter depth is more like stage.

- Orb: mono chorus, maybe something for bass sounds, the effect is elegantly reserved

- Phaser: very nice effect, also good as a chorus

- Delay: Stereo delay effect, does what it's supposed to, very well.

- Shimmer: Overdrive reverb, delay effect. I don't understand the effect, maybe a high-pass filter is missing. But it work.

The effects work very well individually but also in combination and are an improvement for the miniDexed. The effects do not affect the latency.

I have a question about the routing: is it possible for each TG to have its own FX send for each individual effect or, if there is no other way, an FX send for the entire FXRack?
I am very happy that the RPi 2 works so well with it.

probonopd:
That's great news!

abscisys:
Thanks @Banana71 for your feedback

That is a good news RPi2&3. 

Now related your feedback:

- Flanger: understood very strange as I never encounter such issue during any of my tests. Isn't it Tube that is loud? -> will have a look into it. Note that FX params are saved as part of the performance:
`
# FXRack
FXChainEnable=1
FXChainWet=99
FXChainTubeEnable=0
FXChainTubeWet=50
FXChainTubeOverdrive=10
FXChainChorusEnable=0
FXChainChorusWet=50
FXChainChorusRate=40
FXChainChorusDepth=50
FXChainFlangerEnable=0
FXChainFlangerWet=50
FXChainFlangerRate=3
FXChainFlangerDepth=75
FXChainFlangerFeedback=50
FXChainOrbitoneEnable=0
FXChainOrbitoneWet=80
FXChainOrbitoneRate=40
FXChainOrbitoneDepth=50
FXChainPhaserEnable=0
FXChainPhaserWet=50
FXChainPhaserRate=5
FXChainPhaserDepth=99
FXChainPhaserFeedback=70
FXChainPhaserNbStages=12
FXChainDelayEnable=0
FXChainDelayWet=50
FXChainDelayLeftDelayTime=15
FXChainDelayRightDelayTime=22
FXChainDelayFeedback=35
FXChainShimmerReverbEnable=1
FXChainShimmerReverbWet=70
FXChainShimmerReverbInputGain=55
FXChainShimmerReverbTime=75
FXChainShimmerReverbDiffusion=75
FXChainShimmerReverbLP=80
`
- Chorus: one of the first FX I fixed for this update I miss so test scope here I believe -> will increase auto regression tests to cover param variations
- Delay implementation simulates a Tape Delay and can add fluttering effect using a perlin random noise on the delay time and on Low and High Pass filter. These params are hidden for the moment. Any interest?
- Shimmer: I rather should rename it as Reverberator. I changed the implementation of the filter compared to the previous version. This new uses a ported implementation from Mutable Instrument. I will re-test the implementation.
- Routing: I am preparing an update that integrates the Plate Reverb as the other FX and include a complex routing mechanism with the possibilities

Banana71:
> * Flanger: understood very strange as I never encounter such issue during any of my tests. Isn't it Tube that is loud? -> will have a look into it. Note that FX params are saved as part of the performance:
>   `

For me it was actually the flanger that caused the problems.  Mostly during breaks in playing after 20-30 seconds, a very loud, distorted feedback oscillates, which disappears again after about 5 seconds.

There are already a lot of Performances that are still saved without the parameters for the FXRack.  If there are no parameters saved in performance.ini for the FXRack, the effects should also be muted.

I forgot the Tube 🫣
- Tube: The tube always works with maximum distortion, with the parameter Depth you only change the threshold value from which input volume the effect becomes active.  The parameters are probably assigned incorrectly somehow

> Routing: I am preparing an update that integrates the Plate Reverb as the other FX and include a complex routing mechanism with the possibilities
>   

That would be great

probonopd:
> complex routing mechanism

Hopefully "mere mortals" will still be able to understand it. At least we need to document it very clearly. Not that it ends up as complex as JACK, which always wants to be configured in strange ways before you hear any sound.

Banana71:
> Hopefully "mere mortals" will still be able to understand it. At least we need to document it very clearly. Not that it ends up as complex as JACK, which always wants to be configured in strange ways before you hear any sound.

That would be the time to think about presets for the effects

probonopd:
@abscisys how possible or impossible do you think your code would be to port to JUCE so that we could have it in [MultiDexed](https://github.com/probonopd/MultiDexed) (which at some point in time might eventually become MiniDexed's performance compatible on-PC cousin)?

abscisys:
I do not know JUCE framework other than by its name and how popular it is in the VST industry! But at a first sight I would not see any problem to port it in the PC VST world at least if I had to do it in the native VST SDK environment!

probonopd:
> I compile it on x64 already

Interesting, tell me more about it. What exactly are you compiling there, and how do you test it?

Issue #460, opened 2023-03-31
Expand number of banks to 16384 and implement MIDI Bank Select MSB/LSB


diyelectromusic:
From: _Originally posted by @abscisys in https://github.com/probonopd/MiniDexed/issues/395#issuecomment-1349854379_

Extract the MIDI Bank Select LSB/MSB code from this PR and expand the number of supported banks to 16384 (i.e. 0 to 16383) by increasing MaxVoiceBankID.

Issue #458, opened 2023-03-31
Bank menu has blank entries for unloaded banks


diyelectromusic:
From: _Originally posted by @donluca in https://github.com/probonopd/MiniDexed/issues/455#issuecomment-1484175202_

Blank entries, gaps, or invalid banks should be skipped in the UI but at present it shows blank entries.

donluca:
I'm working already on automatically skipping empty voices.
I think I can use the same approach (read the bank/voice name) for empty banks.

diyelectromusic:
Ok, for ease of integration I would recommend that each fix should have its own PR - this was one of the problems with some of the outstanding PRs in the list - there were a lot of changes in each PR making it hard to integrate them.

What else are you working on, so I'm not wasting my time :)

Kevin

donluca:
Yeah, that's what I've been doing if you look at my PRs: each PR has a single "fix" or improvement.

To be honest I'm making some changes mainly for myself: I've merged #395 so that I can scroll beyond 127 banks and I've made a change in the UI so that in the voice selection you can see the name of the bank at the top (trimmed if longer than 16 characters).
I'm not sure people would agree with those kind of changes, so I've made a [private branch](https://github.com/donluca/MiniDexed/commits/private) for myself and will create separate branches for things I'd like to see merged into MiniDexed.

Right now I'll just be tackling the empty voices thing, if you want to take care of the empty banks, the floor is yours.

diyelectromusic:
Ok :)
In that case I'll kick off a series of commits in my own branch for some of these minor fixes without changing too much at a time, so hopefully we can converge on something useful!  Assuming @probonopd agrees of course.

diyelectromusic:
Ok, so Bank Select for > 127 banks (via menu and MIDI) is now in this branch: https://github.com/diyelectromusic/MiniDexed/tree/bankhandling

I don't suppose you fancy giving it a go?

I'll get going on some of the other changes if that all seems ok.

Kevin

donluca:
I'll give it a go this evening if I have a bit of time.

It would greatly help if you could provide a build (just kernel7.img suffices, as I use a RPi2), unless you want me to try and merge this into my private branch... ?

probonopd:
Build for testing:
[MiniDexed_2023-03-31-36097b0](https://github.com/probonopd/MiniDexed/suites/11955082990/artifacts/626889921)

donluca:
Can confirm this is working as expected.

Issue #457, opened 2023-03-31
m_nNumLoadedBanks is incorrect if there are gaps in the bank numbering.


diyelectromusic:
From: _Originally posted by @diyelectromusic in https://github.com/probonopd/MiniDexed/issues/455#issuecomment-1484234905_

m_nNumLoadedBanks should reference the highest numbered bank loaded, not the number of banks loaded.

Kevin

Issue #455, opened 2023-03-25
Not all sysex files get loaded


donluca:
A huge amount of sysex files which are correctly recognized and loaded on Dexed are being written off as "not supported" in MiniDexed.

I got a [big (3000+) collection of sysex files for the DX7](https://audiosex.pro/threads/dexed-fm-plugin-synth-3650-syx-banks-from-the-yamaha-dx7-27-mb.59385/) and loaded them onto MiniDexed but only ~80 of them are actually getting loaded.

I've quickly tried them in Dexed and most of them are correctly recognized and loaded.
Others give a warning about not recognizing them and giving the option to load as "random data", but it actually correctly loads them.

EDIT: here's the entire collection converted to the MiniDexed format, ready to go:
[sysex.zip](https://github.com/probonopd/MiniDexed/files/11069176/sysex.zip)

donluca:
I think I've found (one of?) the culprit: [this line here](https://github.com/probonopd/MiniDexed/blob/fcea41ead3fadd962342ffdf8241e48c1f84358e/src/sysexfileloader.h#L32) limits the number of patches to load to 127.

What happens probably is that MiniDexed hits this limit and discards everything else.

@probonopd is there any particular reason why this limit was set to 127?

Banana71:
Isn't it so that all sysex files are loaded into ram at startup?

donluca:
Are there ways to manually load sysex files manually?

I've changed that line and now it is correctly loading all the sysex files but it is taking literally forever.

I'd prefer to load them "on demand" rather than loading them all at startup.

It is loading 4 sysex patches per second which means that to load all of my 3200 patches it is going to take a whopping 800 seconds, more than 13 minutes!

Banana71:
I prefer a fast system start. ⌛😴💤

donluca:
I don't mind waiting something like 10-15 seconds, but 9 minutes is just nuts and, on top of that, I can access only 129 Banks of the 3000 loaded?

Something is fishy...

probonopd:
> I think I've found (one of?) the culprit: [this line here](https://github.com/probonopd/MiniDexed/blob/fcea41ead3fadd962342ffdf8241e48c1f84358e/src/sysexfileloader.h#L32) limits the number of patches to load to 127.
>
>What happens probably is that MiniDexed hits this limit and discards everything else.

Indeed.

> @probonopd is there any particular reason why this limit was set to 127?

[...]

PR #461, opened 2023-03-31
Fix some of the issues with bank handling

diyelectromusic:
This looks at the following issues: #457 #458 #460 as part of addressing some of the issues raised in #455.  Each issue is a separate commit so could be handled independently if required.

This should now support:
* Up to 16384 banks (numbered 0 to 16383) via the user menu and MIDI BANK SELECT messages.
* Blank entries for unloaded or invalid banks should be skipped in the menus and ignored over MIDI.
* Selecting banks should now usefully (hopefully) wrap around in the menus.

If this seems ok for people, then I'll start on some of the other issues around bank selection that have been discussed.

Kevin

### DIFF
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -36,7 +36,7 @@ LOGMODULE ("mididevice");
 #define MIDI_AFTERTOUCH		0b1010			// TODO
 #define MIDI_CHANNEL_AFTERTOUCH 0b1101   // right now Synth_Dexed just manage Channel Aftertouch not Polyphonic AT -> 0b1010
 #define MIDI_CONTROL_CHANGE	0b1011
-	#define MIDI_CC_BANK_SELECT_MSB		0	// TODO
+	#define MIDI_CC_BANK_SELECT_MSB		0
 	#define MIDI_CC_MODULATION			1
 	#define MIDI_CC_BREATH_CONTROLLER	2 
 	#define MIDI_CC_FOOT_PEDAL 		4
@@ -281,6 +281,10 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 		
 						case MIDI_CC_PAN_POSITION:
 							m_pSynthesizer->SetPan (pMessage[2], nTG);
+							break;
+		
+						case MIDI_CC_BANK_SELECT_MSB:
+							m_pSynthesizer->BankSelectMSB (pMessage[2], nTG);
 							break;
 		
 						case MIDI_CC_BANK_SELECT_LSB:
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -365,14 +365,44 @@ CSysExFileLoader *CMiniDexed::GetSysExFileLoader (void)
 	return &m_SysExFileLoader;
 }
 
+void CMiniDexed::BankSelect (unsigned nBank, unsigned nTG)
+{
+	nBank=constrain((int)nBank,0,16383);
+
+	assert (nTG < CConfig::ToneGenerators);
+	
+	unsigned nHighestBank = GetSysExFileLoader ()->GetNumHighestBank();
+	
+	if (nBank <= nHighestBank)
+	{
+		m_nVoiceBankID[nTG] = nBank;
+
+		m_UI.ParameterChanged ();
+	}
+}
+
+void CMiniDexed::BankSelectMSB (unsigned nBankMSB, unsigned nTG)
+{
+	nBankMSB=constrain((int)nBankMSB,0,127);
+
+	assert (nTG < CConfig::ToneGenerators);
+	unsigned nBank = m_nVoiceBankID[nTG];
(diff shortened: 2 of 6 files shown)


probonopd:
Build for testing:
[MiniDexed_2023-03-31-36097b0](https://github.com/probonopd/MiniDexed/suites/11955082990/artifacts/626889921)

diyelectromusic:
This build seems to work fine for me (RPI 3, approx 30 odd banks, with some gaps, large numbers, and one invalid file).  MIDI selecting working fine for me from MidiOx too.

Kevin

probonopd:
So this is addressing what https://github.com/probonopd/MiniDexed/pull/395 also (partly) addresses?

diyelectromusic:
> So this is addressing what #395 also (partly) addresses?

Yes.  Support for large numbers of banks was already in but #395 started to add MIDI MSB/LSB handling, although it wasn't fully complete.  The wrap-around was also not taking into account the actual number of loaded banks, so again that is now fixed.

Basically there has been a fair bit of discussion since that first PR and I think this is where we got to with it all.

Kevin

diyelectromusic:
Very probably, yes, it will only scroll through loaded banks now.  I guess we could leave a blank bank in there as an option or give the option in the menu to "create new bank" or something like that?

Kevin

donluca:
Having a blank bank filled with ```USER VOICE``` would be brilliant as it would indeed solve the issue discussed at #454 and it would leave the possibility for users to create their own patches.

We need to make sure that the ```USER VOICE``` bank is accessible not only by reaching the last bank (in an SD card with thousands of sysex files it would take forever), but also by "wrapping around" when going from the first bank backwards, so that it is just one anti-clockwise turn of the knob away.

The issue is coding something in a way that the moment that the ```USER VOICE``` bank gets one voice written to it, a new one becomes available so that when the user fills up the 32 voices he already has another 32 available (and so on).

I honestly have no idea how that blank bank mentioned in the wiki was coded.

diyelectromusic:
I don't think we need to do anything in code - I think just let the user decide how many "user banks" they want by copying the file that many times?

Basically we just give them a bank file.  Once it is loaded, the system will treat it as a valid bank (even if blank) and as long as the voices are called something other then INIT VOICE it should all "just work".

Then the user could decide to have user banks at the start, after the official ROMS, at the end, or wherever.

This mirrors my experiences with other synths (including some DXs) where you have default patches in some banks and then "internal" or "user" or "program" or "cartridge" banks for user edits.

We just need to work out how to create the empty/default bank of 32 voices with an appropriate name!

BTW "NO NAME" was just the default returned when CSysExFileLoader::GetBankName can't get a valid bank filename from the loaded bank structure.  I believe it is only accessed from the UI - it doesn't seem to be an actual value in the system itself anywhere as far as I can see.  It seems to rely on the m_BankFileName[] structure being initialised by the compiler to something that isn't obviously a string... :/

Kevin

Banana71:
My first thought 

Bank:
mDx RAM A 
mDx RAM B
....

INIT Voice:
mDx INIT

donluca:
So in the getsysex.sh file I should add an empty sysex file at the end?

Wouldn't it be better if we had a menu item ```Create blank bank``` which would create a blank bank at the end of the banks?
This way they can create as many as they wish.

diyelectromusic:
You know what... looking through the code, I can't spot any way to actually save voice parameters?  So most of this might be a little academic!  It may well be that hand-edited voices are only in memory until power off!!

The only saving seems to be for performances, but that doesn't seem to include voice data either I think?

Kevin

donluca:
I admit I have zero experience in making patches from scratch straight on the Minidexed, so I really can't help you here.

I just wanted to point out that a feature which is mentioned in the wiki is not there anymore, so I think it would be best to find a way to implement it back one way or another (and possibly be consistent with the directions in the wiki).

Banana71:
> But on saving, even though bank and voice number are saved in the performance, it also includes a Voice data dump. This means that for any of the performances reloaded after saving then really the bank and voice numbers are irrelevant - it doesn't matter what they say the voice used will be that encoded in the voice data.

This is one of the best features of the minidexed.  In the Performances I create, almost every Voice has been customized.  Every performance works on every miniDexed and you don't have to worry about the sound banks.  However, it would be fantastic to be able to store a voice directly into a sound bank.

diyelectromusic:
I don't doubt it, but I'm just pointing out we've introduced weirdness in the mental model around what is a voice, what is a bank, and what is a performance. It feels like it is now a bit of a mess...

Not sure what to suggest tbh, but at the very least some of this ought to be documented somewhere :)

I guess as long as someone either works with individual TGs and banks/voices or performances, they'll probably be ok... but the idea of a performance saving the state of the system almost, but not quite true :) It is really a new level of synth voice with eight tone generators at its command and the loaded banks/voices are just used as "initialisers" for the voices for each TG.

Also, if someone is only working in performances, then having banks loaded is a bit of a waste of resources as they'll never be used.  But I guess that isn't a big deal.

Kevin

Issue #459, opened 2023-03-31
MiniDexed should support "headerless" SysEx voice files


diyelectromusic:
From: _Originally posted by @diyelectromusic in https://github.com/probonopd/MiniDexed/issues/455#issuecomment-1484230482_

It should be possible for MiniDexed to load headerless SysEx voice files by checking the size of the SysEx file and providing a dummy header when required.

PR #463, opened 2023-04-01
Support for headerless SysEx Voice Banks

diyelectromusic:
Fix for Issue #459 to allow for the use of headerless SysEx voice banks.

Note: this is not the most efficient implementation as it will attempt to read a voice bank as a headered version first and when that fails will retry the headerless version.  Ideally the filesize would be checked first (headerless are size 4096 compared to headered 4104) but that doesn't seem straight forward in the circle environment to do that, that I can see.  I've asked on the circle forums so if I get a better answer, this can be updated.

This is configuration with the option:
HeaderlessSysExVoices=0
in minidexed.ini.  Set to 0 (default) to leave disabled.  Set to 1 to allow.

Also note: with a headerless file there is no real checking that it is a valid (Yamaha/DX7) voice bank.  If this option is used then all responsibility is on the user to provide sensible files.  The only checking is that there is enough data to fill the "bank".

Kevin

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -71,6 +71,7 @@ void CConfig::Load (void)
 	m_bMIDIRXProgramChange = m_Properties.GetNumber ("MIDIRXProgramChange", 1) != 0;
 	m_bIgnoreAllNotesOff = m_Properties.GetNumber ("IgnoreAllNotesOff", 0) != 0;
 	m_bMIDIAutoVoiceDumpOnPC = m_Properties.GetNumber ("MIDIAutoVoiceDumpOnPC", 1) != 0;
+	m_bHeaderlessSysExVoices = m_Properties.GetNumber ("HeaderlessSysExVoices", 0) != 0;
 
 	m_bLCDEnabled = m_Properties.GetNumber ("LCDEnabled", 0) != 0;
 	m_nLCDPinEnable = m_Properties.GetNumber ("LCDPinEnable", 4);
@@ -177,6 +178,11 @@ bool CConfig::GetIgnoreAllNotesOff (void) const
 bool CConfig::GetMIDIAutoVoiceDumpOnPC (void) const
 {
 	return m_bMIDIAutoVoiceDumpOnPC;
+}
+
+bool CConfig::GetHeaderlessSysExVoices (void) const
+{
+	return m_bHeaderlessSysExVoices;
 }
 
 bool CConfig::GetLCDEnabled (void) const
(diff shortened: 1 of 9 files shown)


probonopd:
Thanks @diyelectromusic. Which use cases do you see for this? Why not "just" fix the banks in question?

diyelectromusic:
The use case is someone who has a lot of banks in the second format (almost every collection I've downloaded has some).  Hex editing banks to add 6 bytes at the start and 2 at the end isn't an easy thing to do and voice collections seem to use both formats, so seemed like a nice thing to add.

The reason it was in a separate PR is that it can be ignored if not wanted.

Kevin

probonopd:
> The use case is someone who has a lot of banks in the second format (almost every collection I've downloaded has some). Hex editing banks to add 6 bytes at the start and 2 at the end isn't an easy thing to do and voice collections seem to use both formats, so seemed like a nice thing to add.

Would it be a useful thing to write a small utility that does just that? It could be run before copying the banks to MiniDexed.

diyelectromusic:
Up to you, but I tend to subscribe to the "be liberal in what you receive, and strict in what you send" principle of Internet protocols in things like this. If it doesn't cost us much (it doesn't) to be a little more flexible, then why wouldn't we?

The performance hit is only if you are loading headerless or invalid banks.  There is no (as far as I'm aware) hit for "proper" banks.

The "header" versions we already accept are actually complete SysEx messages rather than just voice dumps anyway, so arguably, you could say they were the anomalies as they contain superfluous MIDI message data in them in the first place :)

In short, I don't have a major issue either way, but am struggling to think of a reason not to do this...?

Kevin

probonopd:
> Up to you, but I tend to subscribe to the "be liberal in what you receive, and strict in what you send" principle of Internet protocols in things like this. If it doesn't cost us much (it doesn't) to be a little more flexible, then why wouldn't we?

Makes sense to me. 

> am struggling to think of a reason not to do this...?

Well, it makes the code larger, more complex to understand and maintain. And it's work. But since you've already done the work, it'd be a shame not to merge it.

Thanks! :+1:

diyelectromusic:
> Well, it makes the code larger, more complex to understand and maintain. And it's work. But since you've already done the work, it'd be a shame not to merge it.

Haha!  Compared to some bits of the code, this is "small fry" :)
And don't mind the "work" front - it's useful to get to know new bits of the code, even if the consensus ends up "not required".

Kevin

Issue #391, opened 2022-12-10
In MiniDexed.cpp Program/Bank Change ignore number of programs/banks allowed


diyelectromusic:
Should BankSelectLSB bounds-check the provided MIDI value and limit to the number of loaded banks?
Similarly should ProgramChange limit to the number of voices per bank?

Alternatively should there be an option that if a ProgramChange message comes in over MIDI the values 0..127 map onto the first four banks of 32 voices each?  i.e.:
0..31 -> bank 1 voices 1-32
32..63 -> bank 2 voices 1-32
64..95 -> bank 3 voices 1-32
96..127 -> bank 4 voices 1-32

Of course this means that a MIDI sequence: Set Bank=2; Set Program = 4 would end up selecting bank 1, program 4 (as programs 1-32 map onto bank 1)... so this probably has to be configurable...

BobanSpasic:
DX7 accepts ProgramChanges 0..31
DX7II accepts ProgramChanges 0..63 for internal memory and 64..127 for cartridge.

diyelectromusic:
I'll have a play with this at some point, but I'm waiting for some of the current PRs to go through first so I'm not having to mess around merging minidexed.cpp... :)

abscisys:
Why not to keep limiting prog change to 0 to 31 which is the size of a cartridge and have support bank select LSB and MSB to increase the number of supported cartridges to 2^ 14? 
If any interest, this is what I submitted as the PullRequest #395

diyelectromusic:
The issue is that things like voices per bank requires the thing sending PC messages to know about the synth it is sending them too, whereas sending PC 0 to 127 is universal.

That isn't to say that recognising bank select shouldn't be improved, but for many cases recognising PC 0 to 127 gives access to much of the functionality for lots of users, and is also more "GM like" which again many will be familiar with.

Aside, if you want to allow for 2^14 banks, then I really, really think we have to limit the selection to "installed banks" not "potentially installed" banks... which is what is happening in https://github.com/probonopd/MiniDexed/pull/392

PR #464, opened 2023-04-01
Expand Program Change messages across banks

diyelectromusic:
I've added an option to allow Program Change 0..127 to "stripe" across four consecutive banks, starting with the selected bank. This allows a simple configuration of just four banks to have all 128 voices available over PC messages without requiring an actual bank select message.  This does not affect the bank select mechanism though.

I couldn't think of a downside of this being enabled, so I enabled it by default.  But it is a config option regardless.

This was discussed a little in Issue #391 

Kevin

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,22 +16,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Recursively pull git submodules
+    - name: Get specific commits of git submodules
       run: |
-           set -ex
-           git submodule update --init --recursive
-    - name: Use Circle develop branch for SSD1306 display rotation support until it is merged upstream
-      run: |
-           set -ex
-           cd circle-stdlib/
-           git checkout e318f89 # Needed to support Circle develop?
-           cd -
-           cd circle-stdlib/libs/circle
-           git checkout ec09d7e # develop
-           cd -
-           cd circle-stdlib/libs/circle-newlib
-           git checkout 48bf91d # needed for circle ec09d7e
-           cd -
+           sh -ex ./submod.sh
     - name: Install toolchains
       run: |
            set -ex
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -71,6 +71,7 @@ void CConfig::Load (void)
 	m_bMIDIRXProgramChange = m_Properties.GetNumber ("MIDIRXProgramChange", 1) != 0;
 	m_bIgnoreAllNotesOff = m_Properties.GetNumber ("IgnoreAllNotesOff", 0) != 0;
 	m_bMIDIAutoVoiceDumpOnPC = m_Properties.GetNumber ("MIDIAutoVoiceDumpOnPC", 1) != 0;
+	m_bHeaderlessSysExVoices = m_Properties.GetNumber ("HeaderlessSysExVoices", 0) != 0;
 
 	m_bLCDEnabled = m_Properties.GetNumber ("LCDEnabled", 0) != 0;
 	m_nLCDPinEnable = m_Properties.GetNumber ("LCDPinEnable", 4);
@@ -177,6 +178,11 @@ bool CConfig::GetIgnoreAllNotesOff (void) const
 bool CConfig::GetMIDIAutoVoiceDumpOnPC (void) const
 {
 	return m_bMIDIAutoVoiceDumpOnPC;
+}
+
+bool CConfig::GetHeaderlessSysExVoices (void) const
+{
+	return m_bHeaderlessSysExVoices;
 }
 
 bool CConfig::GetLCDEnabled (void) const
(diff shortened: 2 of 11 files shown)


diyelectromusic:
Yes will do. If accepted I can update the wiki when ready

probonopd:
This branch has conflicts that must be resolved. I kinda feared this would happen...

diyelectromusic:
Yeah - I thought each was building on each other.  I just don't "get" git :)

The conflicts I think are minimal.  Is that something I have to resolve or something you do as part of a merge?

probonopd:
Would be tremendous if you could resolve the conflicts. Thank you so much!

In general, it's best to get one PR merged (all changes you make in your branch that went into the PR get automatically amended to the PR), and once that is merged, start another PR. As soon as there are multiple PRs open at the same time, there is always the chance of conflicts in case both PRs edit the same file(s) in similar places.

diyelectromusic:
Right, looks like both of these have their conflicts resolved... although accepting one may well produce more in the other!  Doesn't really matter though - I'll take another look then.

BTW if there was an order, I'd say accept #463 first.

Kevin

probonopd:
> BTW if there was an order, I'd say accept https://github.com/probonopd/MiniDexed/pull/463 first.

Done :-)

diyelectromusic:
Documentation for this and #459 now added to the wiki.

Keivn

probonopd:
Thanks Kevin, that's very helpful.

diyelectromusic:
If there are no objections, can we merge this one in? :)

Thanks,
Kevin

Banana71:
@diyelectromusic, 
is this the door opener to play GM midi files?  Then we should create a GM bank here (4 banks).  This enables completely new applications.  With two miniDexed you could then play any GeneralMidi®️ file 🎉🧚🏽

diyelectromusic:
> @diyelectromusic, is this the door opener to play GM midi files? Then we should create a GM bank here (4 banks). This enables completely new applications. With two miniDexed you could then play any GeneralMidi®️ file 🎉🧚🏽

Well kind of maybe... ish... not sure.  The main problem is that GM files tend to send a lot of additional data over for standardised controllers and things that I doubt the MiniDexed would support or recognise. It might even map them onto something else which always causes entertaining results.  Also they tend to have a lot more than 8 tracks on the go.  And as you say, no drums :)

But it would certainly be possible to have a "GM equivalent" instrument list across four banks, yes.

That isn't to say you can't send MIDI files through.  I still think my [80s DX7 style "Jupiter" from Holst's Planets](https://www.youtube.com/watch?v=E5cnrfKYE4M) was quite credible on MiniDexed, but I find that these older systems need the MIDI files to be pretty "clean" to sound any good :)

Kevin

diyelectromusic:
Mind you, I'm now wondering about a cheap [VS1053/VS1003 module](https://diyelectromusic.wordpress.com/2022/09/11/arduino-vs1053-general-midi-synth/) on a Pi IO board that could do the GM drums for a MiniDexed FM GM module... or maybe I should stay OPL and add a [YM2413](https://diyelectromusic.wordpress.com/2023/01/27/arduino-opl-fm-synth/)...

probonopd:
And don't forget the https://github.com/dwhinham/mt32-pi, which is built on Circle just like MiniDexed.

diyelectromusic:
> And don't forget the https://github.com/dwhinham/mt32-pi, which is built on Circle just like MiniDexed.

I've got a few of those too (and a real MT-32 as it happens) :)

Kevin

diyelectromusic:
> If there are no objections, can we merge this one in? :)

Hi @probonopd is there an issue with this one?  Just thinking if it isn't going to be merged in I probably should take the setting off the Wiki :)

Kevin

probonopd:
Thanks @diyelectromusic.
Could we add a sentence to the documentation explaining what this is useful for?

diyelectromusic:
> Could we add a sentence to the documentation explaining what this is useful for?

I already added this to the wiki describing the setting - is that what you mean or do you want something else (if so where)?

Kevin

diyelectromusic:
On the one hand it mirrors the voice selection pattern of a DX7II (apparently) as described in the linked issue; on the other it allows more voices to be chosen from just using Program Change messages without having to have knowledge of how voices/banks mix together; if only four banks were installed then all voices could be chosen from a more modern keyboard with 128 voices; also, as was stated above, it could allow someone to build up a General MIDI style set of voices; it also makes it easier for a "universal" (i.e. non-DX7 specific) DIY program changer to select one of 128 voices...

There are many possible use-cases, but ultimately it is just another option for people to use if they wish.  I'm sure people will find all sorts of use cases should they wish to - I don't think we need to be prescriptive in the documentation.  If they need convincing to use it, they probably don't need it.  But if they've played with voice selection from other MIDI devices (rather than a DAW that has been configured for DX7s) it might prove quite useful.

Kevin

PR #466, opened 2023-04-01
Skip empty voices while scrolling

donluca:
There are banks which contain empty voices labeled as "INIT VOICES".
Those are just blank voices which can be skipped to make scrolling between lots of banks faster.

### DIFF
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -582,11 +582,10 @@ void CUIMenu::EditProgramNumber (CUIMenu *pUIMenu, TMenuEvent Event)
 	}
 
 	string voiceName = pUIMenu->m_pMiniDexed->GetVoiceName (nTG); // Skip empty voices
-	if (voiceName == "INIT VOICE" 
-	    || voiceName == "init voice" 
-	    || voiceName == "          " 
-	    || voiceName == "----------" 
-	    || voiceName == "~~~~~~~~~~" ) 
+	if (voiceName == "EMPTY     "
+	    || voiceName == "          "
+	    || voiceName == "----------"
+	    || voiceName == "~~~~~~~~~~" )
 	{
 		if (Event == MenuEventStepUp) {
 			CUIMenu::EditProgramNumber (pUIMenu, MenuEventStepUp);
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -582,7 +582,12 @@ void CUIMenu::EditProgramNumber (CUIMenu *pUIMenu, TMenuEvent Event)
 	}
 
 	string voiceName = pUIMenu->m_pMiniDexed->GetVoiceName (nTG); // Skip empty voices
-	if (voiceName == "INIT VOICE") {
+	if (voiceName == "INIT VOICE" 
+	    || voiceName == "init voice" 
+	    || voiceName == "          " 
+	    || voiceName == "----------" 
+	    || voiceName == "~~~~~~~~~~" ) 
+	{
 		if (Event == MenuEventStepUp) {
 			CUIMenu::EditProgramNumber (pUIMenu, MenuEventStepUp);
 		}
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -587,15 +587,25 @@ void CUIMenu::EditProgramNumber (CUIMenu *pUIMenu, TMenuEvent Event)
 		return;
 	}
 
-	string TG ("TG");
-	TG += to_string (nTG+1);
+	string voiceName = pUIMenu->m_pMiniDexed->GetVoiceName (nTG); // Skip empty voices
+	if (voiceName == "INIT VOICE") {
+		if (Event == MenuEventStepUp) {
+			CUIMenu::EditProgramNumber (pUIMenu, MenuEventStepUp);
+		}
+		if (Event == MenuEventStepDown) {
+			CUIMenu::EditProgramNumber (pUIMenu, MenuEventStepDown);
+		}
+	} else {
+		string TG ("TG");
+		TG += to_string (nTG+1);
 
-	string Value = to_string (nValue+1) + "=" + pUIMenu->m_pMiniDexed->GetVoiceName (nTG);
+		string Value = to_string (nValue+1) + "=" + pUIMenu->m_pMiniDexed->GetVoiceName (nTG);
 
-	pUIMenu->m_pUI->DisplayWrite (TG.c_str (),
-				      pUIMenu->m_pParentMenu[pUIMenu->m_nCurrentMenuItem].Name,
-				      Value.c_str (),
-				      nValue > 0, nValue < (int) CSysExFileLoader::VoicesPerBank-1);
+		pUIMenu->m_pUI->DisplayWrite (TG.c_str (),
+					      pUIMenu->m_pParentMenu[pUIMenu->m_nCurrentMenuItem].Name,
+					      Value.c_str (),
+					      nValue > 0, nValue < (int) CSysExFileLoader::VoicesPerBank-1);
+	}
 }
 
 void CUIMenu::EditTGParameter (CUIMenu *pUIMenu, TMenuEvent Event)


probonopd:
How do you load INIT patches then?

donluca:
What would be the use case of a user loading an INIT VOICE (an empty patch)?

Would he be using it to create a sound from scratch on MiniDexed? That seems, IMHO, highly unlikely as it would require a huge effort just navigating through the various menus to adjust all the operators settings.

Still, an easy workaround would be to create an empty patch with a different name, like ```INIT_VOICE``` or ```INIT VOICE*``` or anything even slightly different from ```INIT VOICE```.

To me that's a small price to pay compared to the benefit of skipping empty patches.
There are banks where only 4 or 5 voices are populated and the rest are empty which translates to a lot of turns of the rotary knob to go past them, not to mention that if you have to go back to a previous patch you have to traverse them again.

probonopd:
> What would be the use case of a user loading an INIT VOICE (an empty patch)?

Yes, as per https://github.com/probonopd/MiniDexed/wiki/Exercises.
I "learned" FM this way.

donluca:
Well, hats off to you guys, I would never have the patience to do that straight on Minidexed.

By the way, on the guide you linked I read:

> Open MiniDexed -> TG1 -> Bank
> Turn the knob right until you have selected a bank called NO NAME
> Open MiniDexed -> TG1 -> Voice
> Turn the knob right until you have selected a voice called INIT VOICE

I can change my code to create an exception where if the bank name is "NO NAME", then it doesn't skip "INIT VOICE", so you can use that to create patches from scratch.

probonopd:
Let's do that @donluca, can you please make that change in your [donluca:skipEmptyVoices](https://github.com/donluca/MiniDexed/tree/skipEmptyVoices) branch, then it should automatically get appended here. Thanks!

probonopd:
Would you agree that we filter out all names but _not_ "INIT VOICE"?

Banana71:
> If I am not mistaken, @Banana71 is also creating his sounds on MiniDexed, not on a PC.

I usually create a new voice on the PC with Dexed.  But most of the time I work on a performance by fine-tuning each voice directly on the miniDexed.  Even if the same voice/bank is in the performance, it is mostly more or less heavily edited patches.

donluca:
> Would you agree that we filter out all "empty" names but not "INIT VOICE" if it is exactly written like this?

99% of the empty patches are called "INIT VOICE" (refer to [this post](https://github.com/probonopd/MiniDexed/discussions/454#discussioncomment-5514621) ), if we remove ```INIT VOICE``` from the filter we've made this change basically useless.

probonopd:
Hmm. I was hoping there was a way to distinguish between "wanted INIT VOICEs" and "unwanted blanks".

donluca:
The "proper" way of doing this would be to read the the content of the voice and compare it to the content of an "empty" voice: if it matches, then skip it.

I took a shortcut by just checking the name of the voice which is less than ideal (my last commit is the perfect example of this: it shows that some people have used different names for "emtpy" voices) because I have no idea of how to make voice comparisons, but if someone more knowledgeable can code this, it would definitely be the best way of doing it.

Still, there is merit to this approach: since it only checks the name of the voice, it become trivial to play around it: with Kevin we decided that we would provide with MiniDexed a "User Bank" made only of empty voices so people can use it to create their custom voices/performance and we can achieve that by naming the empty voices ```USER VOICE``` instead of ```INIT VOICE```.

probonopd:
Thing is, all instructions (e.g., DX7 books) call it "INIT VOICE".

donluca:
Yes, that's kind of my point and why I chose this approach in the first place 😅 😅 😅 

Since those "empty voices" have been described by tech documents to be explicitly called ```INIT VOICE```, I chose this approach so I could be fairly sure that if a patch is called ```INIT VOICE```, then it means that it is empty and there's no point in playing an "empty" voice, hence I can skip it.

To put this back into context: I have done this "hack" (if you will), because normally banks are always fully populated: each voice is written, there are no empty voices.
The issue arose when we started throwing in banks which have been made by third parties who (rightfully) didn't care about filling all the 32 spots of a bank with voices: some wanted to do 3-4 patches, others maybe 20 and so on, which meant that they left several "empty" voices in their custom banks and it gets *really* tiresome to scroll through dozens of these.
This is why I put this check in my private repo to skip them.

The caveat that I didn't consider at the beginning is that by doing so we've removed the possibility of creating a patch from scratch and we're forcing users to edit other patches.
This has been solved with the workaround described in my previous comment.
And, besides, we can always disable this behavior through an option and keep the default (meaning the one we have now where empty patches are not skipped) as the standard option, so that if someone becomes annoyed at scrolling empty voices, he can enable this option so he doesn't have to scroll through dozens of them.

Banana71:
> with Kevin we decided that we would provide with MiniDexed a "User Bank" made only of empty voices so people can use it to create their custom voices/performance and we can achieve that by naming the empty voices `USER VOICE` instead of `INIT VOICE`.

The "UserBank" would therefore be a RAM in which you can store patches. 👍🏼

donluca:
> The "UserBank" would therefore be a RAM in which you can store patches. 👍🏼

Unfortunately, as of now, the solution is to have a "physical" file named 000000_User_Bank.syx which is just a bank with 32 empty voices named ```USER VOICE```.
We'd love to have it in RAM but I have no idea how to do it.

Banana71:
> with Kevin we decided that we would provide with MiniDexed a "User Bank" made only of empty voices so people can use it to create their custom voices/performance and we can achieve that by naming the empty voices `USER VOICE` instead of `INIT VOICE`.

The "UserBank" would therefore be a RAM in which you can store patches. 👍🏼
If I want to save a patch, is the next available memory space taken?  Wouldn't it make more sense if you could choose the storage space freely in the UserBank.
Or how to understand that?

probonopd:
`find /tmp/sysex -type f -exec sed -i '' -e 's|INIT VOICE|EMPTY     |g' {} \;`

donluca:
Oh, I wanted to try and not touch the sysex files at all, but yeah, that can be definitely done.

So your plan is to use my curated library, replace are occurrences of voices named
```INIT VOICE```
```init voice```
```          ``` (10 empty spaces)
```----------```
```~~~~~~~~~~```

with ```EMPTY``` and then change the filter to skip all patches named ```EMPTY```.

That could work as well!

**Comment by probonopd** [comment]
Let's filter out these names.

```
	if (voiceName == "EMPTY     "
	    || voiceName == "          "
	    || voiceName == "----------"
	    || voiceName == "~~~~~~~~~~" )
```

In the zip linked above, I have edited your library to contain `EMPTY     ` insttead of `INIT VOICE` in all but the 01... directory.

diyelectromusic:
Ok this has now really confused things - we are moving towards bespoke sysex files and I'm not sure that is a good direction of travel...

I thought we agreed the solution was a user bank to allow people to create voices, so skipping INIT VOICE was then ok to do?

probonopd:
I can't see any downsides of the current solution, maybe I am overlooking something?

diyelectromusic:
The original aim was to skip blank voices in SysEx banks.  But now we're only saying it will do that with "special" banks we've messed around with - is that really want we want to say?

I guess either choice is deviating from a true DX7, so maybe it doesn't matter...

probonopd:
From what I understood, the purpose of skipping anything is to have no blanks in the curated collection of all known existing sysex voice banks out there.

At the same time, when a user manually adds other banks, things callled "INIT VOICE" should not be skipped imho. If they want them to be skipped, they can use one of the names mentioned above.

diyelectromusic:
Ah ok - if that is the intent, then that should be fine.

So any random sysex the user downloads or adds will be treated as a DX7 would treat it - makes sense.

Kevin

donluca:
In a "real" DX7 there would be no "empty" patches because all internal voices would be populated and the same was true of all the Voice Cards you could buy, both from Yamaha and third parties.

The problem arose when people started making their own banks and weren't interested in making banks filled with 32 patches, so banks with ```INIT VOICE``` started to appear.

We could argue that this is a "post-DX7" problem, when the Internet happened and people started sharing online their custom patches.

Me, personally, I would have gone with the approach me and @diyelectromusic were discussing, which was to filter out ```INIT VOICE``` (which would have never presented themselves in official or commercial cards for the DX7) and use the name ```USER VOICE``` for user banks.

In the end I guess it's just a matter of naming things and whichever approach don't change much.
I don't like messing with sysex files (apart from stripping headers to achieve higher performance) and I would have preferred to keep them "stock".

On the other hand it is more immediate for a user to fire up Dexed and create an empty bank which will be filled with ```INIT VOICE``` instead of going in and changing all the name of the patches to ```USER VOICE```.

I don't have a problem with the approach @probonopd took, but I would have personally preferred the way we were handling things with @diyelectromusic .

diyelectromusic:
> We could argue that this is a "post-DX7" problem, when the Internet happened and people started sharing online their custom patches.

"when the Internet happened"... yep.  All went down hill from there :)

Kevin

probonopd:
> In a "real" DX7 there would be no "empty" patches because all internal voices would be populated and the same was true of all the Voice Cards you could buy, both from Yamaha and third parties.

...unless you used a RAM card for your own creations.

If people are vehemently against changing the names of empty voices in the banks, we could e.g., have files with the same name as the syx but a different extension that would contain auxiliary information about what to do with the bank: which voices to skip, which MIDI range to use, etc. - this way the original syx files would be untouched but we could still deliver additional settings that are not part of the banks themselves...

donluca:
Actually... there is an issue with this approach: if the user wants to get other banks from the internet and add them, they are most likely to have ```INIT VOICE``` in them, so we're kind of back to square one.
In order to skip them the user would have to replace the name with something different than ```INIT VOICE```.

I'm not entirely convinced this is the best approach.

probonopd:
...but that's my point: If the _user_ loads something called `INIT VOICE` then I am assuming that the user _wants_ that. As a starting point for own creations. (Hopefully with the option to save them to the bank at some time.)

diyelectromusic:
> Actually... there is an issue with this approach: if the user wants to get other banks from the internet and add them, they are most likely to have `INIT VOICE` in them, so we're kind of back to square one. In order to skip them the user would have to replace the name with something different than `INIT VOICE`.

As @probonopd said here https://github.com/probonopd/MiniDexed/pull/466#issuecomment-1500631434 the idea is the curated set of voices is the only one with "skipping" behaviour. Any random sysex will be treated as a DX7 would treat it - with a range of INIT VOICE patches.

I can see that working.  If we document it, then people can change the names of voices to EMPTY if they wish too.

Kevin

diyelectromusic:
> (Hopefully with the option to save them to the bank at some time.)

I was thinking once the current PR flurry is over I might take a look...

donluca:
Oh ok, so we want to go all in with the "curated library".

I see, yes, now I understand better this approach and I can get behind it.

Issue #468, opened 2023-04-03
Adding Bank MSB/LSB handling has messed up UI parameter handling


diyelectromusic:
It would appear that there is an undocumented dependency between the TGParameter Enums from https://github.com/probonopd/MiniDexed/blob/main/src/minidexed.h#L151 and the specific ordering of the UI menu parameters in https://github.com/probonopd/MiniDexed/blob/main/src/uimenu.cpp#L219

Consequently, adding the two additional MSB/LSB bank parameters means the UI menu is now out of sync and causes assert failures on certain menus.

I'm working on it :)

Kevin

PR #469, opened 2023-04-03
Fix for Issue #468 - introducing Bank MSB/LSB messed up the UI parameter handling.

diyelectromusic:
Fixes Issue #468 which was introduced as part of PR #461


probonopd:
Thank you @diyelectromusic

Issue #424, opened 2023-01-22
Script to checkout correct versions of submodules


diyelectromusic:
I think it would ease development if there was a script, maybe alongside build.sh, that checks out the same versions of submodules as are used in the github actions.

Of course there may be much better ways to achieve this in GitHub (I'm no GitHub/git expert), so it doesn't have to be done this way - but I think we should have a simple way to ensure anyone contributing code is starting from the same versions built into the releases.

Kevin

probonopd:
Good idea. Basically we could move

https://github.com/probonopd/MiniDexed/blob/fcea41ead3fadd962342ffdf8241e48c1f84358e/.github/workflows/build.yml#L26-L33

to its own script that could also be run locally.
Wdyt?

diyelectromusic:
Well I must confess I don't really understand how the whole things hang together (yet), so whatever you think works best.  As long as it won't overwrite any local changes in progress that could be an option.

I suppose I was thinking the build is something that is done quite regularly as part as normal development, but the checkout and initialisation of any specific version of the repository is something only done at specific controlled points.

I guess the ideal would be that updating a fork means it gets updated with the right versions of submodules automatically, but I don't know how the versions listed in the code itself relate to the versions in build.yml...

Kevin

trygvelu:
I don't think this should be necessary. Running 'git submodule update --recursive' after checking out the main project should check out the referenced commits in the submodules, provided that the _references to the submodules' commits have also been committed_ in the main project and "chain" of submodules. That is, whenever a developer has updated a submodule in the project in order to bring in new functionality, squash bugs etc, then: git add/commit the submodule in the project to bring in the new commit/state of the submodule.

Creating a script that checks out specific commits sounds like extra work to me - how often would the references/script be updated? IMO the build workflow should just check out from origin/main branch, then update submodules and proceed with building from there.

However - I fully agree that it should be easier to build MiniDexed. I had to fiddle quite a bit to build a functioning version.

probonopd:
Thing is, it is a pain in the * to update git revisions on GitHub (e.g., switch Circle to another develop git commit). The only pragmatic solution I found so far is the one that is in place now.

diyelectromusic:
Any further thoughts on this one?  When setting up a new MiniDexed source tree for the latest round of updates I was doing all this by hand again... :)

It would be really nice to just run a "refresh" or "resync" type action and have it all ready to go again.  But very happy to be told there is a better way!

Kevin

[...]

PR #473, opened 2023-04-05
Support for subdirectories in the SysEx voices directory

diyelectromusic:
Initial support for (nested) subdirectories in the sysex/voices folder.  Splitting up directories into less than 100 files per directory seems to really help with the performance issues discussed in https://github.com/probonopd/MiniDexed/discussions/456.

Kevin

### DIFF
--- a/src/sysexfileloader.cpp
+++ b/src/sysexfileloader.cpp
@@ -84,6 +84,7 @@ CSysExFileLoader::~CSysExFileLoader (void)
 void CSysExFileLoader::Load (bool bHeaderlessSysExVoices)
 {
 	m_nNumHighestBank = 0;
+	m_nBanksLoaded = 0;
 
     DIR *pDirectory = opendir (m_DirName.c_str ());
 	if (!pDirectory)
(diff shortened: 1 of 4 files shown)


diyelectromusic:
I've just realised this also includes the additional build script for #424 - sorry, that was a mistake!

But as it is just one stand alone file, if you're happy with it, we can just leave it here. 

Kevin

donluca:
@diyelectromusic You're pushing more changes into this PR, I see the fix for bank numbers as well, maybe it would be best to keep those change on separate PRs.

diyelectromusic:
It is all part of sorting out the bank handling... but I don't mind either way.

Up to @probonopd I guess.

Kevin

probonopd:
@diyelectromusic what does this mean for loading banks via MIDI CC messages?

diyelectromusic:
Yes, this shouldn't affect MIDI as we only support single voice loads over SysEx anyway and they go straight into Dexed without touching MiniDexed itself really.

More on how it links together here: https://github.com/probonopd/MiniDexed/discussions/454#discussioncomment-5516966

I should probably document this somewhere useful... :)

Kevin

PR #481, opened 2023-04-17
If running on a RPI 1 or single core Pi then default to the TG1 menu as the home menu

diyelectromusic:
When running on a single core RPi there is only one tone generator, so this change will get the menu system to default to the TG1 menu rather than always having to select TG1 before changing anything.

Note: the higher level menu hasn't gone away and can still be reached by selecting "back" - the highest level menu structure is still: TG1 - Effects - Performance.

But with this change the "home" menu is at the TG1 level, showing "Voice" which is a lot more convenient when only one TG is present.

### DIFF
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -338,6 +338,23 @@ CUIMenu::CUIMenu (CUserInterface *pUI, CMiniDexed *pMiniDexed)
 	m_nCurrentParameter (0),
 	m_nCurrentMenuDepth (0)
 {
+#ifndef ARM_ALLOW_MULTI_CORE
+	// If there is just one core, then there is only a single
+	// tone generator so start on the TG1 menu...
+	m_pParentMenu = s_MainMenu;
+	m_pCurrentMenu = s_TGMenu;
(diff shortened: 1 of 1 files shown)


probonopd:
Undecided about this.

I see that this may make it easier, but it would make documentation/tutorials harder because now we would have different behavior for the different models.

diyelectromusic:
I don't see that tbh.  And I suspect the number of people running on a Pi V1 is probably not that great.  From personal experience, the additional level of menu for non-existent other TGs really doesn't make much sense.

And if someone does happen to be using a Pi 1 and is following the tutorial on the wiki, it is already not really accurate as it talks about TG2, etc and the effects menu which don't apply.

And the menu presented is shown in the graphic just two below the first, so I really don't think it is a major issue compared to the much more meaningful "home" location.

Kevin

probonopd:
OK, please do update the wiki with a note for RPi 1 and Zero (but not Zero 2) users. Thanks!

diyelectromusic:
Done.

probonopd:
Thank you Kevin.

Issue #384, opened 2022-12-02
Initial brass sound stays throughout all patches


Slider2732-2:
A perplexing situation.
Build is a Pi 4, with a rotary encoder, latest MiniDexed but no screen or DAC yet. My Beatstep Pro didn't seem to fire notes, nor Roland JU-06A. Instead, am using a USB qwerty keyboard, due to no dedicated MIDI controller keyboard.
Also, the sysex/voice folder is present and sounds are in there from Dexed cart.
At boot, I can play a brass type of sound.
On pressing the rotary switch twice (slowly) I can then dial in other sounds by moving the encoder anti-clockwise.
However, the brass sound stays and can also be heard along with any new sound.

The brass sound can be lowered, by setting all 8 pan values to zero in performance.ini, but it's still present. 
I tried swapping out the sounds from sysex/voice to other ones and still the brass sound plays.
Next thing, completely deleting the sysex/voice folder.. aha, the initial bank is within the main code!

Though, it can't be a bug if noone else has the problem?
It must be because of having to use a USB qwerty keyboard?

Many thanks for any help :)

Banana71:
Hi Slider2732-2,

the default performance.ini is a layer of four Brasses TG1 to TG4. You can set in Performance.ini the MIDIChannel2, MIDIChannel3 and MIDIChannel4 to "0" than you hear only a single Sound in TG1.

[performance TG1 only.zip](https://github.com/probonopd/MiniDexed/files/10138833/performance.TG1.only.zip)

with pan you can set the left-right balance in the stereo-output.

probonopd:
> I remember having the same issue when first trying MiniDexed. The default performance.ini might be interesting to show the potential, but it isn't very "beginner friendly". I do wonder that the default ought to be a lot simpler?

Yes, the idea of the default performance is to show the potential of the device. For "real work", a different performance is preferable.

diyelectromusic:
> > I remember having the same issue when first trying MiniDexed. The default performance.ini might be interesting to show the potential, but it isn't very "beginner friendly". I do wonder that the default ought to be a lot simpler?
> 
> Yes, the idea of the default performance is to show the potential of the device. For "real work", a different performance is preferable.

In that case, as someone has already got as far as installing MiniDexed at this point then I think the default performance.ini should be the simplest to allow them to explore it further.  As pretty much the first thing people will do is try to change the voice I think that ought to work with no surprises "out of the box" :)

But maybe that is just me!

[...]

PR #500, opened 2023-05-27
Add Program/TG Select buttons and choice for Program Change to act on Voices or Performances

diyelectromusic:
Includes updates to do the following:
* Allow GPIO buttons for Program Change Up/Down and Tone Generator Up/Down for quick voice/TG selection.
* Support "MIDI buttons" using either note or CC messages for the above too.
* Provide UI and config option for above Program Change shortcuts and Program Change MIDI messages to work on either Voices (in current TG) or Performances (NB: only performances 1-128 can be chosen).

Update inspired by discussion on "headless MIDI" patch changing here: https://github.com/probonopd/MiniDexed/discussions/496

Kevin


probonopd:
Hi Kevin, thank you very much for this PR.

What would prevent us from using LSB and MSB to allow for more than 127 performances?

probonopd:
Build for testing:
[MiniDexed_2023-05-27-a27203b](https://github.com/probonopd/MiniDexed/suites/13191112178/artifacts/718153108)

@Stooovie please try it out and let us know whether this works for you. Thanks!

diyelectromusic:
> What would prevent us from using LSB and MSB to allow for more than 127 performances?

Yes, we'd have to introduce the concept of "banks" for performances too, but that might need a little thinking about as the concept doesn't exist at the moment.  Program Change only has a single byte, hence the 128 limitation.

Kevin

Stooovie:
> Build for testing: [MiniDexed_2023-05-27-a27203b](https://github.com/probonopd/MiniDexed/suites/13191112178/artifacts/718153108)
> 
> @Stooovie please try it out and let us know whether this works for you. Thanks!

Hi, thanks! I set MIDIButtonPgmUp and MIDIButtonPgmDown and it DOES change the patches BUT it sort of adds patches to the original default sound - that is always there. I apologize, I'm not a great candidate for testing as my knowledge of DX7 is almost nonexistent, I was always just playing with presets.

Is this something to do with performances vs programs? Again, sorry, I'll have to read up on these. 

Thanks for your effort so far, it probably does what I want when set up correctly :)

Banana71:
Hi @Stooovie , maybe the following will solve your problem:
Plug your SD card into a PC and rename the 000130_MultiTimbre.ini file to performance.ini. And then overwrite the file performance.ini in the root directory.
Now start your miniDexed with the performance "MultiTimbral" and you have the possibility to change the sounds on the MIDI channels 1-8 with a MIDI controller. The MIDI channel is assigned to each TG according to its name TG 1-8.

Peter

probonopd:
> Hi, thanks! I set MIDIButtonPgmUp and MIDIButtonPgmDown and it DOES change the patches 

Please clarify what you mean by "patches": Voices or performances?

I thought you wanted to change the performances (=all voices running on all tone generators). For that, you need to use the "build for testing" and set `PerformanceProgramChange` to `1` in `minidexed.ini`.

@diyelectromusic will doing so also allow MIDIButtonPgmUp and MIDIButtonPgmDown to change performances? (Just double checking.)

> I've afraid you've fallen foul of this one (not your fault - I still think this is a fault with the default "as shipped" configuration): #384

By default we are shipping a performance that contains multiple voices (more than one tone generator is used). But when changing the performance, then of course all tone generators should be configured according to the new performance loaded. So I would be suprised that if changing performances, some voices from the original performance would be "left over". I'd consider that a bug... or am I not understanding the issue?

Thinking a bit more about it: Instead of having `PerformanceProgramChange` yes/no we could have `PerformanceProgramChangeMIDIChannel` (from 1 to 16). Whenever one would send a program change message to that channel, the performance would be changed. Whenever one would send a program change message to the channels assigned to the tone generators, the voice would be changed. As a result, one could change both individual voices and whole performances relatively easily via MIDI without the need for custom sysex. (Bonus feature: The channel set at `PerformanceProgramChangeMIDIChannel` should be unavailable to be set as a MIDI channel for a tone generator.)
Wdyt?

diyelectromusic:
> @diyelectromusic will doing so also allow MIDIButtonPgmUp and MIDIButtonPgmDown to change performances? (Just double checking.)

Yes, that's the idea (hopefully).

> > I've afraid you've fallen foul of this one (not your fault - I still think this is a fault with the default "as shipped" configuration): #384
> 
> By default we are shipping a performance that contains multiple voices (more than one tone generator is used). But when changing the performance, then of course all tone generators should be configured according to the new performance loaded.

But that is the key issue.  By default, as "shipped" things like Program Change work on voices not performances, hence the (continuing) confusion.  I hope this change might help with the issue a little, but we are going to have to update the default configuration at some point, either to force Performance mode (least preferred for me - I don't think the DX7II defaults to performance mode does it?  Doesn't it start in normal mode?) or to "fix" the default performance (still my preferred option for all the reasons I've stated several times before).

> Thinking a bit more about it: Instead of having `PerformanceProgramChange` yes/no we could have `PerformanceProgramChangeMIDIChannel` (from 1 to 16).

I did wonder about that, but instead went for MIDI channel on TG1 for "performance" mode, figuring it was perhaps least likely to cause confusion.  It was also easier to implement for this first try.

We could try for a dedicated "performance channel", but I certainly don't think we should then further limit TG channels if we do.

Either way, as MIDI channels are part of the performance configuration I think neither is ideal and each will have different limitations.

Maybe give this a bit of time to settle and see how people find it?

Kevin

diyelectromusic:
I've just updated this one again. @probonopd I was wondering if it was possible to merge it in so it doesn't keep getting out of date (assuming you're happy with it of course)?

Kevin

probonopd:
Hi @diyelectromusic, can you please suggest the documentation for https://github.com/probonopd/MiniDexed/wiki/Files for this?

I don't know whether the solution here is exactly what is desirable; I was assuming we basically wanted a way to have a sysex that switches `PerformanceProgramChange` on or off?

diyelectromusic:
Well the discussion meandered quite a lot, but the original ask from https://github.com/probonopd/MiniDexed/discussions/496 
 was: "Is there a way to just be able to change patches with MIDI CCs (preferably 112/111), and/or knobs?"

I left the SysEx part as "Well if someone can verify that we can enable performance mode from MIDI using the sequence I posted below, then I'll have a look at allowing program change to be either performance or voices" - not having any original instruments, there is nothing I can do to test any of it to know what it does.

"In fact, I can look at it regardless and make it a UI/config setting, but I'd still like to know if it was possible over MIDI! :)" - So I did the latter and it has been ready to integrate for a while now.

In terms of documentation, it will expand the comments in the minidexed.ini file:

```
# Program Change mode
#   0 = Only recognise Program Change 0-31.
#   1 = Support 0-127 across four consecutive banks.
# NB: Ignored if PerformanceProgramChange=1
ExpandPCAcrossBanks=1
# Program Change action:
#   0 = Select voices
#   1 = Select performances
#
# NB: If Perfromances, then only reacts to Program Change
#     messages received on the MIDI channel for TG1.
# Warning: Make sure all perfromances use the same MIDI
#          channel for TG1 and that MIDI is enabled.
PerformanceProgramChange=0
```
and the settings for configuring buttons themselves:
```
# GPIO Program/TG Selection
ButtonPinPgmUp=0
ButtonActionPgmUp=
ButtonPinPgmDown=0
ButtonActionPgmDown=
ButtonPinTGUp=0
ButtonActionTGUp=
ButtonPinTGDown=0
ButtonActionTGDown=
```

(and the corresponding MIDI buttons too)

If accepted I'll update the Wiki (it doesn't seem appropriate to update the documentation before the feature has even been agreed... let alone integrated?).

Kevin

Banana71:
Hi Kevin,

ExpandPCAcrossBanks=1 works as expected. Voice changes are immediately reflected in the UI.
PerformanceProgramChange=1 changes the performance according to MIDI program change but the UI is not updated.

My thoughts:
PerformanceProgramChange=x (0-16)

X= 1-16: On MIDI channel x, program change commands change the performance. Voices are changed on all other channels.

PerformanceProgramChange=0 (off): ProgramChange commands only affect the Voices, Performances are not changed.

This gives you the opportunity, for example from a DAW, to change the Performances and the Voices. If I understand correctly, the Yamaha TX-802 behaves the same way.

diyelectromusic:
> PerformanceProgramChange=1 changes the performance according to MIDI program change but the UI is not updated.

I think this was because of how the performance menus are structured. The voice menus work on a "if it is visible it is selected" basis but the performance is on a "you have to choose it to load it" basis, so the visible one isn't the selected one until it is actually selected.  So atm the performance menu /is/ updated (the [L] symbol moves) but the name doesn't auto-scroll to the active one.  I'm not sure if that is actually possible tbh with how it is structured, but I'll take another look...

> My thoughts: PerformanceProgramChange=x (0-16)

I did wonder about this too.  I shied away from it as I thought we wanted a menu option to mimic the "performance" button on the real synth UI, so it was either on or off.  Maybe I should add PerformanceProgramChangeChannel (yes I know... yet another config option...), but maybe the simpler is to update the menu to select between OFF then 1-16... I'll have another look.

Although again, I can't guarantee this will be easy to slot into the MIDI handling, as ProgramChange is definitely a Voice channel message in the handling, but I'll see what I can do - I did need to introduce "system level channel handling" for MIDI control buttons, so it might already have the necessary glue in place...

> X= 1-16: On MIDI channel x, program change commands change the performance. Voices are changed on all other channels.

So you want to be able to select a performance using the PerformanceChange MIDI channel, but then override the individual selections for TGs based on their own channels and receiving ProgramChange messages for voices?  I think we could do that, but are you sure that is what the TXX-802 does - that sounds like a recipe for getting things really out of sync in terms of DAW vs MiniDexed UI vs Performance Settings vs what is actually playing! :)

I guess part of the problem is that as MIDI channel is stored within a Performance (is that the same on a TX-802?), selecting a different performance will be changing MIDI channels "under the hood" for individual TGs so sending a Program Change might mean different things for different performances - but maybe that doesn't matter.  I've no idea how we describe that in the documentation mind :)

... but if that is the consensus of what we think it ought to be, I think that is possible. It just sounds very complicated and ripe for all sorts of confusion! :)

Kevin

probonopd:
> I think this was because of how the performance menus are structured. The voice menus work on a "if it is visible it is selected" basis but the performance is on a "you have to choose it to load it" basis, so the visible one isn't the selected one until it is actually selected. So atm the performance menu /is/ updated (the [L] symbol moves) but the name doesn't auto-scroll to the active one. I'm not sure if that is actually possible tbh with how it is structured, but I'll take another look...

Hi Kevin, thanks for having a look. If you find a way to make it consistent, I'd be all for it. Personally I dont like that voices and performances behave differently in this regard.

diyelectromusic:
We also need to be careful not to just focus on the TX-802 (which I think was the rack version of the DX7-II?), as we have the TX-7, TX-816/216 too and MiniDexed is become a weird mix of them all :)

It's nice to support as much as we can, but it is starting to get real messy! I'll have a proper read and see what I can decipher too!

Kevin

diyelectromusic:
Ok, so isn't that flow chart basically saying:
IF PerformanceSelectChannel is NOT ZERO THEN treat PC messages as choosing performances
ELSE treat PC message as choosing voices

So the idea that it will select performances if on the selected channel and voices on any other channel doesn't seem right?

Or is it... actually, I've no idea what that flow chart is saying!  Anyone want to buy me a TX-802? :)

Banana71:
My mistake. The Yamaha TX-802 actually behaves differently.
Regardless, it's not all that complicated.
PerformanceProgramChange=0 (off): The miniDexed works as usual. ProgramChange commands affect all active MIDI channels of the performance accordingly.
PerformanceProgramChange=x (1- 16)
Program Change commands have an effect on all active MIDI channels of the Performance; the Performance is only changed when it is received on the set MIDI channel x.

probonopd:
We need a MIDI (sysex?) command to toggle PerformanceProgramChange on/off, no?

Banana71:
I'm also still looking for original DX7 equipment. My favorite is the DX7IID or DX7IIFD.

Banana71:
> We need a MIDI (sysex?) command to toggle PerformanceProgramChange on/off, no?

The best thing would be a utility page in the UI where you can make some settings and also have the option to save them. It sounds like a lot of work though.

diyelectromusic:
This is from the "MIDI Technical Data and Charts" from the DX7II:

image

Which seems to confirm the either/or nature of the setting.

The SysEx stuff was in the previous discussion and taken from my reading of the same charts, but I really need someone to confirm it.  But I don't think we really need to change this over MIDI.  The main ask is a config setting to allow the user of MiniDexed to decide if it is voices or performances, coupled with a UI setting to change "live" that goes quite a long way I'd say?

It certainly doesn't make a lot of sense to me to be switching live - if performances are being selected but you then want to select individual voices on specific TGs then you really have to re-initialise the MiniDexed into 8 TGs of known configuration before anything you send makes much sense.  As I say, the performances could have completely messed with the MIDI channels for TGs and so on, not to mention the fact that the voices are not linked to ROMs when loaded from performances...

No, I think it is pretty good to have people initialise the unit in either voice or performance mode - otherwise there is room for huge amounts of confusion!

Kevin

probonopd:
OK; I see. So do you think this should be merged as-is?

diyelectromusic:
I think part of the reason we're struggling is that the actual performance data for the DX7II doesn't include the MIDI channel (see table "5-3 Performance Parameters" in the "MIDI Technical Data and Charts" manual.

This means that fiddling with performances (if I'm understanding it correct) won't mess with the MIDI configuration.  In our case that isn't true as the MIDI channels for each TG is part of the performance configuration.

We also store bank/voice numbers even though they aren't used (which is also a bit confusing - more on that here https://github.com/probonopd/MiniDexed/discussions/454#discussioncomment-5516966)... but there is only so much messing around I'm prepared to do as part of this one change :)

Kevin

diyelectromusic:
I'll do a few updates - at the very least to see if I can make the Performance MIDI channel a little more obvious.  I'm trying to remind myself what I did now :)

It might also be worth changing the config setting to use the words used in the original manuals too - I'll see if I can match or not!

Kevin

diyelectromusic:
Ok, on reading this... no way am I implementing a mapping table of performances to MIDI PC messages - sorry!  0..127 will map directly onto performances 1 to 128 :)

image

diyelectromusic:
Haha!  Just fell foul of some of that confusion I was talking about myself...

When testing the PC changing, my initial config file performance.ini has TG1-8 mapped to MIDI channels 1-8 and all seems fine.  Then I switch to select performances and all seems ok.  Then I switch back to voices and weirdly I see six sets of Program Change messages going to MIDI channel 1... and I can't work out why...

... and then I realise that when testing the performance selection I've selected a voice that has six TGs all configured for MIDI channel 1 :/

I'm really starting to think we should not be having MIDI channel (or voice or bank) in the saved performance files...

Looking at the TX802 I see the parameter dumps are in several parts: the performance file has the effects/etc in it; but there is also a "system parameters" dump which is defined as follows:

image

That would be quite a big change, but it is also probably the way this kind of thing ought to be handled.... but I'm not going to even attempt to think about that as part of this change.

Kevin

diyelectromusic:
Ok, so this is where we are with this one.

There are now GPIO and MIDI button options for a PGM Up/Down and TG Up/Down selection UI elements.

PGM Up/Down can be configured to act on either Voices within a TG or on Performances.  This is determined by the PerformanceSelectChannel setting in minidexed.ini or the UI setting under "Performances -> MIDI PC" which determines if Program Change messages act on voices or Performances.

PerformanceSelectChannel is set as follows:
0 (default) - Program Change (and PGM buttons) act on voices on the MIDI channels associated with specific TGs.
1-16 - Program Change and PGM buttons) act on Performances.  Only Program Change messages on the specified channel are recognised, all others are ignored.
>16 - As above but in Omni mode, so Program Change messages on ANY channels will select on performances.

This interacts with other config settings as follows:
MIDIRXProgramChange still has to be 1 for any Program Change messages to be recognised at all.
ExpandPCAcrossBanks still maps PCs 1-128 across four voice banks started from the selected bank. It has no effect if selecting performances.
When PerformanceSelectChannel is 0 then Program Change messages will only be received on the MIDI channels set up for the TGs themselves.

Note that only the first 128 performances can be selected over MIDI as we have no concept of LSB/MSB or banks for performances.

Limitations:
- The UI for performances still doesn't update as perhaps wanted, but the [L] loaded indication does update.
- Selecting performances will change the MIDI configuration of the TGs, but not affect the PerformanceSelectChannel setting.
- It might be nice at some point to create a system-setup.ini for the system-level parameters as described above.
- There is no way to change between PerformanceSelectChannel modes over MIDI (it is unconfirmed if virtually "pressing the button on the original keyboard" was possible over SysEx).
- The default performance.ini is still confusing for people when selecting voices rather than performances.

I think this is about the shape of it.

Anyone fancy giving it a go!!? :)

When this gets to the point where we're happy with it and it is merged in then I'll attempt to update the wiki along the above lines!

Kevin

probonopd:
> Note that only the first 128 performances can be selected over MIDI as we have no concept of LSB/MSB or banks for performances.

This part I don't understand.

Why can't we just use the same MSB and LSB program change messages for voices and performances, depending on whether we are in "voice switching mode" or "performance switching mode"?

diyelectromusic:
> Why can't we just use the same MSB and LSB Program Change (PC) for voices and performances, depending on whether we are in "voice switching mode" or "performance switching mode"?

Because we'd have to implement performance "banks" first which is something we don't have at all in the system.  If you think we need that then that is something for a different PR as it will be quite an extensive set of changes.

> Maybe we should make it switch performances rather than voices by default now?

I still think that only plays to how some people will be using MiniDexed and takes it away from being a DX7(II) in a box as the default DX7II mode is to switch voices not performances.  The DX7s are a multi-timbral synth first and foremost, whereas treating performances as the "native voice" is more like treating it as a 8x6-operator massively polyphonic single FM synth.  As I understand things, most of the performances that are shipped with MiniDexed are effectively single combined voices.

But you can set the default in minidexed.ini however you wish.  There will still be some confusion I feel just from the terminology if nothing else - people will almost certainly (as a default) expect Program Change to change the FM voices used on the MIDI channel they use to send notes to a TG.  But if we default to performances that only makes sense if all TGs have the same MIDI channel which kind of becomes "the MIDI channel of the synth" (but it has to be configured in 9 different places - 8 TGs and the performance select channel).

It is perhaps more in keeping with what MiniDexed is now to treat performances this way, but at this point it is diverging from the nature of the original synths....  It might be good for an audience not knowing anything about a DX7 and wanting a "cool FM synth on a Pi" but all the reputation of MiniDexed was based on being "DX7s in a box" so I'd advise caution about diverging too far away from what the DX7s actually were...

Kevin

probonopd:
Yes, MiniDexed started out as a "DX7 or TX812 in a box" but the more I learn, the more I think we actually want a "DX7IId or TX802 in a box". But hey, all of this is new to me since I never had the original hardware back in the day to begin with.

Yes, I think we need performance "banks" because @Banana71 alone is easily providing more than 127 of them. But I agree that's for another PR then. I will open an issue so that it won't be forgotten.

@diyelectromusic kindly let me know once you think this PR is ready for being merged.

diyelectromusic:
I think I'm done with it now, but it would be useful for someone to give it a spin too! :)

probonopd:
Build for testing: [MiniDexed_2023-08-12-c598e2e](https://github.com/probonopd/MiniDexed/suites/15042348813/artifacts/858985697)

@Banana71 it'd be very helpful if you could test the heck out of this, as you know what to expect ;-)

Banana71:
**My RPi 4 does not start with the MiniDexed_2023-08-12-c598e2e version.** (I can't easily connect a monitor to HDMI).
With an RPi 3 everything works as described.

With the ExpandPCAcrossBanks=1 function, the Voices Bank are loaded across the board, but the bank is not updated in the UI. This means that if I load a Voice greater than 32 from Bank A, the display stays on Bank A instead of switching to Bank B, C or D. I have to get used to it first. However, it is clear to me that you cannot update the bank because otherwise there would be no meaningful scrolling up.


> But if we default to performances that only makes sense if all TGs have the same MIDI channel which kind of becomes "the MIDI channel of the synth" (but it has to be configured in 9 different places - 8 TGs and the performance select channel).


I don't understand you there @diyelectromusic 

I think a combined program change function is still the most sensible.

Here is an example: 
- PerformanceSelectChannel=1
- Controller Keyboard sends to a DAW on MIDIChannel 1. 
- miniDexed receives data from the DAW. 
- Controller keyboard sends program change command on channel 1 and the performance is changed in the miniDexed.
- The DAW sends sequences on MIDI channel 3, 4 and 5 plays different instruments on the miniDexed

Performance: 
- TG1 and TG2 play an electric piano on MIDI channel 1.
- TG3 and TG4 - MIDI Channel 3 - Strings
- TG5 and TG6 - MIDI channel 4 - bass
- TG 7 - MIDI channel 5 - FX sounds for a bridge

For the next bridge in the song, the possibility to change only the sound of TG7 would be desirable.
Of course I know in which bank I can find the right sound for the song on my miniDexed.

diyelectromusic:
> With the ExpandPCAcrossBanks=1 function, the Voices Bank are loaded across the board, but the bank is not updated in the UI. This means that if I load a Voice greater than 32 from Bank A, the display stays on Bank A instead of switching to Bank B, C or D. I have to get used to it first. However, it is clear to me that you cannot update the bank because otherwise there would be no meaningful scrolling up.

Yes, I think this is one case where people will largely be doing one or the other - i.e. either running it as a "MIDI voice bank of 128 voices" or treating it as banks/voices in their own right (either via DAW or UI).  (BTW - the expand function is already in there, that isn't part of this PR, just the bit about performances).
 
> > But if we default to performances that only makes sense if all TGs have the same MIDI channel which kind of becomes "the MIDI channel of the synth" (but it has to be configured in 9 different places - 8 TGs and the performance select channel).
> 
> I don't understand you there @diyelectromusic

I was attempting to show how the terminology is going to confuse.  In this mode it is largely acting as a single synth - changing performances changes all TGs etc, but the MIDI configuration could still be split causing confusion as bits might be acting on their own...

> I think a combined program change function is still the most sensible.
> 
> Here is an example:
> 
> * PerformanceSelectChannel=1
> * Controller Keyboard sends to a DAW on MIDIChannel 1.
> * miniDexed receives data from the DAW.
> * Controller keyboard sends program change command on channel 1 and the performance is changed in the miniDexed.
> * The DAW sends sequences on MIDI channel 3, 4 and 5 plays different instruments on the miniDexed

I like the idea, but I really don't think we can do this whilst MIDI channels (and voice/banks for that matter, although they are ignored) are part of the performance configuration.  You might think you have TGs responding on MIDI channels 3,4,5 but if you load a performance with a different MIDI config you're a bit stuffed :)

I think you have to mentally model the device either as a multi-timbral 8 TGs device, giving each TG its own setting; or a single 8x6 operator device where you assume the performance will be the entirety of what you need to load to make it do its thing.

As I say, if we somehow separate out the "system configuration" from the "performance configuration" then what you are suggesting probably makes more sense (to me at least), but right now they are combined and I think the only safe thing is to force the user to look at the device one way or the other... a hybrid just leaves too many avenues for getting the UI/DAW/sound/MIDI config out of sync in my view and leaving them wondering why nothing is responding anymore - I already had that problem myself during testing and I'd like to think I know how it is supposed to be working! :)

Kevin

diyelectromusic:
> **My RPi 4 does not start with the MiniDexed_2023-08-12-c598e2e version.** (I can't easily connect a monitor to HDMI). With an RPi 3 everything works as described.

Ah, failing an assert in MiniDexed SetParameter... probably a timing thing on a Pi4, I'll take a look (all my testing tends to be on a Pi 3 as its easier to plug into screens!)

Kevin

diyelectromusic:
Right an update is now building. I've also had to change the text for the Performance menu from "MIDI PC" to "PCCH" to make it shorter so it doesn't wrap on a HD44780 display - again my Pi 3 has a SSD1306 display which is two characters longer! My Pi 4 setup is using the HD44780, so note to self: don't be lazy and test on the PI 4 too :)

Kevin

probonopd:
Build for testing: [MiniDexed_2023-08-13-f6a53c2](https://github.com/probonopd/MiniDexed/suites/15053918582/artifacts/859650509)

probonopd:
RPi 3, seems to work.
It would be nice if the perfomance loading menu would show the loaded performance like it works for the voices (currently it only shows whether "[L]" or not), but I guess that is another story.
My use case: I have + and - buttons on my MIDI keyboard that I can use to send Program Change messages. When doing so, I like to look at the display to see which voice (works) or performance (does not work currently) I have selected.

diyelectromusic:
It would be nice, but I refer you to my previous comments:

>>PerformanceProgramChange=1 changes the performance according to MIDI program change but the UI is not updated.

> I think this was because of how the performance menus are structured. The voice menus work on a "if it is visible it is selected" basis but the performance is on a "you have to choose it to load it" basis, so the visible one isn't the selected one until it is actually selected. So atm the performance menu /is/ updated (the [L] symbol moves) but the name doesn't auto-scroll to the active one. I'm not sure if that is actually possible tbh with how it is structured, but I'll take another look...

And after taking another look concluded:

> The UI for performances still doesn't update as perhaps wanted, but the [L] loaded indication does update.

I think it is a different issue and not for this change to fix the inconsistency in voices vs performances in the menus.  Might be worth going back to the original author of the performance menus and asking why they were different to make sure we're not missing something?

(this change does seem to be prompting a lot of discussion about things that weren't actually in scope in the first place!)

Kevin

diyelectromusic:
> In terms of documentation, it will expand the comments in the minidexed.ini file:

Wiki now updated in hardware and files sections.  Probably should expand on the MIDI implementation section at some point too, but I need to think about how best to describe it there, so haven't done that yet.

Kevin

probonopd:
Thank you very much @diyelectromusic

probonopd:
Looks like there was no universal standard how to do this. Found this in the Yamaha manual for the PF2000, one of their lesser known FM instruments from the 80s:

image

If you ask me, this is a _really_ awkward way of doing it. Especially since the numbering on the cartridge is not a 1:1 match to the numbering of the MIDI messages.

diyelectromusic:
All part of the reasons for inventing General MIDI... :)

PR #531, opened 2023-08-11
Print performance number

probonopd:
In the Performance menu, print the number of the performance.

https://github.com/probonopd/MiniDexed/discussions/414#discussioncomment-6697774


probonopd:
[MiniDexed_2023-08-11-15bc1a8](https://github.com/probonopd/MiniDexed/suites/15008148433/artifacts/856729664)

Banana71:
That was very fast :-)

The numbering is not displayed when the performance is selected, only when it is loaded.
The numbering is wrong, the number from the file name is not displayed but a consecutive number.

Banana71:
MiniDexed_2023-08-11-8a2c8e2 shows the whole file name including the file extension

probonopd:
[MiniDexed_2023-08-12-9ebbff6](https://github.com/probonopd/MiniDexed/suites/15035774283/artifacts/858585978)

Banana71:
Thank you @probonopd ,

The numbering is correctly taken from the file name.
The performance number is displayed on the second line before the performance name.
The file name is shortened by the corresponding digits of the numbering.

miniDexed Display s

probonopd:
Looks like we have 3 more characters than on the OLED than on the 16x2 display. We should probably make use of them. :)

Banana71:
> Looks like we have 3 more characters available on the OLED than on the 16x2 display. We should probably make use of them. :)

20 - 16 = 3 😂🤷‍♂️🥰🖖🏼

probonopd:
Don't forget the `<` and `>`
So I was assuming 16-2 = 14. But on the OLED we have more (depending on the font used).

probonopd:
@diyelectromusic do we have "if OLED then use more characters" like logic in other places already?

diyelectromusic:
> @diyelectromusic do we have "if OLED then use more characters" like logic in other places already?

There is a configuration option to extend the width for oled displays: https://github.com/probonopd/MiniDexed/wiki/Files#ssd1306-oled

But it isn't ideal. I seem to recall there is at least one place where there were hard-coded 16-characters, but can't remember exactly where.... might be something like the volume control.

Kevin

probonopd:
Closing in favor of https://github.com/probonopd/MiniDexed/pull/581, thanks @diyelectromusic

Issue #562, opened 2023-11-17
Modulation Range Error


Banana71:
If any modulation range is changed, the modulation can no longer be heard; the modulation only works again after blindly saving and reloading the performance.
You can easily try it out with the performance 000011_SeventyOne, for example. Simply turn the modulation wheel all the way up until the sound has a lot of vibrato, then change the modulation range for the modulation wheel from 99 to 98 in the settings of TG1 and TG2 and you can no longer hear vibrato.
The problem occurs with all types of modulation. (ModWheel, Breath Controller, Foot Controller, Aftertouch)

probonopd:
Thanks for reporting this issue @Banana71. Do you happen to know whether this works correctly in Dexed on the PC?

Banana71:
In Dexed everything works as it should. Range changes are applied immediately, even when tones are pressed and held.

probonopd:
@dcoredump can you reproduce this on MicroDexed?

dcoredump:
I have to check this, give me some time. I think I know what's going wrong...

Thanks @Banana71 for deep testing and finding problems I never thought they may exist 😁

dcoredump:
I created a PR. Hope this fixes the problems. Cannot test - no keyboard right here.

Banana71:
Hi Holger @dcoredump , 
Unfortunately the error is still there

probonopd:
@Banana71 please try this build:
[MiniDexed_2023-11-20-8d1e592](https://github.com/probonopd/MiniDexed/suites/18351374688/artifacts/1062168230)

It is built using the fixed version from the posting above.

Banana71:
The error has changed.
I set the Mod.Wheel to MAX, the sound has full vibrato. After adjusting the Mod.Wheel range from e.g. 99 to 80, the vibrato remains at MAX and the vibrato stops completely when you move the Mod.Wheel.
In other words: you can adjust the range but the modulation remains at the old value. After moving the Mod.Wheel the modulation goes off and can no longer be restored.

dcoredump:
Thanks for testing!

Ok, I need to get my MIDI keyboard from the rehearsal room to my lab at home for deeper inspection. I hope I can try to dive in deeper at next weekend.

dcoredump:
Hi Peter,

[...]

PR #565, opened 2023-11-20
Hopefully a fix for "Modulation Range Error #562"

dcoredump:
I currently have no keyboard right here, so I cannot check the difference between MiniDexed and MicroDexed, sorry. But looking at the code I saw that I removed the calls to `ControllersRefresh()` when changing the range or target of the controllers in MicroDexed. So I made comments for these calls in MiniDexed, hopefully the problems are fixed.


probonopd:
Thanks @dcoredump. Unfortunately something is still wrong according to the test in https://github.com/probonopd/MiniDexed/issues/562#issuecomment-1819646675, so I am not merging this yet.

probonopd:
Hello @dcoredump. You suspect the issue might be in MiniDexed, do you have a suspicion where we should look? Thanks.

dcoredump:
> Hello @dcoredump. You suspect the issue might be in MiniDexed, do you have a suspicion where we should look? Thanks.

I think it must be MiniDexed because MicroDexed has no problems. [This](https://codeberg.org/dcoredump/MicroDexed/src/commit/9e4c6071c5d3b3abdbef3e69f5d7b3c5a3f69c9d/MicroDexed.ino#L1075) is the place where the modwheel code is used in MicroDexed.

I am currently _very_ short of time and cannot look into MiniDexed code - sorry.

Hope this helps,

Holger

Issue #547, opened 2023-09-30
USB Gadget Support


diyelectromusic:
I'm just parking this here for now as a reminder to take a look at some point: https://github.com/rsta2/circle/discussions/387

This would be nice to include once the bugs have been shaken out.

Kevin

probonopd:
Interesting! Looks like `CUSBMIDIGadget` landed in Circle [release 45.3.1](https://github.com/rsta2/circle/releases/tag/Step45.3.1).

probonopd:
Let's see the answers to https://github.com/rsta2/circle/discussions/387#discussioncomment-7412396

diyelectromusic:
Yep.  I'll take a look when I've stopped messing around with 16 tone generators! :)

I don't think its a big change - all the hard work is done in circle.

Kevin

diyelectromusic:
BTW @probonopd would you like to see if MiniDexed could request a VID/PID for use as a USB Device?  I can make things up for experimenting, but for proper builds there ought to be a proper code - the alternative is that users have to find their own, but they can't be set without recompilation at present so that would really limit their use!

Options seem to be:
- But one from USB foundation.  Cost ~$6000... :)
- Via Raspberry Pi, but this looks more for people building actual products on a Pi: https://github.com/raspberrypi/usb-pid
- Via pid.codes which looks more for open source projects: https://pid.codes/howto/
- OpenMoko appear to be allowing their VID for use with open source projects too: https://wiki.openmoko.org/wiki/USB_Product_IDs

I think we only need one VID/PID as all our builds are for a class compliant MIDI USB device, so driver wise, it should all be fine at the OS end...

I could do it of course, but I thought a request from your own account would be more appropriate?

Kevin

probonopd:
Wouldn't it be sufficient to have a USB VID/PID for Circle?

diyelectromusic:
That's a question for Rene, really, but as he had the choice already and has explicitly decided not to...

In future I'd imagine Circle may support other USB device modes, not just MIDI, so I think the right place is in the projects that are using circle rather than the core code itself.

Kevin

probonopd:
Applied at https://github.com/pidcodes/pidcodes.github.com/pull/895. Thanks @diyelectromusic

diyelectromusic:
Update:

I now have USB Gadget mode working on the following that I've been able to test:
* Raspberry Pi A (kernel.img)
* Raspberry Pi 3A+ (kernel8.img)
[...]

PR #567, opened 2023-11-21
Add USB Gadget Device support

diyelectromusic:
This fixes issue https://github.com/probonopd/MiniDexed/issues/547 and provides an option to enable USB gadget support in MiniDexed on those Raspberry Pi hardware devices that support it.

This is configuring using the following option in MiniDexed (assumed set to 0, off by default):
USBGadget=1

Note: This is not being added to MiniDexed as a comment though as USB Gadget support should not be enabled without full understanding of the hardware implications around powering the RPi and so on.  Consequently, this will all be explained on the wiki and considered an "advanced" feature of MiniDexed.

USB Gadget support in Circle is still a little experimental, but this has been initially tested and shown to work on the following:
* Raspberry Pi A (kernel.img)
* Raspberry Pi 3A+ (kernel8.img)
* Raspberry Pi 4B (kernel8-rpi4.img)
* Raspberry Pi Zero (kernel.img)
* Raspberry Pi Zero W (kernel.img)
* Raspberry Pi Zero 2 (kernel8.img)

It should work on a Pi A+ too, but I don't have one to test.  I don't think there is a Pi V2 that supports USB Gadget mode.

If Gadget mode is enabled on a Pi that doesn't support it then MiniDexed will halt and there will be messages logged on the console explaining why.  It would be possible to allow MiniDexed to start and run normally but with no USB support (we could just not return False after initialising USB in kernel.cpp), but it was felt that this would just hide the configuration problem and cause more user confusion than forcing a halt.

There is some discussion of the implications here: https://github.com/probonopd/MiniDexed/issues/547#issuecomment-1793866888

IMPORTANT: This requires an upgrade to the latest version of Circle and consequently circle-stdlib.  But it also uses the first new commit to the development version of Circle after the release as that includes the check of USB Gadget functionality to Pi hardware,  submod.sh has been updated accordingly.

Also note: USBID.sh contains the now officially allocated (by pid.codes) MiniDexed USB Vendor and Device IDs.  These should only be changed if replaced by another officially allocated USB Vendor and Device ID.  These should not just be "made up".

Kevin


BobanSpasic:
Any chance to have two separate modes, selectable at booting over a switch or something?
Namely, it would be interesting to have one mode where the MiniDexed is a SD Card reader, to expose the SD Card to the PC.
It would be a blessing for people (like me) who needs to copy files to the card all the time.

diyelectromusic:
I see no reason in the future why we couldn't explore USB mass storage support at some point... but we need to wait for it to appear in Circle first :)

Kevin

diyelectromusic:
Instructions now added to the Wiki, see: https://github.com/probonopd/MiniDexed/wiki/Hardware#usb-gadget-mode

Kevin

probonopd:
Thank you very much @diyelectromusic.

Build for testing:
MiniDexed_2023-11-21-669a8ac

diyelectromusic:
> Build for testing

I've done some very basic testing on the following:
* Pi Zero (USB Gadget)
* Pi 1 (USB Host)
* Pi 3A+ (USB Gadget/USB Host)
* Pi 3B (USB Host)
* Pi 4 (USB Gadget/USB Host)

And USB functionality seems to be working as I'd expect it to.

Kevin

Banana71:
My Akai MPC One recognizes the miniDexed RPi 3A+ as MIDIGadget and they work great together.
This function expands the area of application enormously and makes many things easier, thank you. 

For your information: The entry `USBGadget= ` is missing in the minidexed.ini and must be entered manually.

diyelectromusic:
> For your information: The entry `USBGadget= ` is missing in the minidexed.ini and must be entered manually.

Yes.  That is deliberate.  I don't want people experimenting with enabling this without first understanding the issues around how to power the Pi and the importance of not trying to use a USB gadget link and separate power at the same time.

So for the hardware reasons (which are RPi reasons, not MiniDexed reasons), although this is a nicer option for many cases, I regard this as an advanced option and thus would like to reserve it for those who have found it explicitly, and therefore have had at least a reasonable chance of reading, the wiki.

Kevin

probonopd:
Thank you very much @diyelectromusic!

PR #575, opened 2023-12-02
Make performance directory switchable

probonopd:
With [so many performances](https://github.com/probonopd/MiniDexed/discussions/560) to choose from, we need a way to switch between multiple directories (folders) of performances.

Intended behavior:
* We can have `performance/000000_factory/`, `performance/000001_somefolder/,  `performance/000002_someotherfolder/` instead of a hardcoded `performance/` directory
* By sending the MIDI Bank Change messages on the MIDI channel configured for selecting performances (rather than voices), the directory can be switched
* It can also be switched in the UI

TODO:
- [ ] The changes in `performanceconfig.*` need to be reviewed; in `CPerformanceConfig::SetPerformanceDir` the performances from the new performance directory need to be loaded
- [ ] The changes in `src/mididevice.cpp` are pseudo-code and need to be implemented properly
- [ ] The Performances menu entry should get a new submenu to change the performance directory without using MIDI bank change messages

**This PR is currently not functional because the TODOs need to be implemented. Any help appreciated. Volunteers?**


probonopd:
Closing in favor of @diyelectromusic's solution that has been merged.

PR #581, opened 2023-12-05
Performance file handling

probonopd:
**Build for testing.**

For details please see https://github.com/probonopd/MiniDexed/discussions/580.

Not ready for merging yet, this PR only exists so that we have builds for testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,13 +67,15 @@ jobs:
            echo "usbspeed=full" > sdcard/cmdline.txt
            cd sdcard
            cp ../kernels/* . || true
-           zip -r ../MiniDexed_$GITHUB_RUN_NUMBER_$(date +%Y-%m-%d)-$(git rev-parse --short HEAD).zip *
-           echo "artifactName=MiniDexed_$GITHUB_RUN_NUMBER_$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)" >> $GITHUB_ENV
            cd -
     - name: Get performance files
       run: |
            git clone https://github.com/Banana71/Soundplantage --depth 1 # depth 1 means only the latest commit
            cp -r ./Soundplantage/performance ./Soundplantage/*.pdf ./sdcard/
+           cd sdcard
+           zip -r ../MiniDexed_$GITHUB_RUN_NUMBER_$(date +%Y-%m-%d)-$(git rev-parse --short HEAD).zip *
+           echo "artifactName=MiniDexed_$GITHUB_RUN_NUMBER_$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+           cd -
     - uses: actions/upload-artifact@v3
       with:
         name: ${{ env.artifactName }} # Exported above
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -184,9 +184,31 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 	else
 	{
 		// Perform any MiniDexed level MIDI handling before specific Tone Generators
+		unsigned nPerfCh = m_pSynthesizer->GetPerformanceSelectChannel();
 		switch (ucType)
 		{
 		case MIDI_CONTROL_CHANGE:
+			// Check for performance PC messages
+			if (nPerfCh != Disabled)
+			{
+				if ((ucChannel == nPerfCh) || (nPerfCh == OmniMode))
+				{
+					if (pMessage[1] == MIDI_CC_BANK_SELECT_MSB)
+					{
+						m_pSynthesizer->BankSelectMSBPerformance (pMessage[2]);
+					}
+					else if (pMessage[1] == MIDI_CC_BANK_SELECT_LSB)
+					{
+						m_pSynthesizer->BankSelectLSBPerformance (pMessage[2]);
+					}
+					else
+					{
+						// Ignore any other CC messages at this time
+					}
+				}
+			}
+			break;
+
 		case MIDI_NOTE_OFF:
 		case MIDI_NOTE_ON:
 			if (nLength < 3)
@@ -195,11 +217,11 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 			}
 			m_pUI->UIMIDICmdHandler (ucChannel, ucStatus & 0xF0, pMessage[1], pMessage[2]);
 			break;
+
 		case MIDI_PROGRAM_CHANGE:
 			// Check for performance PC messages
 			if( m_pConfig->GetMIDIRXProgramChange() )
 			{
-				unsigned nPerfCh = m_pSynthesizer->GetPerformanceSelectChannel();
 				if( nPerfCh != Disabled)
 				{
 					if ((ucChannel == nPerfCh) || (nPerfCh == OmniMode))
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -207,6 +207,10 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 					}
 				}
 			}
+			if (nLength == 3)
+			{
+				m_pUI->UIMIDICmdHandler (ucChannel, ucStatus & 0xF0, pMessage[1], pMessage[2]);
+			}
 			break;
 
 		case MIDI_NOTE_OFF:
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -227,6 +227,7 @@ bool CMiniDexed::Initialize (void)
 		reverb_send_mixer->gain(i,mapfloat(m_nReverbSend[i],0,99,0.0f,1.0f));
 	}
 
+	m_PerformanceConfig.Init();
 	if (m_PerformanceConfig.Load ())
 	{
 		LoadPerformanceParameters(); 
@@ -234,12 +235,6 @@ bool CMiniDexed::Initialize (void)
 	else
 	{
 		SetMIDIChannel (CMIDIDevice::OmniMode, 0);
-	}
-	
-	// load performances file list, and attempt to create the performance folder
-	if (!m_PerformanceConfig.ListPerformances()) 
-	{
-		LOGERR ("Cannot create internal Performance folder, new performances can't be created");
 	}
 	
 	// setup and start the sound device
@@ -392,6 +387,11 @@ CSysExFileLoader *CMiniDexed::GetSysExFileLoader (void)
 	return &m_SysExFileLoader;
 }
 
+CPerformanceConfig *CMiniDexed::GetPerformanceConfig (void)
+{
+	return &m_PerformanceConfig;
+}
+
 void CMiniDexed::BankSelect (unsigned nBank, unsigned nTG)
 {
 	nBank=constrain((int)nBank,0,16383);
@@ -413,10 +413,11 @@ void CMiniDexed::BankSelectPerformance (unsigned nBank)
 
 	if (m_nParameter[ParameterPerformanceSelectChannel] != CMIDIDevice::Disabled)
 	{
-	//	if (GetSysExFileLoader ()->IsValidBank(nBank))
+		if (GetPerformanceConfig ()->IsValidPerformanceBank(nBank))
 		{
 			// Only change if we have the bank loaded
 			m_nVoiceBankIDPerformance = nBank;
+			GetPerformanceConfig ()->SetPerformanceBank (nBank);
 
 			m_UI.ParameterChanged ();
 		}
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -489,10 +489,7 @@ void CMiniDexed::ProgramChangePerformance (unsigned nProgram)
 	if (m_nParameter[ParameterPerformanceSelectChannel] != CMIDIDevice::Disabled)
 	{
 		// Program Change messages change Performances.
-		unsigned nLastPerformance = m_PerformanceConfig.GetLastPerformance();
-
-		// GetLastPerformance actually returns 1-indexed, number of performances
-		if (nProgram < nLastPerformance - 1)
+		if (m_PerformanceConfig.IsValidPerformance(nProgram))
 		{
 			SetNewPerformance(nProgram);
 		}
@@ -1629,6 +1626,11 @@ std::string CMiniDexed::GetNewPerformanceDefaultName(void)
 void CMiniDexed::SetNewPerformanceName(std::string nName)
 {
 	m_PerformanceConfig.SetNewPerformanceName(nName);
+}
+
+bool CMiniDexed::IsValidPerformance(unsigned nID)
+{
+	return m_PerformanceConfig.IsValidPerformance(nID);
 }
 
 void CMiniDexed::SetVoiceName (std::string VoiceName, unsigned nTG)
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1546,6 +1546,11 @@ unsigned CMiniDexed::GetLastPerformance()
 	return m_PerformanceConfig.GetLastPerformance();
 }
 
+unsigned CMiniDexed::GetPerformanceBank()
+{
+	return m_PerformanceConfig.GetPerformanceBank();
+}
+
 unsigned CMiniDexed::GetLastPerformanceBank()
 {
 	return m_PerformanceConfig.GetLastPerformanceBank();
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -53,8 +53,11 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	m_bSavePerformance (false),
 	m_bSavePerformanceNewFile (false),
 	m_bSetNewPerformance (false),
+	m_bSetNewPerformanceBank (false),
+	m_bSetFirstPerformance (false),
 	m_bDeletePerformance (false),
-	m_bLoadPerformanceBusy(false)
+	m_bLoadPerformanceBusy(false),
+	m_bLoadPerformanceBankBusy(false)
 {
 	assert (m_pConfig);
 
@@ -302,14 +305,30 @@ void CMiniDexed::Process (bool bPlugAndPlayUpdated)
 		m_bSavePerformanceNewFile = false;
 	}
 	
-	if (m_bSetNewPerformance && !m_bLoadPerformanceBusy)
+	if (m_bSetNewPerformanceBank && !m_bLoadPerformanceBusy && !m_bLoadPerformanceBankBusy)
+	{
+		DoSetNewPerformanceBank ();
+		if (m_nSetNewPerformanceBankID == GetActualPerformanceBankID())
+		{
+			m_bSetNewPerformanceBank = false;
+		}
+		
+		// If there is no pending SetNewPerformance already, then see if we need to find the first performance to load
+		// NB: If called from the UI, then there will not be a SetNewPerformance, so load the first existing one.
+		//     If called from MIDI, there will probably be a SetNewPerformance alongside the Bank select.
+		if (!m_bSetNewPerformance && m_bSetFirstPerformance)
+		{
+			DoSetFirstPerformance();
+		}
+	}
+	
+	if (m_bSetNewPerformance && !m_bSetNewPerformanceBank && !m_bLoadPerformanceBusy && !m_bLoadPerformanceBankBusy)
 	{
 		DoSetNewPerformance ();
 		if (m_nSetNewPerformanceID == GetActualPerformanceID())
 		{
 			m_bSetNewPerformance = false;
 		}
-		
 	}
 	
 	if(m_bDeletePerformance)
@@ -419,7 +438,7 @@ void CMiniDexed::BankSelectPerformance (unsigned nBank)
 		{
 			// Only change if we have the bank loaded
 			m_nVoiceBankIDPerformance = nBank;
-			GetPerformanceConfig ()->SetPerformanceBank (nBank);
+			SetNewPerformanceBank (nBank);
 
 			m_UI.ParameterChanged ();
 		}
@@ -1566,6 +1585,16 @@ void CMiniDexed::SetActualPerformanceID(unsigned nID)
 	m_PerformanceConfig.SetActualPerformanceID(nID);
 }
 
+unsigned CMiniDexed::GetActualPerformanceBankID()
+{
+	return m_PerformanceConfig.GetActualPerformanceBankID();
+}
+
+void CMiniDexed::SetActualPerformanceBankID(unsigned nBankID)
+{
+	m_PerformanceConfig.SetActualPerformanceBankID(nBankID);
+}
+
 bool CMiniDexed::SetNewPerformance(unsigned nID)
 {
 	m_bSetNewPerformance = true;
@@ -1574,11 +1603,18 @@ bool CMiniDexed::SetNewPerformance(unsigned nID)
 	return true;
 }
 
-unsigned CMiniDexed::SetFirstPerformance(void)
+bool CMiniDexed::SetNewPerformanceBank(unsigned nBankID)
 {
-	unsigned nID = m_PerformanceConfig.FindFirstPerformance();
-	SetNewPerformance(nID);
-	return nID;
+	m_bSetNewPerformanceBank = true;
+	m_nSetNewPerformanceBankID = nBankID;
+
+	return true;
+}
+
+void CMiniDexed::SetFirstPerformance(void)
+{
+	m_bSetFirstPerformance = true;
+	return;
 }
 
 bool CMiniDexed::DoSetNewPerformance (void)
@@ -1600,6 +1636,25 @@ bool CMiniDexed::DoSetNewPerformance (void)
 		m_bLoadPerformanceBusy = false;
 		return false;
 	}
+}
+
+bool CMiniDexed::DoSetNewPerformanceBank (void)
+{
+	m_bLoadPerformanceBankBusy = true;
+	
+	unsigned nBankID = m_nSetNewPerformanceBankID;
+	m_PerformanceConfig.SetNewPerformanceBank(nBankID);
+	
+	m_bLoadPerformanceBankBusy = false;
+	return true;
+}
+
+void CMiniDexed::DoSetFirstPerformance(void)
+{
+	unsigned nID = m_PerformanceConfig.FindFirstPerformance();
+	SetNewPerformance(nID);
+	m_bSetFirstPerformance = false;
+	return;
 }
 
 bool CMiniDexed::SavePerformanceNewFile ()
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -170,6 +170,8 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	SetParameter (ParameterCompressorEnable, 1);
 
 	SetPerformanceSelectChannel(m_pConfig->GetPerformanceSelectChannel());
+		
+	SetParameter (ParameterPerformanceBank, 0);
 };
 
 bool CMiniDexed::Initialize (void)
@@ -830,6 +832,10 @@ void CMiniDexed::SetParameter (TParameter Parameter, int nValue)
 
 	case ParameterPerformanceSelectChannel:
 		// Nothing more to do
+		break;
+
+	case ParameterPerformanceBank:
+		BankSelectPerformance(nValue);
 		break;
 
 	default:
@@ -1533,6 +1539,11 @@ unsigned CMiniDexed::GetLastPerformance()
 	return m_PerformanceConfig.GetLastPerformance();
 }
 
+unsigned CMiniDexed::GetLastPerformanceBank()
+{
+	return m_PerformanceConfig.GetLastPerformanceBank();
+}
+
 unsigned CMiniDexed::GetActualPerformanceID()
 {
 	return m_PerformanceConfig.GetActualPerformanceID();
@@ -1549,6 +1560,13 @@ bool CMiniDexed::SetNewPerformance(unsigned nID)
 	m_nSetNewPerformanceID = nID;
 
 	return true;
+}
+
+unsigned CMiniDexed::SetFirstPerformance(void)
+{
+	unsigned nID = m_PerformanceConfig.FindFirstPerformance();
+	SetNewPerformance(nID);
+	return nID;
 }
 
 bool CMiniDexed::DoSetNewPerformance (void)
@@ -1666,6 +1684,11 @@ void CMiniDexed::SetNewPerformanceName(std::string nName)
 bool CMiniDexed::IsValidPerformance(unsigned nID)
 {
 	return m_PerformanceConfig.IsValidPerformance(nID);
+}
+
+bool CMiniDexed::IsValidPerformanceBank(unsigned nBankID)
+{
+	return m_PerformanceConfig.IsValidPerformanceBank(nBankID);
 }
 
 void CMiniDexed::SetVoiceName (std::string VoiceName, unsigned nTG)
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -432,16 +432,13 @@ void CMiniDexed::BankSelectPerformance (unsigned nBank)
 {
 	nBank=constrain((int)nBank,0,16383);
 
-	if (m_nParameter[ParameterPerformanceSelectChannel] != CMIDIDevice::Disabled)
+	if (GetPerformanceConfig ()->IsValidPerformanceBank(nBank))
 	{
-		if (GetPerformanceConfig ()->IsValidPerformanceBank(nBank))
-		{
-			// Only change if we have the bank loaded
-			m_nVoiceBankIDPerformance = nBank;
-			SetNewPerformanceBank (nBank);
+		// Only change if we have the bank loaded
+		m_nVoiceBankIDPerformance = nBank;
+		SetNewPerformanceBank (nBank);
 
-			m_UI.ParameterChanged ();
-		}
+		m_UI.ParameterChanged ();
 	}
 }
 
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -407,6 +407,22 @@ void CMiniDexed::BankSelect (unsigned nBank, unsigned nTG)
 	}
 }
 
+void CMiniDexed::BankSelectPerformance (unsigned nBank)
+{
+	nBank=constrain((int)nBank,0,16383);
+
+	if (m_nParameter[ParameterPerformanceSelectChannel] != CMIDIDevice::Disabled)
+	{
+	//	if (GetSysExFileLoader ()->IsValidBank(nBank))
+		{
+			// Only change if we have the bank loaded
+			m_nVoiceBankIDPerformance = nBank;
+
+			m_UI.ParameterChanged ();
+		}
+	}
+}
+
 void CMiniDexed::BankSelectMSB (unsigned nBankMSB, unsigned nTG)
 {
 	nBankMSB=constrain((int)nBankMSB,0,127);
@@ -422,6 +438,12 @@ void CMiniDexed::BankSelectMSB (unsigned nBankMSB, unsigned nTG)
 	m_nVoiceBankIDMSB[nTG] = nBankMSB;
 }
 
+void CMiniDexed::BankSelectMSBPerformance (unsigned nBankMSB)
+{
+	nBankMSB=constrain((int)nBankMSB,0,127);
+	m_nVoiceBankIDMSBPerformance = nBankMSB;
+}
+
 void CMiniDexed::BankSelectLSB (unsigned nBankLSB, unsigned nTG)
 {
 	nBankLSB=constrain((int)nBankLSB,0,127);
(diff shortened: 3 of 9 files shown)


probonopd:
Build for testing:
[MiniDexed_2023-12-05-e5b2656](https://github.com/probonopd/MiniDexed/suites/18779064606/artifacts/1094269937)

diyelectromusic:
> What has changed?

The key change in this first version is to align the performance filename numbers with the performance numbers themselves, so that if there is a gap, the numbers still correspond to the right performances.

So

000002_Perf2
000003_Perf3
000005_Perf5
000008_Perf8

Will be selectable using MIDI PC voice instruments 2,3,5,8 (which "on the wire" is 1,2,4,7).  Previously, they were loaded and sorted in numerical order, but without leaving gaps, so the above would have been 2,3,4,5 - meaning any gaps in numbering would get "out of sync" with the filenames.

I considered sorting this out a pretty extensive change in terms of changing how performance files work (see the files changed), but need this done to be able to do the rest. I wanted to know this was done and working ok before changing anything else.

(There is more on the problem here: https://diyelectromusic.wordpress.com/2023/11/22/a-dx7-usb-dongle/)

Kevin

probonopd:
Build for testing:
[MiniDexed_2023-12-09-a79acdf](https://github.com/probonopd/MiniDexed/suites/18902914135/artifacts/1103790835)

diyelectromusic:
> Build for testing: [MiniDexed_2023-12-09-a79acdf](https://github.com/probonopd/MiniDexed/suites/18902914135/artifacts/1103790835)

There's nothing really to be testing yet, I'm just snapshotting a useful point for me to ensure everything is backed up online as things progress.  I'll keep tinkering on and off over the next week or so and shout when there is something worth looking at :)

Kevin

Banana71:
Okay, good to know.

diyelectromusic:
Latest "in development" build (for testing, not for using) now does the following:
* Needs PCCH set for this to work either via menus or by setting PerformanceSelectChannel to a valid MIDI channel (1-16).
* Subdirectories of performance directory with name in the format 01_Name will be turned into performance banks taking the user-facing number from the two digits (in decimal).
* Up to 64 banks are supported.
* Banks can be selected via MIDI only at this point.  Note the MIDI "on the wire" bank number is User Bank Number - 1 (as is typical for MIDI voices and banks).
* Bank 01 is reserved for the legacy performance directory itself, so additional banks should start from 02_....
* Bank 01/Performance 00001 is still reserved for the legacy performance.ini in the root.

Limitations:
* No UI handling of performance banks yet.
* If a performance is saved, then it can only add them on the end of the performance list, not in any empty slots.
* Very verbose on startup so that I can see what is loaded and when, so will scroll off the screen pretty quickly if you have lots of performances!

Changes in behaviour compared to existing releases:
* If the performance directory doesn't already exist, it no-longer creates it.
* Recall that my previous version made the numbers in performance names significant, so performance 00005_PerformanceName will be stored in "slot 5" (1-indexed) which can be selected by MIDI PC number 4 (0-indexed).

What needs testing if anyone fancies helping....

Check file/directory handling:
* Check range of filenames in both main performance directory and subdirectories.
* Check range of directory names.
* Check handling of duplicates.
* Check up to the limits as much as possible - note that in theory it supports up to 256 performances in each of up to 64 banks.
* Check gaps in numbering, subdirectories in subdirectories, and so on...

Check MIDI handling when PCCH=0:
* Check original voice/bank handling still works ok via UI.
* Check original voice/bank handling still works ok via MIDI.
* Check performance selection still works ok in UI.

Check MIDI handling when PCCH = valid MIDI channel:
* Check performance selection still works ok via MIDI.
* Change banks and recheck performance selection via MIDI.
* Keep changing banks/performances include changing back to those already selected.
* Check default bank/performance selection - bank 01/voice 01 should be the original performance.ini in the root.

General performance file handling:
* Check saving as new file and verifying saving has worked and is preserved on power up.
* Check saving overwriting and verifying saving has worked and is preserved on power up.
* Check deleting and verifying it has worked and is preserved on power up.
* Save multiple files before power cycling then verify all is good.
* Delete multiple files before power cycling then verify all is good.
* Do both saves and deletes before power cycling then verify all is good.
* Save a file and delete it before power cycling.

General MiniDexed use to see how it fares.

I think this is largely doing what I'm expecting but it would be good to get some proper verification going on before moving on to the last main phase - the UI stuff.

Report any errors, ideally letting me know what test you're doing, any relevant context (filenames, structure, MIDI msgs, screen output, etc) and what happened!  Especially any assets going off!

Thanks,
Kevin

probonopd:
Build for testing:
[MiniDexed_2023-12-16-3b2ae95](https://github.com/probonopd/MiniDexed/actions/runs/7232945364/artifacts/1119200878)

For the sake of consistency and simplicity we should probably drop the legacy stuff?

diyelectromusic:
> For the sake of consistency and simplicity we should probably drop the legacy stuff?

I almost proposed that, but I don't think we need to and it does allow someone to run it with a fairly minimal configuration.

Kevin

Banana71:
Thank you @diyelectromusic Kevin for your hard work.

I tested with my RPi4, SSD1306, GY-PCM5102, classic MIDI input.

### General performance file handling:

- [x] Check saving as new file and verifying saving has worked and is preserved on power up.
- [x] Check saving overwriting and verifying saving has worked and is preserved on power up.
- [x] Check deleting and verifying it has worked and is preserved on power up.
- [x] Save multiple files before power cycling then verify all is good.
- [x] Delete multiple files before power cycling then verify all is good.
- [x] Do both saves and deletes before power cycling then verify all is good.
- [x] Save a file and delete it before power cycling.

### Check MIDI handling when PCCH=0:

- [x] Check original voice/bank handling still works ok via UI.
- [x] Check original voice/bank handling still works ok via MIDI.
- [x] Check performance selection still works ok in UI.

I use my Keystep37 to control MIDI CC 32 (LSB Controller) and MIDI CC 0 (Bank Select). I'm not sure if I'm doing something wrong, the program change doesn't work as expected

pCCH=1, MIDI CC 0=0: then I can choose between the standard performance bank and the sub performance folder 02_Name with MIDI CC 32 =0 or 1. I would have expected to choose the performance with MIDI CC 32 and not the bank.

Which MIDI CC commands do I have to use exactly to select the performances in miniDexed? Or what tools do you use to choose the bank and performance?

Banana71:
With the Hydrasynth I can choose the banks and performances. Every now and then, graining occurs in the sound after a program change, but after switching back and forth several times, the graining disappears again.

diyelectromusic:
Banks are selected with a combination of the LSB and MSB for BankSelect, so MIDI CC 0 (LSB) and 32 (MSB).  But as there are only 64 banks supported, actually we don't need to use the MSB, but I think the MIDI spec says to always send them.  The performances within a bank are selected using Program Change messages.

If you use something like MidiOx then it will always send the combination of Bank MSB, Bank LSB, PC for any instrument selection, which again is what I think the MIDI spec suggests if using banks.

So the Banks will be numbered as follows:
MSB=0; LSB=0 - Bank 1 (the default performance set without any subdirectories) PC=0 to 127 for performances 001 (the default performance.ini); 002_Concert D; 003_The Grand; etc...
MSB=0; LSB=1 - Bank 2 - whatever is called 02_Some Performance Bank
... and so on down to ...
MSB=0; LSB=63 - Bank 64 - whatever is called 64_The Last Performance Bank

Kevin

Banana71:
>Banks are selected with a combination of the LSB and MSB for BankSelect, so MIDI CC 0 (LSB) and 32 (MSB).

I think you made a mistake there
https://anotherproducer.com/online-tools-for-musicians/midi-cc-list/
MIDI CC 0  (MSB)
MIDI CC 32 (LSB)

diyelectromusic:
Latest update: Performance bank selection now included in the performance menu.

Notes:
* I've tried to mirror the menu handling of selecting a performance - so it will respect the SelectPerformanceToLoad config setting, skip blank banks, wrap around, etc.
* On changing banks it will select the first valid performance it finds in that bank.
* No options to edit/save/delete/rename banks - it will only select.

So outstanding questions now:
* Is it ok just to implement save for the end of banks or do I need to do something more complex there?  Ideally it would find the first free slot, but in reality how many people are deleting performances for this to be an issue?
* Also, is it ok to allow anyone to delete/change any performances in any banks - or should editing be limited to a specific "user" bank to allow the "presets" to always be available?
* Is it ok to switch to the first valid performance in a bank when changing via the UI?  If there are gaps, then I'm not sure what else we can realistically do.  It isn't quite the same as voices - I know the current playing voice is left when changing banks (we had a big discussion about it somewhere, but I don't think we resolved what behaviour was "right" as different synths do different things).
* Is 64 the right number of banks and 256 the right number of performances?

As before, this needs some proper testing (see my previous list of ideas, which really need repeating!).  Also need to verify the operation of SelectPerformanceToLoad being 0 and 1 and make sure performance bank/voice loading from UI works as expected.

I still need to remove some debug strings, but they are quite useful atm for seeing what is going on, so they are still in for now.

Please let me know if you manage to give this a go.  I'm not sure how much time I'll have over the next week now myself.  Also if anyone feels like actually reviewing the code changes for any obvious things I've missed that would be cool too :)

I'm currently expecting a final update (to tidy up/remove debug strings/etc) after Christmas probably now.

Kevin

probonopd:
Hi Kevin, thanks for your work. Here are my 2 cents:

 > Is it ok just to implement save for the end of banks

Yes, I think that's ok. I don't assume that anyone will want to do more complicated "performance management" on the MiniDexed device itself.

> Ideally it would find the first free slot

Something to consider for later in case it turns out to be a commonly requested feature.

> Also, is it ok to allow anyone to delete/change any performances in any banks - or should editing be limited to a specific "user" bank to allow the "presets" to always be available?

Again, I'd say let's not worry about that unless it turns out to be a commonly requested feature. (We could always use e.g., a .factory suffix-suffix on all kinds of files and directories and prevent those from being deleted.)

> Is it ok to switch to the first valid performance in a bank when changing via the UI? 

Yes. The question is, when does the performance switching occur. As soon as you switch to another bank, or only once you select a performance in the bank? (Ideally the behavior between voices and performances could be made consistent in this regard. What do real instruments do?)

> Is 64 the right number of banks and 256 the right number of performances?

If you ask me, "unlimited" is the correct number of banks. (Not all of those will be addressable over MIDI, though.)

diyelectromusic:
> What do real instruments do?)

As I say, I think we ended up being inconclusive (more here: https://github.com/probonopd/MiniDexed/discussions/425).  But where we differ is that we might have a gap where an instrument is in another bank, so what happens then? Anyway, for now it switches to the first performance in the bank on bank select.  People can try it and see how that feels.

I tend to like the behaviour that changing banks changes to the same selected voice in the new bank.  But I think that comes from a keyboard (I forget which) where the same patch numbers across different banks have variations of a similar voice...

> If you ask me, "unlimited" is the correct number of banks. (Not all of those will be addressable over MIDI, though.)

Well as we're not really using any kind of dynamic memory management in MiniDexed all structures are pre-allocated, so we do need to pick a size and go with it :)

diyelectromusic:
> That's concerning me. Which memory structure are we talking about? Can't we read the filesystem as-needed?

Why?  Pretty much every structure in MiniDexed is pre-allocated - that's why we set number of voices, number of banks, number of tone generators... if we have several of something, the number we handle is in a header file somewhere.

There are some cases where the object oriented nature of C++ is used for more dynamic creation of objects, but generally speaking this mostly happens on startup from what I've seen.

Kevin

probonopd:
My comment is not specifically about this PR but is about how we handle files and directories in general. Whenever it comes to interacting with the filesystem, I think we should not hardcode things like the maximum number of files in a directory. Also, I think we should load files just-in-time when we need it (no sooner). If we did this, then the number of files on the SD card should not influence the boot time in any significant way. Does this make sense? (Again, not specific to this PR.)

diyelectromusic:
> My comment is not specifically about this PR but is about how we handle files and directories in general. Whenever it comes to interacting with the filesystem, I think we should not hardcode things like the maximum number of files in a directory. Also, I think we should load files just-in-time when we need it (no sooner). If we did this, then the number of files on the SD card should not influence the boot time in any significant way. Does this make sense? (Again, not specific to this PR.)

That's easy to say if we were running with a proper, full operating system, but with an embedded 'bare metal' OS, we have to think about such things a lot more up front.  That's just embedded systems.

In practical terms, we already know that having too many files in a directory causes everything to go really slowly due to the nature of FAT support in circle.  But ultimately whilst there are no limits to the numbers of files and directories on the filesystem, by the nature of the system we only pay attention to a fixed number of them.

And yes, "load at point of use" is pretty good when you can do it, but it is quite complicated to do.  And I do that (pretty much) for the performance banks anyway (a lot more so that we do with voices), but we still need some kind of "up front" knowledge about how many banks we know about so that we don't drag the performance of the UI to a halt - we don't want it parsing the filesystem for every click of the encoder... really, we don't :)

Kevin

probonopd:
Possibly :)

Banana71:
>So outstanding questions now:

Without reading the other answers first, I'll give you my unbiased opinion. 

>- Is it ok just to implement save for the end of banks or do I need to do something more complex there? Ideally it would find the first free slot, but in reality how many people are deleting performances for this to be an issue?

It's okay, but the optimal solution would be to freely choose the storage space. You could choose a free storage space or overwrite an existing performance. The cherry on the cake would be if you could also change banks when saving. This way you could easily create your own favorites bank directly on the device.

>- Also, is it ok to allow anyone to delete/change any performances in any banks - or should editing be limited to a specific "user" bank to allow the "presets" to always be available?

Yes, everyone should be able to do everything. The best solution might be to be able to set write protection on a folder (bank) yourself.

>- Is it ok to switch to the first valid performance in a bank when changing via the UI? If there are gaps, then I'm not sure what else we can realistically do. It isn't quite the same as voices - I know the current playing voice is left when changing banks (we had a big discussion about it somewhere, but I don't think we resolved what behaviour was "right" as different synths do different things).

Absolutely yes. And it's also ok that the last performance remains active.

>- Is 64 the right number of banks and 256 the right number of performances?

I think 128 performances per bank makes sense. Anything beyond that cannot be accessed via Midi Program Change. If you want to quickly scroll between performances using the encoder, you probably won't want to have more than 64 performances in a bank. A limit of 256 performances is okay, but I wouldn't use it myself.
What are the loading times when changing banks? Are all performances then loaded to implement the sorting?
Does the number of banks influence the boot time or when changing banks?
128 banks with a maximum of 128 performance are the MIDI limits.

I think you all know that I really like the miniDexed and I can work well with it as it is. Any improvement you make is fantastic. Thanks

diyelectromusic:
> I think 128 performances per bank makes sense.

Actually, I probably agree with you. I'll probably change it to 128.  We will need to renumber the default set though as some of those are currently numbered more than 128.

> What are the loading times when changing banks? Are all performances then loaded to implement the sorting?

Pretty quick in tests, but I've not had many performances in each bank.  I'm no longer sorting as the number is now significant, so the order is derived directly from the filenames in performances (and banks) and there might be gaps and that is fine.

> Does the number of banks influence the boot time or when changing banks? 128 banks with a maximum of 128 performance are the MIDI limits.

I don't think so. I have to scan the bank names/numbers on power up, that is all. It only scans the directory itself when changing banks.

I've not tested it to the full numbers yet.  Also, it will be slower in this version as it is printing a lot of debug info out to the console.

Kevin

probonopd:
> With MSB/LSB the MIDI limit is actually 16384 possible banks.

Let's use that then

diyelectromusic:
> > With MSB/LSB the MIDI limit is actually 16384 possible banks.
> 
> Let's use that then 👍

Really?  Over 2 million performance voices?  I suspect if you added up all the permutations of FM parameters for all 6 operators in all 8 TGs you'll still have spare performances... and it would take you something like a month (with no sleeping) to scroll through them all!!??

I suggest we stick to 128 banks of 128 performances, which is still over 16 thousand performances, and we can see how we go.

Kevin

diyelectromusic:
Right assuming this builds ok, I think I'm ok with this one now.  I've removed the debug messages, set things to 128 performances in up to 128 banks, and fixed a couple of bugs I'd spotted in UI handling and performance loading.

Bank names are now all in the format "123_Bank Name" to allow for banks 001 through to 128 with bank 001 as the legacy (i.e. in no subdirectory) bank.

I've testing the program change up/down now; the UI; invalid performance and bank numbers; MIDI selection.

I'd like others to give this a go with their own configurations to make sure the default modes haven't been messed up in any way.

Note: The default performance set includes performances numbered > 128 so these will need to be changed.

If this checks out ok with everyone, I'd be happy for this to be merged in now I think - but please do check your own configurations - as many as possible please!

Kevin

probonopd:
Build for testing:
[MiniDexed_2023-12-30-7637fc6](https://github.com/probonopd/MiniDexed/actions/runs/7364952597/artifacts/1140375931)

probonopd:
Hi @diyelectromusic, thanks again for your great work!

Personally I would say "let's remove the legacy cruft to make the system simpler and more consistent for the future" - what do others think?

diyelectromusic:
Well I've just found an issue if there is no performance directory, so it's tempting!

But as most of the code is in there and working ok, I'm not sure we gain much from removing it and it would be a shame to break all the existing installations out there just for this.

Kevin

probonopd:
Imho it's better "break the installations" one time (especially if the solution is very simple for users) rather than having to explain in the documentation why some files need to be in directories while some others don't forever.

diyelectromusic:
Well untangling all the legacy work is quite a task in its own right.  There are quite a few bits of MiniDexed that would be almost "start from scratch" areas if we were to remove the legacy behaviours and performance is probably top of the list...

That will have to be a separate change if people want it I think...  otherwise this PR will go on forever...

Kevin

diyelectromusic:
Ok, sorry - can you do another build?  I think this is really the one now (famous last words!).

Kevin

Banana71:
I tested the version "MiniDexed_2023-12-30-cd19344" and so far everything works as described. I will reduce the “default performances” to below 129.

When I want to save a new performance in the default bank, the error message "Error" (>128 performances) comes up. The only way to save a performance in the default bank is by overwrite it.
Maybe it would be better if no performance was loaded automatically when changing banks. So it would be possible to create a performance and select a bank before saving it and then save the new performance. This would also make it possible to copy performances between the benches.

diyelectromusic:
Whilst background processing, I've realised another minor bug on startup so will need to do an update anyway.

So do you think that changing banks should just leave the existing performance loaded?  The only issue with that is that the UI will be out of date when you go to the performance Load menu until the encoder is turned - the selected performance won't actually be the loaded performance.  And in fact, it might be that the loaded performance doesn't exist (numerically) in the new bank...

I'll code that in as a test and see if what I think will happen is really the case and see if there is anything that can be done that is better.

Basically, the issue is one of how to keep the UI accurately matching the state of the synth.  If a new bank is reloaded, the existing performance no longer exists in the UI sense.

Kevin

diyelectromusic:
Actually thinking some more - the way to fix this is to make the save work properly with banks now.  So when saving it should be possible to select which bank and then find a free slot in that bank, rather than only save on the end.

For now the work around is to leave free slots at the end of banks if editing performances.  I don't know how many people are regularly saving performances, but I suspect changing banks and selecting performances could be happening all the time.

Either way, I propose we log this as an issue that save isn't working fully yet and leave this for a future change.

Kevin

diyelectromusic:
I've just realised I can link to the builds myself :)
Here is the latest one: [MiniDexed_2023-12-31-814837b](https://github.com/probonopd/MiniDexed/actions/runs/7369725128/artifacts/1140935093)

diyelectromusic:
Latest, latest build: [MiniDexed_2024-01-03-d05f5f4](https://github.com/probonopd/MiniDexed/actions/runs/7403630422/artifacts/1146480803)

Banana71:
Better late than never I tested the “Latest, latest build” (MiniDexed_2024-01-03-d05f5f4).

After saving a new performance, The flag "[L]" jumps to the second-to-last performance and is therefore one position too far forward.
Only the lack of the performance number in the performance menus prevents me from using this version permanently.

diyelectromusic:
Thanks for that - I'll look at the jumping [L] thing... I hadn't noticed an issue, but I've not played around with saving/deleting much beyond the basics!

Where did you want the number - I expect that could be added, screen-size permitting! :)

Kevin

diyelectromusic:
Ah, you mean like this: https://github.com/probonopd/MiniDexed/pull/531

That approach won't do it anymore I'm afraid once this change takes effect (sorry @probonopd!) .  Shall I just add this in to my change too?

Kevin

diyelectromusic:
Ok, so there's a proposal that is close to what was mentioned here: https://github.com/probonopd/MiniDexed/discussions/414#discussioncomment-6698376

We can't use the bank name as they'll probably be too long, but we can use numbers.

I know this is now different to the voice menus, but then pretty much everything about performance handling is!

Kevin

Banana71:
I like lexonic's solution approach best. The top left would have the most space. There would be space problems right before the name. 
https://github.com/probonopd/MiniDexed/discussions/551

diyelectromusic:
That doesn't leave any room for the bank though, so I've done this:

IMG_7603

So the format is:
```
BNK:PRF [L]
<PerfName  >
```

But open to suggestions!

Kevin

Banana71:
Yes, that's right, I didn't even think about displaying the bank number.  Very well solved.

diyelectromusic:
Well I think this is probably there now, but I'd still like some reassurance it hasn't broken anything else.
Ideally anyone with a working system would try this to give it the widest range of configs possible!

Also, we will need to renumber the existing performances now :)

Kevin

probonopd:
Cool! Thanks a lot @diyelectromusic and @Banana71.

> Also, we will need to renumber the existing performances now :)

Let me know what I need to do and I'll happily do it 👍 

Build for [726470a](https://github.com/probonopd/MiniDexed/pull/581/commits/726470abfd8549b609c702551e5a58a6dc00e08d) for testing:
[MiniDexed_2024-01-08-8aae591](https://github.com/probonopd/MiniDexed/actions/runs/7451262015/artifacts/1154762874)

Banana71:
I tested the version "MiniDexed_2024-01-08-8aae591" and everything looks as expected. :+1:

>Also, we will need to renumber the existing performances now :)

What do you mean by that?

diyelectromusic:
The numbers in the performance filenames now define their place in the system, and consequently which MIDI patch number to use to select them.  But the system only allows 128 performances in a bank (so they can all be selectable via MIDI Program Change messages), but up to 128 banks,

If you install the existing set then anything over 128 will throw a warning to the console and not be available for use.

So we can either renumber them so they all have a number in the range 2 to 128 in a single bank (in the default bank voice 1 is the root performance.ini default performance); or split the default set up into banks.

As we don't have more than 128 initial performances (I think) I'd be tempted to say the initial set doesn't bother with banks so we just renumber them 2..128.  This then leaves banks for additional sets as required...

But to be honest, I don't mind either way :)

Kevin

Banana71:
Since the end of last year the number of Factory Performances has been 123, so there is still some space if someone wants to save their own. 😃 
The folder structure (banks) of the performances converted by Boban could still be adjusted. But we should first agree on which bank number the historical performances should start from.

PerformanceList

probonopd:
So, do we all think this is ready for being merged?

Banana71:
> So, do we all think this is ready for being merged?

No, ....
I'm having problems changing performance with an external midi controller.
When changing the sound, including the bank, there are often disturbing noises in the sound. With a bit of luck the sound will be fine again after the next program change.
In the sound example I switch back and forth between two performances and play a note, you will easily hear the error (after 15 seconds).
I filtered everything possible with MIDI-OX and came to the conclusion that the LSB command was causing the problem.


https://github.com/probonopd/MiniDexed/assets/104296205/a4cb7e34-dd64-4cc8-842d-65d1a26ed37e



MIDI-OX Programm Change msb lsb

diyelectromusic:
Hmm.  That's a fun one :)

And it definitely doesn't do this with the normal release?  This is definitely something as a consequence of this change?

Kevin

diyelectromusic:
Ok, yes I can reproduce it too.  Some observations so far:
* Doesn't seem to happen no matter how many times I change bank via the menu.
* Doesn't seem to happen when changing performances in the same bank if the bank hasn't been changed.
* Reselecting the same bank/performance repeatedly can cause it too.

There is some weird logic in there around switching performances - it isn't instantaneous, it is asynchronous, so I'm guessing some weird timing issue when the commands come over MIDI, but I'm not sure what. I wonder if the PC can come in mid bank select somehow - as when selecting a new bank, we basically have to do the power-up "list performances" thing to load in the new bank...

I'll add some debug code and take a look.

In the mean time, if you have any idea what parameters could be getting corrupted or out of sync to make that distortion occur,  then that might give me a clue where to look...?

Kevin

diyelectromusic:
Ok, give this one a go. The original performance loading was asynchronous with respect to the UI and MIDI, so I've done the same for bank selection too.  But its a little complicated...

Anyway, see if this throws up any new issues and more importantly, if it fixes the sound corruption.

I said this needs some thorough testing - and I wasn't joking :)

Give this new version a good going over too please!

Kevin

probonopd:
Build for testing:
[MiniDexed_2024-01-14-4999590](https://github.com/probonopd/MiniDexed/actions/runs/7521736444/artifacts/1168514544)

Banana71:
Good morning,
In this version "MiniDexed_2024-01-14-4999590" I can switch between performances without any problems. :smiley: :+1:

While testing I noticed another error. The error occurs in the previous version and in the last version.
When the MSB is changed, the Raspberry often crashes. The miniDexed freezes and you can only hear a humming noise from the audio output. All that's left is a cold start to bring the Raspberry back to life. :fearful: 
I am aware that the MSB is not relevant for the miniDexed, but it can always happen that someone sends an incorrect MIDI program change to the miniDexed.

The miniDexed has a maximum of 128 Performance banks then you could ignore the MSB messages that come via MIDI or simply use a "logically" AND 0 for MSB changes. 

MIDI-OX MSB Change

diyelectromusic:
Ah ok - looks like I'm missing some validation somewhere of the combined MSB/LSB... I'll take a look.  We have to assume we'll receive a MSB/LSB pair as that is what MIDI is saying should happen.

Thanks,
Kevin

diyelectromusic:
It was an assert that should have been an IF.  That should do it hopefully.

Kevin

Banana71:
Wonderful :banana:, my miniDexed runs stably in every situation. 
Thank you Kevin

diyelectromusic:
Yay!  I suggest we attempt to get a few other people to give it a go and see how it behaves before we merge.

Kevin

Banana71:
ModWheel issue

Problem example:
Performance "063 - Wilkinson" (TG1-TG8 is used) is played with the Mod Wheel turned up.
Then switch to "064 - Flute" (TG1&TG2), the tone has vibrato and the modwhell is turned down.
Then switch to Performance "067 - Tom's OB_DX" (TG1-TG6), the sound has vibrato again because TG3-TG6 still have the ModWheel position of Performance 063.
I hope you can understand what my problem is with the ModWheel.

When a performance is changed, the position of the ModWheel should be written to zero for all TG#s.

Peter

diyelectromusic:
Ok, I don't /think/ I've changed anything that would cause that ModWheel issue - would you mind seeing if that happens on the previous release too?  If so, then log it as a new issue.  If not, I'll see if I can work out why!! :)

Thanks,
Kevin

diyelectromusic:
By the way @probonopd I've pinged you a msg on Mastodon about something completely different :)
(Not sure I know of any other way to get in touch!?)

Kevin

diyelectromusic:
Right, couple more bug fixes: basically ensure performance bank select still works even if not enabled via MIDI with PCCH.

Build here: [MiniDexed_2024-01-23-f881b5f](https://github.com/probonopd/MiniDexed/actions/runs/7632808247/artifacts/1190213164)

I think this is probably it for now.  What do you think...?

Kevin

diyelectromusic:
OK, so I've done functionality tests on my Pi 3A+ for PCCH enabled/disabled; PerformanceSelectToLoad on or off; and various valid/invalid banks, gaps in numbering, etc, etc.

Also I've done very basic (boot/menu/MIDI) tests on RPi 400, RPi 4, Pi V1, Pi Zero 1 in USB gadget mode.

I think I'm probably ready unless anyone else finds anything odd happening?

Kevin

Banana71:
Runs like clockwork, Raspberry PI 4
Thank you @diyelectromusic Kevin 🦄

probonopd:
Excellent, so do we all agree that this is ready for being merged?
Thanks @diyelectromusic @Banana71 

> Mastodon

Looks like I really need to find my password again ,-)

diyelectromusic:
> > Mastodon
> Looks like I really need to find my password again ,-)

Or ping me an email at diyelectromusic at gmail :)

I'm good to go with this one and will just have to bug fix if more things turn up.

Kevin

probonopd:
Merged. @diyelectromusic please feel free to edit the wiki accordingly.

You can reach me at probono at puredarwin dot org, but the quickest way is usually via GitHub. Thanks!

diyelectromusic:
Some notes now added to the wiki, plus a bit of general tidying up of the performance section.

https://github.com/probonopd/MiniDexed/wiki/Files#performanceini

Kevin

Lonolo:
Wow - what a fantastic implementation!

Like clockwork! Using five Pi Zero2 modules MIDI-Thru-ed together.

Having the option to load the Performances from MIDI is awesome!

Much thanks for all the time put into organizing and building this!!!

diyelectromusic:
> Like clockwork! Using five Pi Zero2 modules MIDI-Thru-ed together.

Are these your eurorack modules?  I bet they sound amazing!! :)

Kevin

Lonolo:
Thanks again for this! They really do sound incredible! It's been great to use full Performances on 4 tracks and still have 8 DX7's to spare.
dXeus-racked

Regarding the new 'Program Change action' in the minidexed.ini parameter, couldn't the Performance>PCCH parameter override these as: 
  Off= 0
  1-16 = 1-16
  OMNI = >16

I don't mean to add any "feature creep" to this implementation, but having multiple minidexed modules, I'd have to pull each card and change that parameter every time I need to change the MIDI channel for choosing performances on that channel. That module then becomes dedicated until I pull the card and reset that minidiexed.ini prog-change-action.

```
# Program Change action:
#   0 = Program Change messages select voices on the channel associated with each TG.
#   1-16 = Program Change messages on this channel select performances.
#   >16 = Program Change messages on ANY channel select performances.
# NB: In performance mode, all Program Change messages on other channels are ignored.
PerformanceSelectChannel=0
```
I realize this is a specific user-case, but still wondering why the minidexed.ini parameters are necessary since the PCCH options are in the Menu. Would this work with other use cases or am I completely off base?

diyelectromusic:
> Regarding the new 'Program Change action' in the minidexed.ini parameter, couldn't the Performance>PCCH parameter override these as: Off= 0 1-16 = 1-16 OMNI = >16

It should do, although it isn't permanent - is that what you mean?  Basically on boot, the PCCH minidexed.ini value should set the default, but if you change PCCH via the menu then that should take effect instead until next power off/on...

Are you asking for a means of allowing the menu PCCH to be saved back to minidexed.ini?  We don't have any settings that are saved back like that atm apart from the contents of the performance.ini itself.

I keep thinking we ought to have some kind of system.ini, but we don't yet.

Kevin

Lonolo:
OK, I thought it was as you say, Including not being permanant - I get that too. 

Then is it possible that these Program change parameters are Not being sent to MIDI thru (MIDI thru is pre-Performance-change code)?

Here's what happens:
These modules are set up PCCH 1-5 from left to right. As I change the MIDI Channel and program on the MPC, the modules connected downstream of this MIDI thru chain get changed too:

https://github.com/probonopd/MiniDexed/assets/1200412/99495d50-7abb-4f93-8517-6defb7dd1f5b

diyelectromusic:
System level MIDI messages should be passed on via THRU, so that shouldn't be an issue.

I'm not entirely sure, if I'm honest, what you're describing...

But do you have an overlap between PCCH MIDI channels and channels in the performances?  Could that be an issue?  Try PCCH channels 2-6 maybe and leave channel 1 for playing?

@Banana71 are all the default performances configuring all TGs for channel 1?

> As I change the MIDI Channel and program on the MPC, the modules connected downstream of this MIDI thru chain get changed too:

I don't quite follow what you mean here - when you say change MIDI channel on the MPC do you mean changing what channel it is transmitting?  When you say "downstream get changed too" - what their MIDI channels?  I don't think that is possible via MIDI...?

Kevin

Lonolo:
What I'm trying to do is have the MPC sequencer load Performances for each MIDI-track.

I have five tracks of MIDI transmitting respectively on MIDI channels 1-5 from the MPC to Module 1 on the left, the other modules are chained from the previous modules' thru.

The PCCH channels are set 1-5 from left to right. In the video I'm transmitting from the MPC, changing the channel, then the program change parameter. 

I'm not comprehending when transmitting a prog change on let's say channel 3, that modules 4 & 5 change TG voices when their PCCH's are set 1-5 respectively & modules 1 & 2 don't change, only module 3 should change.

diyelectromusic:
> I'm not comprehending when transmitting a prog change on let's say channel 3, that modules 4 & 5 change TG voices when their PCCH's are set 1-5 respectively & modules 1 & 2 don't change, only module 3 should change.

Ha!  At the moment - you and me both!! :)

I've been reading through the code again and can't see a way that will do that atm... you might need to enable the MIDI debug log on say 4 and 5 (to keep it simple) and show me what is being received by each module (sorry, it will go to the display, so I know that will be a bit of a pain!) - but that might give a clue!

I'll try to cobble together some boards here and give it a go.

What are the MIDI channels of the performances you're using - just the defaults?  I think they're all set for channel 1... or have you changed the channels in the performances on 2-5 too?

Kevin

Lonolo:
I didn't change MIDI channels in any of the Performances, so all the modules should have Performances at their defaults which is typically channel 1 (unless it's a split type of performance ie 1 & 2).

I'll do a screenshot of the MIDI debug log - I'm curious what module 1 will say compared to 5...

...gotta go dig for a second mini-HDMI cable first!

diyelectromusic:
Ok - so this is what might be happening.

When a performance is selected, it does a sequence of internal "program changes" for each TG, but if the setting: `MIDIAutoVoiceDumpOnPC` is set to 1, which is the default, then there is a SysEx voice dump sent out to any active MIDI interfaces...

... so as all TGs are listening on MIDI 1, this gets sent out on MIDI channel 1 and received by any active TG.

So I don't think it is passing the performance selection via the THRU but the TG SysEx on MIDI channel 1 is being passed to all TGs.

So the performance isn't being changed, but the "live" DX voice is being set, via SysEx, to whatever voice is chosen as part of the selected performance...

Hopefully, setting `MIDIAutoVoiceDumpOnPC=0` will solve it.

I don't like this feature tbh - and think we should default to off, but it is legacy behaviour that someone coded in at one time and I'm reluctant to change the default atm...

But this seems like a strong reason to change it! :)

Kevin.

Lonolo:
BINGO!!!!

Works as expected now!

Stupid boolean, ate my entire day.

Thanks again for your time on this and agree it would make a great change.

Banana71:
Great that it works for you @Lonolo  now. 👍 

` MIDIAutoVoiceDumpOnPC=0` Basically, the first thing I do is set the parameter like this when I configure a new version of minidexed.

Anyone who works with more than one miniDexed at the same time or who wants to work with a MIDI channel other than 1 has the problem that the MIDI channel in the "default Performances" is set to "1" for the active TGs.

I haven't come up with a really good solution yet.

diyelectromusic:
> That was from turning the encoder one click on the first module's TG voice change!

There is definitely something odd going on there - F7 should be a valid end of SysEx marker - looks like its missed the F0 start marker - maybe it got spammed part way through another MIDI message and screwed up the parsing...

Anyway - from both these comments that settles it - that should be off by default :)

Kevin

Banana71:
A solution for multiple miniDexed:
```
#
# performance.ini
#

# TG#
#BankNumber#=0		# 0 .. 127
#VoiceNumber#=1		# 1 .. 32
#MIDIChannel#=1		# 1 .. 16, 0: off, 17: Midi Channel is adopted by PCCH ;  >17: omni mode
```

If the Midi channel in the performance is set to 17, the Midi channel is adopted with the value of PCCH.
The PCCH would therefore also be a kind of local MIDI channel

So I could set the MIDI channel to 17 in all default performances. The performances would then also be compatible with older systems.
If the PCCH is set to "off" you could leave the value at 17, which would again be OMNI. Difficult to explain.

Lonolo:
Thanks @Banana71 that makes sense,

So I'll write a script to modify all my Performance files changing MIDIChannel#=1 to MIDIChannel#=17, that would be ideal for a rack of MIDI selectable and controlled Performances played back.

Banana71:
Unfortunately I have no idea about C++
I'll try to describe in my own words (Basic) what function the code should have:



performanceconfig.cpp
```
                      READ PCCH: READ MIDIChannel# : IF PCCH=0 THEN GOTO  Assign MIDI channel  (ROW 111)
                      IF MIDIChannel#=17 THEN MIDIChannel#=PCCH ELSE GOTO  Assign MIDI channel (ROW 111)

111                Assign MIDI channel
112                PropertyName.Format ("MIDIChannel%u", nTG+1);
113                unsigned nMIDIChannel = m_Properties.GetNumber (PropertyName, 255);
114                if (nMIDIChannel == 0)
115                {
116                        m_nMIDIChannel[nTG] = CMIDIDevice::Disabled;
117                }
118                else if (nMIDIChannel <= CMIDIDevice::Channels)
119                {
120                        m_nMIDIChannel[nTG] = nMIDIChannel-1;
121                        bResult = true;
122                }
123                else
124                {
125                        m_nMIDIChannel[nTG] = CMIDIDevice::OmniMode;
126                        bResult = true;
127                }

```
I have no idea if this is the right place in the program but I hope you understand what I mean.

diyelectromusic:
> So I'll write a script to modify all my Performance files changing MIDIChannel#=1 to MIDIChannel#=17, that would be ideal for a rack of MIDI selectable and controlled Performances played back.

I'm afraid this doesn't exist - this is a proposal by @Banana71

I need to think about it properly and see how it might interact with everything else! :)

Kevin

Lonolo:
Doh!I see.

Then I second the motion!

Let me know how I can help and or test.

Banana71:
I plan to use a Raspberry Pi Zero 2 with miniDexed as a drum, bass and FX player. 16 TGs would be ideal for this, but the polyphony could be low. Kevin, can you compile a version in which I can freely set the parameters for the number of TGs and the polyphony? To save resources, the reverb could be deactivated.

PR #588, opened 2023-12-09
Patch Circle USB strings

### LINKED ISSUE #587: Rename "MIDI Gadget MIDI 1" to "MiniDexed MIDI 1"
Currently, REAPER (on WINE) shows MiniDexed as "MIDI Gadget MIDI 1".

We should rename "MIDI Gadget MIDI 1" to "MiniDexed MIDI 1" if possible to make it easier to find.

probonopd:
So that descriptive strings are used in dmesg and in DAWs.
Closes #587

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,8 +30,11 @@ if [ "${USB_DID}" ] ; then
 fi
 
 # Patch Circle so that descriptive strings are used in dmesg and in DAWs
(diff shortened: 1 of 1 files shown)


diyelectromusic:
No, this is brittle... if we want it changing, we need to get it changed in circle properly.  We really shouldn't be doing ad-hoc patching like this in my opinion...

Kevin

probonopd:
Solved properly by @diyelectromusic by subclassing in #587, hence closing this.

PR #614, opened 2024-02-05
Default MIDIAutoVoiceDumpOnPC to 0

### LINKED ISSUE #611: Default for MIDIAutoVoiceDumpOnPC should be 0 (OFF)
See discussion here: https://github.com/probonopd/MiniDexed/pull/581#issuecomment-1924560309

Basically any SysEx transmitted on voice change is also sent over any THRU link to a downstream MiniDexed/synth.  When selecting a performance, this will result in up to 8 seperate SysEx dumps all on MIDI channel 1 - as the default performances are configured all for MIDI channel 1.

This will make a mess of any downstream MiniDexed/DX7 as it will suddenly receive the sequence of SysEx dumps and thus change the "live" sounding voice to whatever the last active TG on the first MiniDexed selected as its voice in that performance...

Basically this setting makes no sense in a "performance"-selecting MiniDexed configuration.

Kevin

probonopd:
As suggested by @diyelectromusic, fixes #611

### DIFF
--- a/src/userinterface.cpp
+++ b/src/userinterface.cpp
@@ -390,10 +390,10 @@ void CUserInterface::UISetMIDIButtonChannel (unsigned uCh)
 		m_nMIDIButtonCh = CMIDIDevice::Disabled;
 		LOGNOTE("MIDI Button channel not set");
 	}
-	else if (uCh < CMIDIDevice::Channels)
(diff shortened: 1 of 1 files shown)


diyelectromusic:
I've not tested the build itself but on visual inspection, those changes look fine to me! :)

PR #638, opened 2024-04-15
Very early initial support for RPi 5 based on dev branch of circle.

diyelectromusic:
Initial support for RPI 5 based on develop branch of circle.

This is ***very experimental***.  Not to be merged at this time!

Known not to work:
* HDMI sound support (disabled in minidexed.cpp)
* USB Gadget support (disabled in kernel.cpp and usbminidexedmidigadget.h

I've seen it working with PCM5102 I2S sound, HD44780 (non I2C) display, HDMI console output, rotary encoder, USB keyboard.  It probably works with PWM sound output if you have the additional PWM filter hardware on GPIO pins 12/13...

Update: Looks like I2C SSD1306 is fine, so I'd guess I2C HD44780 ought to work too.
Update: Serial MIDI now working fine.

Everything else: I've no idea :)

Note: to get this running needs the latest boot files from Raspberry Pi (see the circle documentation), an updated config.txt (part of this commit) and the new kernel_2712.img file built when RPI=5.

I've not done anything with the automated build process, but if someone wants to fight with the build.yml file, feel free!  

Otherwise to build you'll need to ensure you are building the develop branches of circle-stdlib and circle itself (see the submod.sh file).

Apart from that, well, it seems to work :)

By the way, I don't think we say this enough... Rene is amazing for his circle environment that is doing all the heavy lifting here :)

Also, on that point, now we've upgraded to a more recent circle, we need to be testing the builds for RPi 1,2,3,4 too before accepting this change!

Kevin

IMG_7785

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,12 @@ jobs:
            wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-arm-none-eabi.tar.xz
            tar xf *-arm-none-eabi.tar.xz
            mkdir -p kernels
+    - name: Build for Raspberry Pi 5
+      run: |
+           set -ex
+           export PATH=$(readlink -f ./gcc-*aarch64-none*/bin/):$PATH
+           RPI=5 bash -ex build.sh
+           cp ./src/kernel*.img ./kernels/
     - name: Build for Raspberry Pi 4
       run: |
            set -ex
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ Video about this project by [Floyd Steinberg](https://www.youtube.com/watch?v=Z3
 
 ## System Requirements
 
-- Raspberry Pi 1, 2, 3, 4, or 400 (Zero and Zero 2 can be used but need HDMI or a supported i2s DAC for audio out). On Raspberry Pi 1 and on Raspberry Pi Zero there will be severely limited functionality (only one tone generator instead of 8)
+- Raspberry Pi 1, 2, 3, 4, or 400. Raspberry Pi Zero and Zero 2 can be used but need HDMI or a supported i2s DAC for audio out. On Raspberry Pi 1 and on Raspberry Pi Zero there will be severely limited functionality (only one tone generator instead of 8)
+- Raspberry Pi 5 can be used but currently support is experimental; HDMI sound and USB Gadget mode are not available yet 
 - A [PCM5102A or PCM5122 based DAC](https://github.com/probonopd/MiniDexed/wiki/Hardware#i2s-dac), HDMI display or [audio extractor](https://github.com/probonopd/MiniDexed/wiki/Hardware#hdmi-to-audio) for good sound quality. If you don't have this, you can use the headphone jack on the Raspberry Pi but on anything but the Raspberry 4 the sound quality will be seriously limited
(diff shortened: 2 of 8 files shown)


rsta2:
Thank you, Kevin! The serial device defaults to the UART JST socket on the RPi 5. You have to append an additional parameter `0` [here](https://github.com/probonopd/MiniDexed/blob/349100d65ad59e73b091031ea011b0f73242890b/src/serialmididevice.cpp#L35) to direct it to GPIO14/15. This will work with all RPi models.

diyelectromusic:
> Thank you, Kevin! The serial device defaults to the UART JST socket on the RPi 5. You have to append an additional parameter `0` [here](https://github.com/probonopd/MiniDexed/blob/349100d65ad59e73b091031ea011b0f73242890b/src/serialmididevice.cpp#L35) to direct it to GPIO14/15. This will work with all RPi models.

Ah great - I'll take a look later this evening :)  I've not really dug into it at all yet - I only built the thing!!

Kevin

diyelectromusic:
If anyone is able to test this for side-effects of the circle upgrade on a RPi 1,2,3,4 then the build is here:
[MiniDexed_2024-04-16-96294ed](https://github.com/probonopd/MiniDexed/actions/runs/8704863700/artifacts/1417933621)

Note: This does NOT include the RPi 5 version yet.  That requires an update to the automated build setup on Github.

Kevin

diyelectromusic:
Ok, so this build works on the following for me:
* RPi V1
* RPi V1 Zero (Gadget Mode)
* RPi V3A+ (Gadget Mode)
* RPi 4

Showing mixture of HD77480 and I2C SSD1306 screens, rotary encoder, buttons, PCM5102 I2S DAC, and serial and USB MIDI.

So I'd say the basic build seems pretty ok to me at the moment.

Kevin

probonopd:
> Note: This does NOT include the RPi 5 version yet. That requires an update to the automated build setup on Github.

Let's see if we can change that :)

Build for testing: [MiniDexed_2024-04-16-a031f41](https://github.com/probonopd/MiniDexed/actions/runs/8709482473/artifacts/1419079001)

@rsta2: https://github.com/rsta2/circle?tab=readme-ov-file#building should probably be updated.

rsta2:
> @rsta2: https://github.com/rsta2/circle?tab=readme-ov-file#building should 
probably be updated.

@probonopd You have to look into the AArch64 section. Historically the 
Building section is for AArch32 builds only, and the RPi 5 does not support 
AArch32. I think, this is expressed in the comments at the beginning of the 
Building section.

Banana71:
With my Raspberry Pi 4 (SSD1306, GY-PCM5102, classic MIDI Input) everything seems to be fine. 👍

diyelectromusic:
Latest build now tested on:
* RPi V1 (MIDI + USB)
* RPi V1 Zero (MIDI + USB Gadget)
* RPI V3A+ (MIDI + USB Gadget)
* RPi 4 (MIDI + USB)
* RPi 400 (MIDI + PC Keyboard)
* RPi 5 (MIDI + USB)

With various combinations of HD77480 or ssd1306 display, rotary encoder, PCM5102 I2S DAC and GPIO buttons.

I think we can probably merge this in with a note on the main page that RPi 5 support is present but experimental and doesn't support HDMI sound or USB Gadget mode.  What does everyone think?

Then I might revisit the polyphony/TG situation - I see no reason it shouldn't be 32 note/16 TG by default on a Pi 5... anyone?

Kevin

PR #643, opened 2024-04-20
Update README.md

diyelectromusic:
Added recommendation to use an older or cheaper Pi (not a Pi 5) for a first build.


probonopd:
Thanks @diyelectromusic. Merged with slight wording changes.

Issue #648, opened 2024-04-25
Warnings about truncated strings related to VoiceName


diyelectromusic:
The build is producing these warnings:

```
minidexed.cpp: In member function 'void CMiniDexed::SetVoiceName(std::string, unsigned int)':
minidexed.cpp:1771:9: warning: 'char* strncpy(char*, const char*, size_t)' specified bound 10 equals destination size [-Wstringop-truncation]
 1771 |  strncpy(Name, VoiceName.c_str(),10);
      |  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
../Synth_Dexed/src/dexed.cpp: In member function 'void Dexed::setName(char*)':
../Synth_Dexed/src/dexed.cpp:1699:10: warning: 'char* strncpy(char*, const char*, size_t)' output may be truncated copying 10 bytes from a string of length 155 [-Wstringop-truncation]
 1699 |   strncpy(name, (char*)&data[DEXED_VOICE_OFFSET + DEXED_NAME], 10);
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Should probably have a look into that at some point - I think it's fine, but it would be good to get rid of the warnings if so!

Kevin

probonopd:
Should hopefully be fixed no; comment here if it is not. Thanks!

Issue #501, opened 2023-05-27
Dexed:setName/getName weirdness


diyelectromusic:
In MiniDexed.cpp the functions GetVoiceName (https://github.com/probonopd/MiniDexed/blob/main/src/minidexed.cpp#L930) and SetVoiceName (https://github.com/probonopd/MiniDexed/blob/main/src/minidexed.cpp#L1566) use Dexed::setName/getName (https://codeberg.org/dcoredump/Synth_Dexed/src/branch/master/src/dexed.cpp#L1699), but I'm wondering if the sense of these functions in the Dexed library might be incorrect?

GetVoiceName uses setName to read the name from Dexed, and SetVoiceName uses getName to write it.

In Dexed the code is:
```
void Dexed::setName(char* name)
{
  strncpy(name, (char*)&data[DEXED_VOICE_OFFSET + DEXED_NAME], 10);
 }
  
void Dexed::getName(char* buffer)
{
   strncpy((char*)&data[DEXED_VOICE_OFFSET + DEXED_NAME], buffer, 10);
   buffer[10] = '\0';
}
```

So as strncpy takes parameters in the order (dest, src, len) these do seem to be doing the opposite of what their names imply, which means our code is correct.

However, there are some oddities:
* getName sets the last character of the provided buffer to \0, which implies it was meant to be filling in the buffer not reading from it.
* setName does nothing with the provided buffer but the VOICE DATA name is not a NULL terminated string, it is simply 10 characters.
* In SetVoiceName a buffer of only 10 characters is provided to getName, which implies that the final "buffer[10]='\0'" will cause a buffer overrun...
* In GetVoiceName we use a buffer of 11 characters and preset them all to 0 which means the non-null terminated string is ok.

So in short, I think the parameters to the strncpy calls in get/setName are the wrong way round, that we have a bug in our SetVoiceName and that our functions ought to be properly paired with the get/set in Dexed....

Thoughts?
Kevin

BobanSpasic:
I do not know the data structures in Dexed, but it does not seem wrong to me.
SetName puts the string 'name' into the DEXED_VOICE_OFFSET structure at the position DEXED_NAME, and just the 10 chars, ignoring the rest if longer.
GetName gets 10 chars from DEXED_VOICE_OFFSET and put them into 'buffer' (initialized somewhere else, before the function call), again just in length from 10 chars. Add a null char (string terminator) to make it a full qualified string.

probonopd:
[...]

PR #649, opened 2024-04-25
Attempt to fix #646

### LINKED ISSUE #646: Option to have reverb (or other future effects) in mono mode
I use Minidexed as 2 synth engines TG1 & TG2" panned hard left and right. The output channels (left & right) go to different destinations in my music setup. If I use reverb on either of the engines, the reverberated sound will start leaking out from the other channel as well, rendering it useless for my use case.

Therefore I'd propose having the option to set reverb into mono-mode (i.e. no stereo widening) to keep the stereo image clean.

probonopd:
https://github.com/probonopd/MiniDexed/issues/648

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1763,20 +1763,14 @@ bool CMiniDexed::IsValidPerformanceBank(unsigned nBankID)
 	return m_PerformanceConfig.IsValidPerformanceBank(nBankID);
 }
 
-void CMiniDexed::SetVoiceName(std::string VoiceName, unsigned nTG)
+void CMiniDexed::SetVoiceName (std::string VoiceName, unsigned nTG)
 {
-	assert(nTG < CConfig::ToneGenerators);
-	assert(m_pTG[nTG]);
-
-	// Define a buffer with enough space for the string and null terminator
-	char Name[VoiceName.size() + 1];
-
-	// Copy the string and ensure null termination
-	strncpy(Name, VoiceName.c_str(), VoiceName.size());
-	Name[VoiceName.size()] = '\0';
-
-	// Pass the null-terminated string to getName
-	m_pTG[nTG]->getName(Name);
+	assert (nTG < CConfig::ToneGenerators);
+	assert (m_pTG[nTG]);
+	char Name[11];
+	strncpy(Name, VoiceName.c_str(),10);
+	Name[10] = '\0';
+	m_pTG[nTG]->getName (Name);
 }
 
 bool CMiniDexed::DeletePerformance(unsigned nID)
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks @diyelectromusic. It looks like my PR fixes one instance of this issue (at least the build log doesn't show one of the errors anymore); the other would need to be fixed by a similar change in Synth_Dexed.
Please do review my change and, if you have a chance, test the build. Thanks!

diyelectromusic:
I think the issue is that the fixed length 10 might be correct - it is setting the voice string within the fixed DX voice structure I think...

... but I'll need to have a proper look as it is also tied up with the mess that is get/setName in Dexed (more here: https://github.com/probonopd/MiniDexed/issues/501)

Kevin

diyelectromusic:
Ok, yes, looking again, we do need the fixed 10 in the strncpy as that is the size of the voice name in the voice structure for the DX7.

So I think the fix will be the following:
```
void CMiniDexed::SetVoiceName (std::string VoiceName, unsigned nTG)
{
	assert (nTG < CConfig::ToneGenerators);
	assert (m_pTG[nTG]);
	char Name[11];
	strncpy(Name, VoiceName.c_str(),10);
	Name[10] = '\0';
	m_pTG[nTG]->getName (Name);
}
```
A similar thing is probably required in Synth_Dexed/dexed.cpp but as I say that is mixed up with the problems described in #501  so that would be a breaking change in the API to fix... but it really ought to be fixed!?

It does address my comment in #501 about SetVoiceName only have a buffer of 10 though - that is wrong and does need to be 11 as shown above.

Kevin

probonopd:
Why not use `VoiceName.size()` as suggested in my PR? That way, we don't need to hardcode any sizes. I try to never hardcode sizes, as it is a frequent source of bugs.

diyelectromusic:
> Why not use `VoiceName.size()` as suggested in my PR? That way, we don't need to hardcode any sizes. I try to never hardcode sizes, as it is a frequent source of bugs.

Because this function is about truncating whatever voice name is passed in (which could be any size) to the explicit 10 characters defined in the DX7 SysEx voice structure, so as I say, the 10 is actually correct in this case, just not really accounted for properly in how its used.

We could make the argument that GetName should do all that - and yes, I agree it ought to - but GetName/SetName is a mess as per https://github.com/probonopd/MiniDexed/issues/501 and can't be fixed without breaking the API so I doubt that will happen.

So best just to do the robust thing here.  In an ideal world I agree we wouldn't have hardcoded sizes like this, but the DX7 voice structure has been fixed for over 40 years, so I don't think we need to worry about that in this case :)

Kevin

probonopd:
> the DX7 voice structure has been fixed for over 40 years, so I don't think we need to worry about that in this case :)

Good point 👍

PR #690, opened 2024-07-20
Another attempt to support configuration TGs and polyphony

diyelectromusic:
Supports TG configurations as follows:
Pi 1 = 1 TG
Pi 2,3 = 8 TG
Pi 4,5 = 8 or 16 TG (default 8)

and polyphony as follows:
Pi 1 = 8
Pi 2,3 = 16
Pi 4 = 24 (default 16)
Pi 5 = 32 (default 16)

Relevant minidexed.ini parameters are:
```
ToneGenerators=
Polyphony=
```

Menus seem to working ok for different TG configurations.  Note - the main menu (TG/effects/performance) now wraps around.

If you fancy having a look, the following all has to be tested:
* Are the configured TGs actually working ok over MIDI (up to 16 TGs with various MIDI channel configs/OMNI)?
* Home/back/select on various menus
* Pgm Up/Down buttons
* TG Up/Down buttons
* Is the correct polyphony configured?
* Reducing polyphony - does it really do it?
* Pi 1, 2, 3 configurations for backwards compatible/default behaviours
* Home/back/select on a Pi V1 (with only one TG)
* Performance loading/saving:
-- 8TG performance into 16TG system
-- 16TG performances into 8TG system
-- loading/saving 8TG performances with 8TG system
-- loading/saving 16 TG performances with 16TG system

And probably a whole pile of other things too.  Basically if you fancy giving this a go with your standard configuration, please do!

I'll start on with some this myself too, so expect updates over the next week or so as I find things wrong, but please report any errors, asserts from the console, weirdness, etc here with details of your configuration and I'll take a look!

Kevin

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -45,7 +45,7 @@ void CConfig::Load (void)
 	{
 		m_nToneGenerators = DefToneGenerators;
 	}
-	if (m_nPolyphony >= MaxNotes)
+	if (m_nPolyphony > MaxNotes)
 	{
 		m_nPolyphony = DefaultNotes;
 	}
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -36,6 +36,20 @@ void CConfig::Load (void)
 {
 	m_Properties.Load ();
 	
+	// Number of Tone Generators and Polyphony
+	m_nToneGenerators = m_Properties.GetNumber ("ToneGenerators", DefToneGenerators);
+	m_nPolyphony = m_Properties.GetNumber ("Polyphony", DefaultNotes);
+	// At present there are only two options for tone generators: min or max
+	// and for the Pi 1,2,3 these are the same anyway.
+	if ((m_nToneGenerators != MinToneGenerators) && (m_nToneGenerators != AllToneGenerators))
+	{
+		m_nToneGenerators = DefToneGenerators;
+	}
+	if (m_nPolyphony >= MaxNotes)
+	{
+		m_nPolyphony = DefaultNotes;
+	}
+	
 	m_bUSBGadget = m_Properties.GetNumber ("USBGadget", 0) != 0;
 	m_nUSBGadgetPin = m_Properties.GetNumber ("USBGadgetPin", 0); // Default OFF
 	SetUSBGadgetMode(m_bUSBGadget); // Might get overriden later by USBGadgetPin state
@@ -175,6 +189,48 @@ void CConfig::Load (void)
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
 	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 1) != 0;
 	m_bPerformanceSelectChannel = m_Properties.GetNumber ("PerformanceSelectChannel", 0);
+}
+
+unsigned CConfig::GetToneGenerators (void) const
+{
+	return m_nToneGenerators;
+}
+
+unsigned CConfig::GetPolyphony (void) const
+{
+	return m_nPolyphony;
+}
+
+unsigned CConfig::GetTGsCore1 (void) const
+{
+#ifndef ARM_ALLOW_MULTI_CORE
+	return 0;
+#else
+	if (m_nToneGenerators > MinToneGenerators)
+	{
+		return TGsCore1 + TGsCore1Opt;
+	}
+	else
+	{
+		return TGsCore1;
+	}
+#endif
+}
+
+unsigned CConfig::GetTGsCore23 (void) const
+{
+#ifndef ARM_ALLOW_MULTI_CORE
+	return 0;
+#else
+	if (m_nToneGenerators > MinToneGenerators)
+	{
+		return TGsCore23 + TGsCore23Opt;
+	}
+	else
+	{
+		return TGsCore23;
+	}
+#endif
 }
 
 bool CConfig::GetUSBGadget (void) const
--- a/src/config.h
+++ b/src/config.h
@@ -70,11 +70,11 @@ public:
 	static const unsigned MaxNotes = 8;
 	static const unsigned DefaultNotes = 8;
 #elif RASPPI == 4
-	static const unsigned MaxNotes = 24;
-	static const unsigned DefaultNotes = 16;
+	static const unsigned MaxNotes = 32;
+	static const unsigned DefaultNotes = 24;
 #elif RASPPI == 5
 	static const unsigned MaxNotes = 32;
-	static const unsigned DefaultNotes = 16;
+	static const unsigned DefaultNotes = 32;
 #else
 	static const unsigned MaxNotes = 16;
 	static const unsigned DefaultNotes = 16;
--- a/src/config.h
+++ b/src/config.h
@@ -35,18 +35,49 @@
 class CConfig		// Configuration for MiniDexed
 {
 public:
+// Set maximum, minimum and default numbers of tone generators, depending on Pi version.
+// Actual number in can be changed via config settings for some Pis.
 #ifndef ARM_ALLOW_MULTI_CORE
-	static const unsigned ToneGenerators = 1;
+	// Pi V1 or Zero (single core)
+	static const unsigned MinToneGenerators = 1;
+	static const unsigned AllToneGenerators = 1;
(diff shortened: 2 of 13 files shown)


Banana71:
Hello Kevin, thank you very much for your efforts.

I tested my miniDexed with a Raspberry 4 and found out the following: 
- The menu navigation works perfectly (TG/effects/performance).
- I can set the TG configuration to 8 or 16. Intermediate values ​​or <8 have no effect (you didn't describe that either).
- I can set the maximum polyphony however I want and it stays at 16 voices.

Notes:
Since the March version (https://github.com/probonopd/MiniDexed/pull/623) , I have been playing my RPi 4 with 32 voices and 8 TG. I have never noticed any problems. Why can't I set this setting in the new version? (8TGs with 32 voices on a RPi 4).

If you work with 16 TGs, I would divide the miniDexed into a lower and an upper area. The following menus would then be conceivable:
Load lower performance (TG1-TG8)
Load upper performance (TG9-TG16)
The performances could remain as they are and you could load into both areas.

diyelectromusic:
Brilliant - thanks for that - yes, I've just noticed the polyphony too! I'll take a look.

I think the idea was really to go with safe defaults re TGs and polyphony, to ensure no issues - there was some issue with different engines?  Anyway, I set the max for a Pi 4 to 24, but we can up it to 32 if that works ok.

A dual-performance idea sounds really cool!  But I won't think about that as part of this change, I think for now lets get the basics assured and then we can move on from there.

Thanks for giving it a go!  I'm slowly working through a few different configurations myself this morning :)

Kevin

diyelectromusic:
Ok, so limiting polyphony is working, just not in a very obvious way!

This is a printout of an instrumented Synth_Dexed showing which note "slot" it is using as I press 6 notes, then release them in the reverse order, when the system is configured for 4 note polyphony.  Notice how it keeps reusing "slot 1" which isn't the first note pressed...  Then when the other notes that were in "slot 1" are released nothing happens as the slot was reused and already released...

This is just the algorithm within Synth_Dexed so not related to anything we do.

Note there are 16 printouts for max_notes = 4 as each TG is initialised for 4-note polyphony and I have 16 configured.

Kevin

IMG_7981

diyelectromusic:
Ah this is fun.  When loading a 8 TG performance into a 16TG setup, all unused TGs get set to MIDI OMNI... I'll have to change that otherwise every performance is a little "Brass 1" heavy :)

diyelectromusic:
Ok, so I've tested the build on the following:
* Pi 1 - checked home/back/TGs/polyphony
* Pi Zero in USB Gadget mode
* Pi 3 - configured for more TGs/polyphony - defaults to right values
* Pi 4 - 8 TGs, 16 TGs (with different MIDI channels), loading performances, etc.
* Pi 5 - Quad DAC mode for 16 TGs correctly defaults back to non-Quad DAC mode; 16TGs seem fine.

Haven't checked higher polyphony, but have checked configuring for lower.

Latest build fixes performance issue with MIDI channels and sets max polyphony to 32 on Pi 4 and 5 with the defaults being a "safe" 24 and 32 respectively.

Do please check with your existing configurations - I want to ensure all the standard options haven't messed up! :)

I still need to check MIDI TG or Program buttons Up/Down, so if anyone is already using those and gets there first - let me know!

Note: I didn't do wrap around with the TG buttons - just with the menu.

Kevin

probonopd:
As always, thanks a ton for your great contribution! Allow me a question though:

> Note: I didn't do wrap around with the TG buttons - just with the menu.

Did you do this for a special reason or just because you didn't come around to doing it yet? Personally I am a _huge_ believer in consistency. I think things should work in the same way across an user interface. It already drives me nuts that the performance selection works differently from the voice selection (having to click vs. not having to click). So I am curious to understand whether you made that decision intentionally or just due to not having it implemented yet.

Banana71:
> Build for testing: [MiniDexed_2024-07-21-9231493](https://github.com/probonopd/MiniDexed/suites/26245669085/artifacts/1723145176) Use at your own risk.

with a RPi 4 I get the following polyphony:
`polyphony=16`   plays 16 voices
`polyphony=24`   plays 24 voices
`polyphony=32`   plays 24 voices :cry:

diyelectromusic:
> As always, thanks a ton for your great contribution! Allow me a question though:
> 
> > Note: I didn't do wrap around with the TG buttons - just with the menu.
> 
> Did you do this for a special reason or just because you didn't come around to doing it yet?

I agree consistency is important, but the TG up/down is a very different interaction to the menus, so I thought that it was more important, especially if a remote button (i.e. MIDI TG buttons) is being used that one can't "get lost" working out which TG is selected.  I thought that with wrap around that was a real danger.  This way, keep pressing "down" and you'll eventually get to the first TG, and same with "up" to the last, so it would be easy to re-orient yourself if selecting them with no UI nearby to see what is going on.

That was my thinking anyway - happy to consider alternatives!

(the logic in the code is very different for the two cases anyway, so I'm not making things more complicated by doing things differently)

Kevin

diyelectromusic:
> > Build for testing: [MiniDexed_2024-07-21-9231493](https://github.com/probonopd/MiniDexed/suites/26245669085/artifacts/1723145176) Use at your own risk.
> 
> with a RPi 4 I get the following polyphony: `polyphony=16` plays 16 voices `polyphony=24` plays 24 voices `polyphony=32` plays 24 voices 😢

Hmm.  No idea why that is.  Unless this is the first version linked in this thread - the first build limited the Pi 4 to 24 note polyphony, but the latest build should allow 32 I think?

Kevin

diyelectromusic:
> Hmm. No idea why that is.

Ok, actually I might have an idea... looks like this ">=" should be a ">"  :)
https://github.com/probonopd/MiniDexed/pull/690/files#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R48

Kevin

Banana71:
Now I can play the RPi 4 with 32 voices polyphony.
Thank you Kevin

Onosendaifr:
Just to be sure, is it polyphony per TG or for the whole ?

Banana71:
> Just to be sure, is it polyphony per TG or for the whole ?
It means the polyphony for each TG 🤩

Onosendaifr:
> If you work with 16 TGs, I would divide the miniDexed into a lower and an upper area. The following menus would then be conceivable:
> Load lower performance (TG1-TG8)
> Load upper performance (TG9-TG16)
> The performances could remain as they are and you could load into both areas.

I love your idea ! It's perfect for having 8x TG in the lower part, and load a complete drum kit section in the upper part, i really hope this will be push later :)

diyelectromusic:
I'm afraid that is probably a fair bit of significant work and we'd need to really pin down the behaviours we want in the various different configurations (8TG vs 16TG vs single vs split performances, etc).

But I'll ponder while we're giving this one a shakedown first!

Kevin

probonopd:
Would be interesting to know how the Montage and similar instruments behave in this regard.

Banana71:
What comes to mind in this context

Performance handling: https://github.com/probonopd/MiniDexed/pull/581#issuecomment-1924735513

diyelectromusic:
Has anyone else been able to check this firmware on their own configurations?  I'd like to know we haven't broken backwards compatibility and "normal" configurations with this change.

Kevin

Onosendaifr:
> Note: I didn't do wrap around with the TG buttons - just with the menu.

Build tested on my Rasp 3 and everything is working fine ! About this tho, you should add wrap to TG because as soon as we can see theTG number on the screen, we are never lost.

miotislucifugis:
Im using a pi 3 so no big changes in the # of TGs or poly, but everything seems to be working fine for me.

diyelectromusic:
> Im using a pi 3 so no big changes in the # of TGs or poly, but everything seems to be working fine for me.

Great!  Thanks for letting us know.

Kevin

diyelectromusic:
I think this is probably at a point where it can be merged in, then I can start thinking about what might be worth doing next - I'm trying not to have several things on the go at once and I really don't want to be continually merging against an updating main branch!

(This is where having a more quickly merged dev branch would make life quite a bit easier and mean things could be added to it more quickly ;))

Kevin

probonopd:
Thank you very much @diyelectromusic. As always, please don't forget to update the wiki accordingly.

PR #708, opened 2024-08-23
Add USB audio support

probonopd:
Thanks @rsta2

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -23,6 +23,7 @@
 #include <circle/sound/pwmsoundbasedevice.h>
 #include <circle/sound/i2ssoundbasedevice.h>
 #include <circle/sound/hdmisoundbasedevice.h>
+#include <circle/sound/usbsoundbasedevice.h>
 #include <circle/gpiopin.h>
 #include <string.h>
 #include <stdio.h>
@@ -200,6 +201,16 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 		// The channels are swapped by default in the HDMI sound driver.
 		// TODO: Remove this line, when this has been fixed in the driver.
 		m_bChannelsSwapped = !m_bChannelsSwapped;
+#endif
+	}
+	else if (strcmp (pDeviceName, "usb") == 0)
+	{
+#if RASPPI<=3
+		LOGNOTE ("USB mode NOT supported on RPI 1-3.");
+#else
+		LOGNOTE ("USB mode");
+
+		m_pSoundDevice = new CUSBSoundBaseDevice (pConfig->GetSampleRate ());
 #endif
 	}
 	else


rsta2:
@probonopd You are welcome. The RPi 4 or 5 must run in host mode, not gadget mode, and the USB audio interface must be connected, when the system is started (no hot-plugging here).

probonopd:
Thanks for the clarification @rsta, of course I meant "must run in host mode, not gadget mode". Corrected above.

Koshdukai:
Well, tried the [MiniDexed_2024-08-23-9ade4be](https://github.com/probonopd/MiniDexed/suites/27531746246/artifacts/1847937561) in a RPi400 and it was able to connect to an Arturia MiniFuse1 but the audio comes out distorted.
Tried with no soundopt=24, with soundopt=16, soundopt=24 and soundopt=32 on the cmdline.txt but the style of digital distortion kept being similar.
Also tried SampleRate=48000 and SampleRate=44100 on the minidexed.ini with no change either.

So I guess, the SoundDevice=usb does work but there's something still needing to be tweaked, most likely on the settings side but all I know I could try would be audio bit depth and audio rate ...ah, maybe ChunkSize? gonna try that one now

probonopd:
Thanks for testing @Koshdukai. Any chance you could record and upload this for us to listen to?

Koshdukai:
I did try to attach a zip with a .wav of the audio but the attachment didn't "stick"
Trying again here
[2-Audio 0001 [2024-08-28 200218].zip](https://github.com/user-attachments/files/16788822/2-Audio.0001.2024-08-28.200218.zip)

ah, it worked this time :)

rsta2:
I will try this MiniDexed test version by myself here soon.

Koshdukai:
btw, I initially tried it with an old Propellerhead Balance interface that I'm already using with no issues with that same RPi400 running Zynthian, but this MiniDexed build didn't look like it was able to use it.
That's why I tried with the MiniFuse1 (yet another small audio interface that I'm trying to put to some use instead of having it stored, doing nothing).

paolofalcoruegg:
Unfortunately I wasn't successful testing with my [Alesis Core1](https://podcastage.com/rev/core1) interface. I set `soundopt=24` as well as `SoundDevice=usb`. I used a RPi4 that was working before with the current Production build of MiniDexed. Are there any logs that I could provide?

rsta2:
I have to confirm, that this test version produces noise via USB audio streaming, but not a clear sound. I tested with an ASUS XONAR U5 interface with 48000 Sa/s and 16-bits sample size. Because after a lot of changes to MiniDexed in the meantime I cannot quickly say, what the reason is or may be.

I can only say, that USB audio works with my MiniSynth Pi, but MiniSynth Pi uses the (older) [alternate programming interface](https://circle-rpi.readthedocs.io/en/47.0/devices/audio-devices.html#alternate-interface), provided by the class `CSoundBaseDevice`, while MiniDexed uses the standard interface. I tested USB audio streaming successfully with the standard interface with Circle sample/34-sounddevices and the test/sound-controller, but these are not real synth apps and do not use multiple cores.

Taking into account that USB audio devices do not work with a fixed chunk size, this may affect the behavior of the class `CSoundBaseDevice`. The number of transferred audio samples changes with each USB transfer.

Now I have to port MiniSynth Pi to the standard interface of `CSoundBaseDevice`, to see, if this works or not. This will take a few days. Problems with specific interfaces could be sorted out afterwards.

rsta2:
I got MiniDexed working with the initial USB build with these USB audio streaming devices:

* ASUS XONAR U5 (16 or 24 bits sample size)
* Unitek Y-247A (16 bits sample size)
* C-Media PD552 (16 bits sample size)

The ChunkSize parameter is very critical. I used these parameters in *minidexed.ini*:

```
SoundDevice=usb
SampleRate=48000
ChunkSize=384
```

A chunk size of 256 did not work, neither did 1024. I think, it should be a multiple of `SampleRate / 1000` and greater than 192.

It's important to add the following parameter to the file cmdline.txt on the SD card, if your USB audio streaming device does support a sample size of 24 bits only:

```
soundopt=24
```

The default is `soundopt=16`. `soundopt=32` is not supported.

There are many different USB audio streaming devices out there and it's likely, that some of them will not work. To be able to debug the reason, if it does not work, the log messages starting with `uaudio1` and `sndusb`, are of interest.

probonopd:
Thank you very much for looking into this @rsta2.

Koshdukai:
ah so it was ChunkSize related, as I was kinda suspecting/hoping :)
Will try with the "SampleRate / 1000 and greater than 192" rule, which basically should be these:
SampleRate=48000
ChunkSize=384

In 44.1KHz case, I'm guessing 441 should work but I'm wondering how to deal with any of the non-integer multiples below it.

Anyway, time to test this latest build :)
Thanks @rsta2 and @probonopd !

Koshdukai:
Got it working on the Arturia MiniFuse1 with both soundopt=16 or soundopt=24 (it supports both formats, so that checks out).
Tried with SampleRate=48000 and ChunkSize=384 and the audio is near perfect, except for a cyclic "click" (like a DC jump, probably 1 or a couple audio samples/frames wide)
Also tried with with ChunkSize=768, but I still get it:
image
So no matter if soundopt is 16 or 24 and obeying the SampleRate/1000*n>192 rule, that cyclic click shows up.
[MiniDexed_2024-08-31-9b93d18 Arturia MiniFuse1 soundopt=16 SampleRate=48000 ChunkSize=768.zip](https://github.com/user-attachments/files/16824547/MiniDexed_2024-08-31-9b93d18.Arturia.MiniFuse1.soundopt.16.SampleRate.48000.ChunkSize.768.zip)

I'm now going to check if minimum changes to the ChunkSize affect that cycle, hopefully finding a value that makes it disapear... if still ChunkSize related.

Koshdukai:
after a lot of testing with various ChunkSizes, I end up returning to the 384 value

**SoundDevice=usb
SampleRate=48000**
#ChunkSize=192
#ChunkSize=240
#ChunkSize=288
#ChunkSize=336
#ChunkSize=383 -> fuzzy
#125b/s:ChunkSize=384 -> click
**ChunkSize=384**
#ChunkSize=385 -> click
#ChunkSize=386 -> garbled
#ChunkSize=432
#100b/s:ChunkSize=480
#96b/s:ChunkSize=500
#80b/s:ChunkSize=600
#75b/s:ChunkSize=640
#64b/s:ChunkSize=750
#ChunkSize=768 -> more clicks
#60b/s:ChunkSize=800
#50b/s:ChunkSize=960
#48b/s:ChunkSize=1000
#40b/s:ChunkSize=1200
#32b/s:ChunkSize=1500
#ChunkSize=1536
#30b/s:ChunkSize=1600
#ChunkSize=3072
#

even trying some of the usually suggested ones for 48KHz on Windows:
image

rsta2:
@Koshdukai Thanks for testing! Please try `soundopt=24` in *cmdline.txt* with the Propellerhead Balance. Unfortunately I cannot hear these clicks here with `ChunkSize=384` with my interfaces. Let's see, what happens with other interfaces.

Koshdukai:
I think I tried both, without and with 16 or 24 on Balance but I'll recheck ASAP :)

miotislucifugis:
This is a pretty exciting development.   Apologies if Im getting to far ahead,  but would it be possible to set up USB audio so it defaults to I2s (or other "internal" audio hardware)  if no USB audio source was detected and switches to USB audio out when a device was connected?

rsta2:
@miotislucifugis Yes, it's possible to detect, if an USB audio streaming 
device with the configured sample size (16 or 24 bits) is connected, but you 
have to know, which sample size the device has, which you want to use. Perhaps 
it's preferable to use 24 bits sample size by default (option soundopt=24 in 
cmdline.txt), because normally only cheap devices do not support this.

But let us see first, if more devices show this click issue. Currently there 
are three known devices, which work and one, which shows clicks. Another two 
devices are not detected, but this may be a configuration problem (soundopt=).

Koshdukai:
> Please try `soundopt=24` in _cmdline.txt_ with the Propellerhead Balance.

Just re-checked and confirmed that it doesn't make a difference. Balance still gives the "Cannot initialize function" usbdev error.
It works fine with regular linux builds or Zynthan, so there might be some specific difference on this one that might be known already on the regular builds doing usb audio handling (maybe?)

oh! speaking of weird USB audio... let me check the TE OP-1 :D

rsta2:
@Koshdukai Thanks for testing! :D

rsta2:
@paolofalcoruegg When you attach a HDMI display, you will get the MiniDexed log. I'm interested in the lines, which are starting with "uaudio" or "sndusb" (after the time).

rsta2:
[Here](https://github.com/rsta2/circle/wiki/Device-compatibility-list#usb-audio-streaming-devices) is a list of the USB audio streaming devices, which have been successfully tested with Circle. Other devices may or may not work.

Unfortunately there is no quick way to support many more devices. Each problem would have to be sorted out separately with user support, which can take longer. We [did this before](https://github.com/rsta2/circle/discussions/336) for some devices. My assumption was, that Circle would already work with more devices, but maybe this was wrong.

Finally I have to say, that Circle is a bare metal environment, which has been developed by very few developers. You cannot compare it with Linux in this field. Someone who creates a new USB audio interface, will probably test it with Linux on its own, but not with Circle. And I cannot buy all these interfaces to test them.

probonopd:
This will become really interesting if it can work for more RPi models as discussed here:

* https://github.com/rsta2/circle/issues/222#issuecomment-2497117744

Issue #713, opened 2024-09-01
Fast GPIO button clicks go undetected


miotislucifugis:
Ive recently completed a minidexed build that uses mechanical keyboard switches for GPIO buttons and Im finding that with these switches which allow for faster click action (vs the standard tiny pcb tactile switch) ,  many times the quick taps go undetected and it requires sort of a longer deliberate press to be recognized.    I assume that this is due to a switch debounce delay in software.   Is there any way to customize this? ... or can it be added as a custom setting, like the double click and longpress timeouts?

probonopd:
Possibly we can lower `BUTTONS_UPDATE_NUM_TICKS` and/or `DEBOUNCE_TIME` in https://github.com/probonopd/MiniDexed/blob/main/src/uibuttons.h.

probonopd:
Please test the build from https://github.com/probonopd/MiniDexed/pull/714 - does it improve/solve the issue?

diyelectromusic:
If the issue is detecting short presses in the first place, rather than repeated presses, then it is probably BUTTONS_UPDATE_NUM_TICKS that needs decreasing.  This would however mean that more time is spent checking buttons rather than processing audio, so we'd need to check performance on lower-end devices too to ensure audio is still ok.

DEBOUNCE_TIME I'd expect would affect how quickly a second press of the button is recognised, e.g. in a double click.

I don't recall how we came up with those numbers - I expect it was an "initial finger in the air" guess and then see how it works :)

Kevin

miotislucifugis:
(copied from my response in https://github.com/probonopd/MiniDexed/pull/714) Wow, Thanks!   This seems to be working great.  Even the fastest, lightest taps get registered.     However, Im finding that if i do a couple of fast successive taps, like 3 quick ones in a row to go back 3 steps,  only only one press gets identified....but I guess that this is a limit of the double tap timeout.

probonopd:
@diyelectromusic agree to merge this?

miotislucifugis:
this can be closed.  Everything is working fine now

miotislucifugis:
this can be closed. Everything is working fine now

PR #714, opened 2024-09-01
Lower debounce time

probonopd:
https://github.com/probonopd/MiniDexed/issues/713

### DIFF
--- a/src/uibuttons.h
+++ b/src/uibuttons.h
@@ -26,7 +26,7 @@
 #include "config.h"
 
 #define BUTTONS_UPDATE_NUM_TICKS 100
-#define DEBOUNCE_TIME 100
+#define DEBOUNCE_TIME 20
 #define MAX_GPIO_BUTTONS 9  // 5 UI buttons, 4 Program/TG Select buttons
 #define MAX_MIDI_BUTTONS 9
 #define MAX_BUTTONS (MAX_GPIO_BUTTONS+MAX_MIDI_BUTTONS)


miotislucifugis:
Wow, Thanks!   This seems to be working great.  Even the fastest taps get registered.     However, Im finding that if i do a couple of fast successive taps, like 3 quick ones in a row to go back 3 steps,  only only one press gets identified....but I guess that this is a limit of the double tap timeout.

probonopd:
@diyelectromusic agree to merge this? See @miotislucifugis last comment in https://github.com/probonopd/MiniDexed/issues/713.

At this time, I'd rather not touch `BUTTONS_UPDATE_NUM_TICKS` because it might impact overall performance.

diyelectromusic:
It's a little hard to say really, but yes, that last comment is encouraging.

I do think we should wait until someone tests this with other buttons though.  We can't be sure we haven't introduced a debouncing issue with slower/noisier buttons, so I think others ought to give it a bit of a test first.

If I get a chance, I'll load it up onto my various builds and see what it does, but that probably won't be for a few days and all my builds include capacitors for some hardware debouncing so might not be representative anyway.

It would be nice to have some proper profiling of the timers/responses involved in buttons and their scanning to put some sense to those numbers, but that would need an instrumented (e.g. with oscilloscope/GPIO outputs) build to be tested which would take a fair bit of time, so I guess we just go with the "is anyone shouting this is broken" approach for now :)

Kevin

miotislucifugis:
I can test that tomorrow.  My switches are on a daughter board and I have an other board that I can swap in that has the standard little pcb push buttons.   For what its worth, I currently see no issue w/ the encoder switch.

miotislucifugis:
Regular switches appear to be working fine in this build... at least for me

probonopd:
Merging. Please comment here if you experience any button related issues.
Thanks!

PR #743, opened 2024-11-05
Add DAW Controller support for various Arturia controllers

soyersoyer:
Some Arturia keyboards have a Display and this can be used to display the menu.
I also added support for relative encoders, which sends 61 62 63 for decrement and 66, 67, 68 for increment.

With this config the MiniDexed can be controlled well with MiniLab3's Main Knob and Display in DAW mode:
```
DAWControllerEnabled=1
```
The Home is the shift+Main click

Keylab Essential, Keylab Essential 3, KeyLab mkII work also

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,7 +9,8 @@ CMSIS_DIR = ../CMSIS_5/CMSIS
 OBJS = main.o kernel.o minidexed.o config.o userinterface.o uimenu.o \
        mididevice.o midikeyboard.o serialmididevice.o pckeyboard.o \
        sysexfileloader.o performanceconfig.o perftimer.o \
-       effect_compressor.o effect_platervbstereo.o uibuttons.o midipin.o
+       effect_compressor.o effect_platervbstereo.o uibuttons.o midipin.o \
+       dawcontroller.o
 
 OPTIMIZE = -O3
 
--- a/src/common.h
+++ b/src/common.h
@@ -2,6 +2,8 @@
 #ifndef _common_h
 #define _common_h
 
+#define ARRAY_LENGTH(x) (sizeof(x) / sizeof((x)[0]))
+
 inline long maplong(long x, long in_min, long in_max, long out_min, long out_max) {
   return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
 }
@@ -15,6 +17,12 @@ inline float32_t mapfloat(int val, int in_min, int in_max, float32_t out_min, fl
 {
   return (val - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
 }
+
+inline long mapfloatr(int val, int in_min, int in_max, float32_t out_min, float32_t out_max)
+{
+  return lround((val - in_min) * (out_max - out_min) / (in_max - in_min) + out_min);
+}
+
 
 #define constrain(amt, low, high) ({ \
   __typeof__(amt) _amt = (amt); \
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -178,11 +178,18 @@ void CConfig::Load (void)
 
 	m_nMIDIButtonCh = m_Properties.GetNumber ("MIDIButtonCh", 0);
 	m_nMIDIButtonNotes = m_Properties.GetNumber ("MIDIButtonNotes", 0);
+
 	m_nMIDIButtonPrev = m_Properties.GetNumber ("MIDIButtonPrev", 0);
 	m_nMIDIButtonNext = m_Properties.GetNumber ("MIDIButtonNext", 0);
 	m_nMIDIButtonBack = m_Properties.GetNumber ("MIDIButtonBack", 0);
 	m_nMIDIButtonSelect = m_Properties.GetNumber ("MIDIButtonSelect", 0);
 	m_nMIDIButtonHome = m_Properties.GetNumber ("MIDIButtonHome", 0);
+
+	m_MIDIButtonActionPrev = m_Properties.GetString ("MIDIButtonActionPrev", "");
+	m_MIDIButtonActionNext = m_Properties.GetString ("MIDIButtonActionNext", "");
+	m_MIDIButtonActionBack = m_Properties.GetString ("MIDIButtonActionBack", "");
+	m_MIDIButtonActionSelect = m_Properties.GetString ("MIDIButtonActionSelect", "");
+	m_MIDIButtonActionHome = m_Properties.GetString ("MIDIButtonActionHome", "");
 
 	m_nMIDIButtonPgmUp = m_Properties.GetNumber ("MIDIButtonPgmUp", 0);
 	m_nMIDIButtonPgmDown = m_Properties.GetNumber ("MIDIButtonPgmDown", 0);
@@ -190,6 +197,13 @@ void CConfig::Load (void)
 	m_nMIDIButtonBankDown = m_Properties.GetNumber ("MIDIButtonBankDown", 0);
 	m_nMIDIButtonTGUp = m_Properties.GetNumber ("MIDIButtonTGUp", 0);
 	m_nMIDIButtonTGDown = m_Properties.GetNumber ("MIDIButtonTGDown", 0);
+
+	m_MIDIButtonActionPgmUp = m_Properties.GetString ("MIDIButtonActionPgmUp", "");
+	m_MIDIButtonActionPgmDown = m_Properties.GetString ("MIDIButtonActionPgmDown", "");
+	m_MIDIButtonActionBankUp = m_Properties.GetString ("MIDIButtonActionBankUp", "");
+	m_MIDIButtonActionBankDown = m_Properties.GetString ("MIDIButtonActionBankDown", "");
+	m_MIDIButtonActionTGUp = m_Properties.GetString ("MIDIButtonActionTGUp", "");
+	m_MIDIButtonActionTGDown = m_Properties.GetString ("MIDIButtonActionTGDown", "");
 	
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
@@ -652,6 +666,31 @@ unsigned CConfig::GetMIDIButtonHome (void) const
 	return m_nMIDIButtonHome;
 }
 
+const char *CConfig::GetMIDIButtonActionPrev (void) const
+{
+	return m_MIDIButtonActionPrev.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionNext (void) const
+{
+	return m_MIDIButtonActionNext.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionBack (void) const
+{
+	return m_MIDIButtonActionBack.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionSelect (void) const
+{
+	return m_MIDIButtonActionSelect.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionHome (void) const
+{
+	return m_MIDIButtonActionHome.c_str();
+}
+
 unsigned CConfig::GetMIDIButtonPgmUp (void) const
 {
 	return m_nMIDIButtonPgmUp;
@@ -680,6 +719,36 @@ unsigned CConfig::GetMIDIButtonTGUp (void) const
 unsigned CConfig::GetMIDIButtonTGDown (void) const
 {
 	return m_nMIDIButtonTGDown;
+}
+
+const char *CConfig::GetMIDIButtonActionPgmUp (void) const
+{
+	return m_MIDIButtonActionPgmUp.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionPgmDown (void) const
+{
+	return m_MIDIButtonActionPgmDown.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionBankUp (void) const
+{
+	return m_MIDIButtonActionBankUp.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionBankDown (void) const
+{
+	return m_MIDIButtonActionBankDown.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionTGUp (void) const
+{
+	return m_MIDIButtonActionTGUp.c_str();
+}
+
+const char *CConfig::GetMIDIButtonActionTGDown (void) const
+{
+	return m_MIDIButtonActionTGDown.c_str();
 }
 
 bool CConfig::GetEncoderEnabled (void) const
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -204,7 +204,9 @@ void CConfig::Load (void)
 	m_MIDIButtonActionBankDown = m_Properties.GetString ("MIDIButtonActionBankDown", "");
 	m_MIDIButtonActionTGUp = m_Properties.GetString ("MIDIButtonActionTGUp", "");
 	m_MIDIButtonActionTGDown = m_Properties.GetString ("MIDIButtonActionTGDown", "");
-	
+
+	m_bDAWControllerEnabled = m_Properties.GetNumber ("DAWControllerEnabled", 0) != 0;
+
 	m_bEncoderEnabled = m_Properties.GetNumber ("EncoderEnabled", 0) != 0;
 	m_nEncoderPinClock = m_Properties.GetNumber ("EncoderPinClock", 10);
 	m_nEncoderPinData = m_Properties.GetNumber ("EncoderPinData", 9);
@@ -749,6 +751,11 @@ const char *CConfig::GetMIDIButtonActionTGUp (void) const
 const char *CConfig::GetMIDIButtonActionTGDown (void) const
 {
 	return m_MIDIButtonActionTGDown.c_str();
+}
+
+bool CConfig::GetDAWControllerEnabled (void) const
+{
+	return m_bDAWControllerEnabled;
 }
 
 bool CConfig::GetEncoderEnabled (void) const
--- a/src/config.h
+++ b/src/config.h
@@ -240,6 +240,8 @@ public:
 	const char *GetMIDIButtonActionTGUp (void) const;
 	const char *GetMIDIButtonActionTGDown (void) const;
 
+	bool GetDAWControllerEnabled (void) const;
+
 	// KY-040 Rotary Encoder
 	// GPIO pin numbers are chip numbers, not header positions
 	bool GetEncoderEnabled (void) const;
@@ -370,6 +372,8 @@ private:
 	unsigned m_nMIDIButtonBankDown;
 	unsigned m_nMIDIButtonTGUp;
 	unsigned m_nMIDIButtonTGDown;
+
+	bool m_bDAWControllerEnabled;
 
 	bool m_bEncoderEnabled;
 	unsigned m_nEncoderPinClock;
--- a/src/config.h
+++ b/src/config.h
@@ -210,11 +210,19 @@ public:
 	// MIDI Button Navigation
 	unsigned GetMIDIButtonCh   (void) const;
 	unsigned GetMIDIButtonNotes (void) const;
+
 	unsigned GetMIDIButtonPrev (void) const;
 	unsigned GetMIDIButtonNext (void) const;
 	unsigned GetMIDIButtonBack (void) const;
 	unsigned GetMIDIButtonSelect (void) const;
 	unsigned GetMIDIButtonHome (void) const;
+
+	// Action type for Midi buttons: "click", "doubleclick", "longpress", ""
+	const char *GetMIDIButtonActionPrev (void) const;
+	const char *GetMIDIButtonActionNext (void) const;
+	const char *GetMIDIButtonActionBack (void) const;
+	const char *GetMIDIButtonActionSelect (void) const;
+	const char *GetMIDIButtonActionHome (void) const;
 
 	// MIDI Button Program and TG Selection
 	unsigned GetMIDIButtonPgmUp (void) const;
@@ -224,6 +232,14 @@ public:
 	unsigned GetMIDIButtonTGUp (void) const;
 	unsigned GetMIDIButtonTGDown (void) const;
 	
+	// Action type for buttons: "click", "doubleclick", "longpress", ""
+	const char *GetMIDIButtonActionPgmUp (void) const;
+	const char *GetMIDIButtonActionPgmDown (void) const;
+	const char *GetMIDIButtonActionBankUp (void) const;
+	const char *GetMIDIButtonActionBankDown (void) const;
+	const char *GetMIDIButtonActionTGUp (void) const;
+	const char *GetMIDIButtonActionTGDown (void) const;
+
 	// KY-040 Rotary Encoder
 	// GPIO pin numbers are chip numbers, not header positions
 	bool GetEncoderEnabled (void) const;
@@ -326,6 +342,18 @@ private:
 	std::string m_ButtonActionTGUp;
 	std::string m_ButtonActionTGDown;
 	
+	std::string m_MIDIButtonActionPrev;
+	std::string m_MIDIButtonActionNext;
+	std::string m_MIDIButtonActionBack;
+	std::string m_MIDIButtonActionSelect;
+	std::string m_MIDIButtonActionHome;
+	std::string m_MIDIButtonActionPgmUp;
+	std::string m_MIDIButtonActionPgmDown;
+	std::string m_MIDIButtonActionBankUp;
+	std::string m_MIDIButtonActionBankDown;
+	std::string m_MIDIButtonActionTGUp;
+	std::string m_MIDIButtonActionTGDown;
+
 	unsigned m_nDoubleClickTimeout;
 	unsigned m_nLongPressTimeout;
 
--- a/src/config.h
+++ b/src/config.h
@@ -217,7 +217,7 @@ public:
 	unsigned GetMIDIButtonSelect (void) const;
 	unsigned GetMIDIButtonHome (void) const;
 
-	// Action type for Midi buttons: "click", "doubleclick", "longpress", ""
+	// Action type for Midi buttons: "click", "doubleclick", "longpress", "dec", "inc", ""
 	const char *GetMIDIButtonActionPrev (void) const;
 	const char *GetMIDIButtonActionNext (void) const;
 	const char *GetMIDIButtonActionBack (void) const;
@@ -232,7 +232,7 @@ public:
 	unsigned GetMIDIButtonTGUp (void) const;
 	unsigned GetMIDIButtonTGDown (void) const;
 	
-	// Action type for buttons: "click", "doubleclick", "longpress", ""
+	// Action type for buttons: "click", "doubleclick", "longpress", "dec", "inc", ""
 	const char *GetMIDIButtonActionPgmUp (void) const;
 	const char *GetMIDIButtonActionPgmDown (void) const;
 	const char *GetMIDIButtonActionBankUp (void) const;
--- a/src/dawcontroller.cpp
+++ b/src/dawcontroller.cpp
@@ -1,0 +1,1648 @@
+// MiniDexed - Dexed FM synthesizer for bare metal Raspberry Pi
+// Copyright (C) 2024  The MiniDexed Team
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+#include <circle/string.h>
+
+#include "dawcontroller.h"
+#include "midikeyboard.h"
+#include "minidexed.h"
+
+#define LINELEN 18
+
+#define MIDI_DAW_CHANGE 0b10000
+#define MIDI_DAW_VOICE 1
+#define MIDI_DAW_TOGGLE_MONO 3
+#define MIDI_DAW_TOGGLE_PORTA_GLISS 4
+#define MIDI_DAW_TOGGLE_TG 5
+#define MIDI_DAW_SELECT_TG 6
+#define MIDI_DAW_SELECT_CHAN_TG 7
+#define MIDI_DAW_MENU_SELECT 8
+#define MIDI_DAW_MENU_BACK 9
+#define MIDI_DAW_MENU_PREV 10
+#define MIDI_DAW_MENU_NEXT 11
+#define MIDI_DAW_MENU_PRESS_PREV 12
+#define MIDI_DAW_MENU_PRESS_NEXT 13
+#define MIDI_DAW_MENU_HOME 14
+#define MIDI_DAW_DISPLAY_MODE_TOGGLE 17
+#define MIDI_DAW_ENC_VALUES_TOGGLE 18
+#define MIDI_DAW_ENC_0 20
+#define MIDI_DAW_ENC_1 21
+#define MIDI_DAW_ENC_2 22
+#define MIDI_DAW_ENC_3 23
+#define MIDI_DAW_ENC_4 24
+#define MIDI_DAW_ENC_5 25
+#define MIDI_DAW_ENC_6 26
+#define MIDI_DAW_ENC_7 27
+
+
+static void ArturiaDisplayWrite (CMIDIKeyboard *pKeyboard, const u8 *pHdr, const unsigned nHdrSize,
+	size_t nLineMaxLen, const bool bFill1, const bool bFill2, const char *pMenu,
+	const char *pParam, const char *pValue,
+	const bool bArrowLeft, const bool bArrowRight)
+{
+	size_t nParamLen = std::min (nLineMaxLen, strlen (pParam));
+	size_t nMenuLen = strlen (pMenu);
+	size_t nFill1Len = bFill1 && nLineMaxLen > nParamLen + nMenuLen ?
+		nLineMaxLen - nParamLen - nMenuLen : 1;
+
+	nFill1Len = std:: min (nLineMaxLen - nParamLen, nFill1Len);
+	nMenuLen = std::min (nLineMaxLen - nParamLen - nFill1Len, nMenuLen);
+
+	size_t nLine1Len = nParamLen + nFill1Len + nMenuLen;
+
+	size_t nValueLen = std::min (nLineMaxLen - 2, strlen (pValue));
+	size_t nFill2Len = bFill2 ? nLineMaxLen - 2 - nValueLen : 0;
+	size_t nLine2Len = 1 + nValueLen + nFill2Len + 1;
+
+	size_t nOffset = 0;
+
+	uint8_t pLines[nHdrSize + nLine1Len + 2 + nLine2Len  + 2];
+
+	memcpy (pLines, pHdr, nHdrSize);
+	nOffset += nHdrSize;
+
+	memcpy (&pLines[nOffset], pParam, nParamLen);
+	nOffset += nParamLen;
+
+	memset (&pLines[nOffset], ' ', nFill1Len);
+	nOffset += nFill1Len;
+
+	memcpy (&pLines[nOffset], pMenu, nMenuLen);
+	nOffset += nMenuLen;
+
+	pLines[nOffset++] = 0x00;
+	pLines[nOffset++] = 0x02;
+
+	pLines[nOffset++] = bArrowLeft ? '<' : ' ';
+
+	memcpy (&pLines[nOffset], pValue, nValueLen);
+	nOffset += nValueLen;
+
+	memset (&pLines[nOffset], ' ', nFill2Len);
+	nOffset += nFill2Len;
+
+	pLines[nOffset++] = bArrowRight ? '>' : ' ';
+
+	pLines[nOffset++] = 0x00;
+	pLines[nOffset++] = 0xF7;
+
+	// block character (0xFF) is not supported over MIDI, change to 0x7f
+	for (unsigned i = 0; i < sizeof pLines; ++i)
+		if (pLines[i] == 0xFF)
+			pLines[i] = 0x7F;
+
+	pKeyboard->SendDebounce (pLines, nOffset, 0);
+}
+
+enum ControlType
+{
+	CT_KNOB = 3,
+	CT_FADER,
+	CT_PAD,
+};
+
+enum HideAfter
+{
+	HA_NO = 0,
+	HA_YES = 2,
+};
+
+static std::string to_percent (int nValue)
+{
+	return std::to_string(mapfloatr (nValue, 0, 127, 0, 100)) + "%";
+}
+
+std::string to_midi_channel (int nValue)
+{
+	switch (nValue)
+	{
+	case CMIDIDevice::OmniMode:	return "Omni";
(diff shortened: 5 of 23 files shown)


probonopd:
Thank you very much for this contribution @soyersoyer.

Unfortunately I don't have the matching hardware to test this with, but I am sure many users will find this extremely useful.

Semi off topic: Do you know any good/straightforward documentation how the display and encoder on the Keystep communicates with the host over USB? It just so happens that recently I was thinking about building something the other way around (unrelated to MiniDexed), a simple microcontroller + display + encoder combination that does what a Keystep can do (e.g., to browse presets of the free Analog Lab Play), but without needing an actual Keystep.

diyelectromusic:
> Unfortunately I don't have the matching hardware to test this with, but I am sure many users will find this extremely useful.

Well users of this specific MIDI controller maybe, but I'm really not sure about hardcoding support for just one vendor's device in MiniDexed... In that SysEx message, 0x00, 0x20, 0x6B is the Arturia SysEx Vendor ID - right?  It's a little hard to tell as I didn't seem to be able to find any mention of the MIDI implementation of Arturia devices from the manufacturer themself.  Do you have a link?

If we were to go down this route, then I think we'd have to have some way of querying a connected device and mapping MiniDexed features to the features of the device itself rather than hard-code in specific SysEx messages in my view.

Do we know what other controllers support displays and how they do it?

@probonopd you weren't keen on me adding bespoke SysEx to MiniDexed to support this kind of thing between MiniDexeds, so I'm a little surprised you seem quite happy to be considering adding bespoke vendor MIDI messages other than Yamaha...?

Kevin

soyersoyer:
> how the display and encoder on the Keystep communicates with the host over USB?

Do you think KeyLab? KeyStep doesn't have a display. In the MiniLab 3 there is an 'Arturia' program for Analog Lab, but I don't know how it communicates. Probably after an initial model query, the main Encoder sends relative CC-s, the other encoders send absolute CCs, the Analog lab sets the encoder values on preset change and updates the display with sysex.  If you think, I can capture it.

For the DAW mode, I found this:
https://github.com/PrzemekBarski/arturia-keylab-essential-mk3-programming-guide

>  hardcoding support for just one vendor's device in MiniDexed...

This is a very popular controller, so I think it might be worth adding.

Additional vendors may be added later.

> In that SysEx message, 0x00, 0x20, 0x6B is the Arturia SysEx Vendor ID - right?

Yes

> Do you have a link?

From Arturia, No.
I found the link above, and some Ableton scripts like these:
https://github.com/gluon/AbletonLive11_MIDIRemoteScripts/blob/main/MiniLab_3/midi.py
https://github.com/bitwig/bitwig-extensions/blob/api-18/src/main/java/com/bitwig/extensions/controllers/arturia/minilab3/SysExHandler.java

We can send generic device inquiry with
F0 7E 7F 06 01 F7
So in theory we can match the model, but we don't have a good spec or access to devices, so we'll probably limit the users by only enabling features for certain models without them being able to control the features.

You are right, the DAWDisplayEnabled config option is too broad and should enable some auto-detection mode.
So I think for this we should use the DAWDisplayArturiaEnabled=1 or DAWDisplay=arturia option.

omersiar:
@soyersoyer You beat me to it while planning to do the same.  It works with my MiniLab3 (firmware 1.1.1).  The only issue is the Dumping MIDI to console which seems not reliable.

Also you don't configure the any of the available sliders, knobs to actual functions like Pan, Volume etc. Let me take a look.

soyersoyer:
I refined it a bit. Now it does a sysex inquiry and only sends sysex if it is MiniLab3 or KeyLab 3 Essential. Does this fix your MIDI issue?

omersiar:
Screenshot 2024-11-10 213302
I currently use like this

> And how to set them up?
Pad RGB can be controlled also you can assign what knobs do and how to represent them.
https://github.com/mhugo/sysex
https://gist.github.com/Janiczek/04a87c2534b9d1435a1d8159c742d260

soyersoyer:
How should they work? Should they send CC-s to channel 1?

soyersoyer:
Now, it maps the encoders. Also updates the encoders values at performance load from TG0.
The order is not the same as you suggested yet.
Isn't the volume better on fader?
I'm a bit unsure about which TG it should set and where to get the values ​​from.

Unfortunately the pads cannot be configured to send CC-s in DAW mode, they send notes only on channel 10.

What I want also:
If a  new value appears from MIDI CC, it would be good to display it for a little time.
Map the MIDI CC buttons automatically.
Remap the pad notes to MIDI CC-s.

soyersoyer:
I think the encoders should set the parameters based on the menu, so if the display is on the TG1, it should set the TG1 parameters, if on the performance, it should set all TG-s parameters, if it is on the effects it should set the effect parameters. I will try to make this later.

soyersoyer:
> Master volume only exposed via Sysex I believe

Yeah, but in performance files, all TGs are usually connected to channel 1.

probonopd:
In case this exclusively works only with Arturia devices, maybe it should be named `ArturiaDAWControllerEnabled`?
Also, I have no device to test with, so I can't test whether any future changes might break this.

soyersoyer:
Additional controllers can be added later if desired. It sends a SysEX inquiry at startup and only does anything if it knows the controller. So I think the DAWControllerEnabled name is fine here.

I can test later, no problem.

Unfortunately, the commit became larger than I initially thought. Should I break it into smaller parts, or is this okay?

probonopd:
I see. No need to break it into smaller pieces as far as I am concerned. My main gripe is that neither @diyelectromusic nor I can test nor maintain this because we can't test it due to the lack of a compatible device. So there is a likelihood of this breaking in the future and @diyelectromusic is worried about the increasing code complexity.

Spacepadrille:
Hello, I have tested the [MiniDexed_2024-11-12-8a5fb08 with my Arturia MiniLab3 ans all works like a charm ! Congratulations for this implementation, it is very usefull. I appreciate a lot to have infos on both screens (minilan3 + original minidexed standard screen). I appreciate a lot to have cutoff, reso, reverb, pan and volume mapped. There are still 3 encoders and 3 faders free for more control like portamento or glide but I don't know how to do that  ;-) If there are no more usefull control to add, maybe it could be nice to have the performance volume on an encoder and keep the daw fonctionnality of the faders (track volume, sends, pan) ? 
I noticed that when I change the bank selected, I have to put again the MiniLab on midi channel 1, or there is no sound....
Thanks again @soyersoyer for these nex features, and all the team for this fantastic project !

soyersoyer:
@Spacepadrille: Thank you for testing.


> There are still 3 encoders and 3 faders free for more control like portamento or glide but I don't know how to do that ;-)

What do you think if PortamentoTime gets an encoder and PortamentoMode gets a pad?

For that I think the MIDI CC 5 (Portamento Time) and CC 65 (Portamento on/off) should be connected to control all TG-s.
Then they can be mapped similar to the other controls in dawcontroller.cpp.


> I noticed that when I change the bank selected, I have to put again the MiniLab on midi channel 1, or there is no sound....

Do you change the Minilab's Pad bank or the Minidexed's Performance bank?

> There are some display glitches if I turn the encoders fast.

This is because at parameter change, both the UI and the DAWController send screen updates. I'll try to fix it.

Spacepadrille:
Hi there,
_"What do you think if PortamentoTime gets an encoder and PortamentoMode gets a pad?"_
Seems nice :-)

_" Do you change the Minilab's Pad bank or the Minidexed's Performance bank?"
The miniDexed one. But this is not linked with the arturia stuff, I have the same issue with the "regular" MiniDexed code

What about using the 8 pads to ON/OFF the 8 TG's ? Can be very usefull in the case of a preset with different voices on each TG. The sustain pad is quite similar to the "hold" button of the MiniLab3, and the "stop all sounds" pad can be moved to a free encoder ?

Really enjoying this stuff, my MiniLab3 is now dedicated to the MiniDexed ;-)

Spacepadrille:
All works perfectly well :-) 

> How do you think OFF/ON should be implemented? Turn their volume down to zero or 100?
That is not perfect as tou can have different volumes set per voice, and the on/off will put all at 100%...
Maybe by changing the midi channel of the TG to mute it ?
Putting that on B bank of pads is a great solution.

soyersoyer:
> Maybe by changing the midi channel of the TG to mute it ?

Or use one of the undefined MIDI CC's bits to activate/inactivate the TG-s.  The tx816 doesn't have anything like this, so it probably needs more thought/iteration. 
But this is not good either, because 7 bits are not enough to activate 8 TGs, and if the LSB is also used, then 14 bits are not enough to activate 16 TGs (rpi5 can handle 16TG).
It might be possible to solve this with a Minidexed specific SysEX.
Or this could be solved specifically for DAWController, but then it is really untestable with other hardware.

> Did you notice that fader 1 is volume of TG1 & TG2, fader 2 volume of TG3 & TG4 and so on ? I observed that while editing different voices of a preset. In performance mode fader 1 is master volume...

Fader 1 sets the volume on MIDI Channel 1, Fader 2 on Channel 2, Fader 3 on 3 , Fader 4 on 4.
Do you use different channels for TG-s?

Now that my LCD has arrived, I can see how the Volume/Pan menu should look. It needs a fix :)

soyersoyer:
initial support for KeyLab mkII

soyersoyer:
This includes the portamento cc [fix](https://codeberg.org/dcoredump/Synth_Dexed/commit/26d9cadb7f596c5d188f406f208733e6de3ff8ca).

probonopd:
Just to understand this correctly. So @soyersoyer, does this essentially implement
https://gist.github.com/Janiczek/04a87c2534b9d1435a1d8159c742d260?
Meaning we are using the Arturia display and controller and need no display/keys/encoder on the RPi at all?

GertSchepens:
As a random tester, I confirm this replaces the display/keys/encoder for me. I can use minidexed with just a pi and the Arturia keyboard. And with extra comfort thanks to the rest of the coloured buttons and inputs working.

I love it :) Thanks for the great work!!

soyersoyer:
@probonopd: yes

probonopd:
Since I don't have suitable Arturia hardware but want to test this, I am trying to mock it:

https://github.com/probonopd/mock-arturia-controller

soyersoyer:
This will also be good if someone wants to build a standalone MiniDexed controller.

probonopd:
@omersiar, @soyersoyer: Do you know what the Arturia controller sends to AnalogLab (the host PC standalone software) when one presses/clicks the encoder button?

So far my mock can display messages on the screen, and can browse through presets, but I am lacking the "OK" button to select the displayed preset.

soyersoyer:
@probonopd: That's great news. MiniLab 3 sends a MIDI CC 115 for clicks in Arturia mode.

soyersoyer:
Now the first 8 TG can be disabled on Pad Bank B.
The color of the pads is based on the TG's channel.
I was wondering what double-click or long-press could be used for. E.g. selecting TGs by channel for editing.
If you have any ideas, feel free to write them.

Spacepadrille:
Hi there, thanks for the new update. TG enable/disable is a great feature, really usefull. I use the Minilab3 arpeggiator, hold button, and I play with the TG's while it runs. A lot of fun and a really nice feature for lives.

probonopd:
@soyersoyer thanks. Got it. `CC 115 value 0...63:   OK/Enter`.

GertSchepens:
awesome update!


Some usability thoughts from a Minilab 3 (though what do I know, really; apologies up front for the long post :/ )

I get that the colours of the pad depend on the midi channel; and they're off when the channel is off.
* While that's kinda nice, I'm not sure that's necessarily that impactful?
* The visual difference between dimmed (disabled) and enabled is very small to my eyes; it feels like different colours might be more usable.. Or at least more of a difference in the state colours.

Pushing the pad changes the colour, but doesn't give screen feedback on what's happening. Most actions (turning knobs etc) show on the screen what's happening, consistency would be convenient,



Though what I'd really enjoy, would be to be able to edit the TG settings using the inputs. The Minilab has 4 sliders, 8 knobs and 8 pads; it would be great if those would control those many settings.

For example; I'm in bank B; and long-push the 1st pad.
That opens the TG1 menu on the screen. With the menu open on that spot, the knobs and sliders change the settings in there. For example, I turn knob 1 and it changes the voice, showing the current selection on screen. I turn knob 2 and that changes the bank. Slider 1 changes the volume; etc..
Long pressing the 2nd pad switches to the tg2 menu; all inputs now affect those settings.

And then maybe the same logic for the voices; when a voice is selected in the menu, the inputs control (some?) of the settings there.

And of course, when in the root of the menu, not in a TG or deeper,  the inputs still control the general settings like now.




And then maybe a final consideration ..
Currently, the input config is (if I'm not mistaken) something like ..

Slider 1-4  = main volume

Knobs
1 cutoff 
2 resonance 
3 Reverb 
4 detune
5 portamento time 
6 (nothing)
7 (nothing)
8 pan

Pads
1 (sound)
2 portamento 
3 sostenuto 
4 sustain
5 all sound off 
6 (sound)
7 (sound)
8 (sound)

It feels like there's some room to put more control in there; for example more of the FX? (Haven't properly looked at that yet tho) And maybe no duplication on the sliders :)



A lot of text, for what it's worth, which I realize really might not be that much :) Thanks for reading.

GertSchepens:
_Another testing result._

At boot, TG 1, 2, 3 are red because they are on midi1
I disable all of them.
When I edit TG8, from disabled to midi Omni, **the pad does not light up**.
When I play, I properly hear TG8
When I tap the pad, the colour lights up muted. (properly disabling it) From there on, it works as expected

It seems like enabling the TG by adding the midi isn't updated on the keyboard.
Nor is the colour updated when I change midi channels; 1 red doesn't change to omni white till I poke the pad



_On a related note;_
When I mute with the pad; I can't unmute via the menu. I have no idea what TG menu item is changed by the pad, everything seems the same and like it should work. The only way I find to unmute (or reenable) is via the pad.

I would expect to be able to do that via the menu equally.

soyersoyer:
Thank you all for the testing!

New version:
 - selecting a performance enables all TGs
 - distinguishable colors for channels, darker disabled state
 - add the missing screen feedbacks
 - mono/poly mode button on Bank A Pad 1
 - voice selection on knob 6
 - algorithm selection on knob 7 (fun but not sure it's useful)
 - pads update on MIDI channel changes
 - faders control the channels of the performance (not just fixed channels 1-4)

Menu integration will come later.

Spacepadrille:
It's becoming impressive... I love the new features. Still playing with it. Still a bit confuse with some things (voice bank, voice selection per TG, faders...). I will do more tests and come back to comment when all is more clear for me.
@soyersoyer : you are nailing it, awesome work !

soyersoyer:
Keylab Essential display and wheel support in DAW mode

probonopd:
@soyersoyer how do we enable this? Is `minidexed.ini` editing needed?

Now that I have a KeyLab Essential 66 emulator, I'd like to test it.

It doesn't automatically switch this feature on e.g., depending on the USB ID of the attached controller (plug and play), does it?

soyersoyer:
@probonopd Probably yes, DAW mode is a bit different than AnalogLab mode.
Minidexed identifies the controller with a sysex inquiry 0xF0, 0x7E, 0x7F, 0x06, 0x01, 0xF7.
The Keylab Essential 61 responds with 0xF0, 0x7E, 0x7F, 0x06, 0x02, 0x00, 0x20, 0x6B, 0x02, 0x00, 0x05, 0x54, 0xAA, 0xBB, 0xCC, 0xDD, 0xF7 (AA BB CC DD is the firmware version)
https://docs.rs/midi-control/latest/midi_control/vendor/arturia/index.html
Then it sets the DAW mode into mackie with 0xF0, 0x00, 0x20, 0x6B, 0x7F, 0x42, 0x02, 0x00, 0x40, 0x51, 0x00, 0xF7
and uses the display with 0xF0, 0x00, 0x20, 0x6B, 0x7F, 0x42, 0x04, 0x00, 0x60, 0x01 line1, 0x02, line2, 0xF7
The main wheel sends MIDI CC messages with CC 0x3C (right 1, left 127),
and MIDI_NOTE_ON 0x54 on click.
These messages are sent on cable 1, which has a MIDI name that ends in DAW or something similar.
I don't have such a controller, I just read the bitwig [code](https://github.com/bitwig/bitwig-extensions/blob/da7d70e73cc055475d63ac6c7de17e69f89f4993/src/main/java/com/bitwig/extensions/controllers/arturia/keylab/essential/ArturiaKeylabEssentialControllerExtension.java)

probonopd:
Hmmm. With my simulator, right now MiniDexed plays the notes (e.g., when I click the encoder button).

soyersoyer:
Send it on "cable" 1.
You should see something like this with `amidi -l`:
```
Dir Device    Name
IO  hw:1,0,0  Arturia KeyLab Essential 88 MID
IO  hw:1,0,1  Arturia KeyLab Essential 88 DAW
```
Minidexed uses the "cable" 1 (Arturia KeyLab Essential 88 DAW) for the main encoder.

probonopd:
How do you determine the "cable"? Based on the Name? (My device can only emulate one MIDI port.)

soyersoyer:
The Circle does it, I think based on the MIDI port order.

probonopd:
I see. So I need to emulate 2 MIDI ports. (Hopefully I can find a way to do that.)

soyersoyer:
The cable nmber is the first half of the USB MIDI packet.
https://www.usb.org/sites/default/files/midi10.pdf
https://github.com/rsta2/circle/blob/c243194670720c8a13e6a553654b6d26ec7845a0/lib/usb/usbmidi.cpp#L250

probonopd:
@soyersoyer: Since CircuitPython currently only allows for 1 MIDI port, would it be possible to determine the "cable" a bit differently? I am assuming you are currently using "the second cable" (since this is what Arturia has). Would it be possible to use "the largest cable number" instead? On Arturia devices, this would still be "the second cable", but in my mock implementation, it would be "the first cable", since there is only one. Would that be doable?

soyersoyer:
@probonopd: I don't think I can query the list of usb interfaces from Circle.
Why don't you use MiniLab 3 mock? It only uses the cable 0.
Extending the circuitpython could also be a solution, tud_midi_stream_write() has [cable_num](https://github.com/adafruit/circuitpython/blob/2d2fa28b4f290cba9ba403d8608cfadf043a0441/shared-module/usb_midi/PortOut.c#L13) parameter.

probonopd:
Thanks @soyersoyer. Well, the idea is to eventually mock all the different models... I will see whether I can get CircuitPython to use cable 1 rather than 0 somehow.

soyersoyer:
News:
 - Pad colors are more consistent
 - Channel aftertouch on Pad 8 (can be fun for AT performances: SawPressive, Wilkinson, Irish-flute, Hybrid-Dreams, LFO-reset, Airbeat, mdX-strg.AT, Tyrell Corp)

probonopd:
Looks like I have some extent of Minilab3 emulation going. I now see MiniDexed messages on the screen of my emulator. It greets me with `MiniDexed On MiniLab 3``.

When I operate MiniDexed using a rotary encoder attached to the MiniDexed device, then the display on my emulator updates its display accordingly.

But the rotary encoder and button on the emulator does not seem to trigger actions yet. Can you show me in you code which messages it expects to get from the Minilab3 for counterclockwise, clockwise, and click? (Is it expecting Mackie/MCU or "Arturia mode"?) Thanks.

soyersoyer:
It works in DAW mode, and in DAW mode it sends different values than 'Arturia mode'.
CC 28 dec/inc = prev/next
CC 118 click = select
CC 118 longpress = back
CC 119 click = home
https://github.com/probonopd/MiniDexed/blob/1fc746a4976fcf0e7a4d2a0e816bac5d5ec0e844/src/dawcontroller.cpp#L324

probonopd:
Thanks @soyersoyer. My Minilab3 emulation is starting to work👍 
But I am missing the `<` and `>` symbols. They show up on the LCD attached to MiniDexed but not on the Minilab3. Is this intentional?

soyersoyer:
Sort of. The arrows don't look good on the Minilab 3 display. It doesn't use monospaced fonts and there is some indentation on the second line.

soyersoyer:
Added the arrows. It's a bit confusing that the top menu is circular, but the arrows don't indicate that.

soyersoyer:
Unfortunately it forgets the scrollbar value, so it no longer draws the left arrow after any autohide Info display (knob/fader/pad). It was probably intended as an autohide control, but we don't want to use it that way. So I reverted the version without arrows.

probonopd:
@soyersoyer after testing this some more, I am quite confident that this will be extremely popular among owners of Arturia controllers (I could even imagine people who don't want to build their own MiniDexed devices to just buy e.g., a MiniLab 3 and attach it to a Pi.)

Great work!

I'd really appreciate if we could show the arrows though. Without them it is not clear in which direction(s) you can move and it makes operation inconsistent from the regular MiniDexed. If you really, really think the arrows annoy you then we should probably add a configuration option to disable them. (Just an idea, probably it would look better to show `<Cinematic >` rather than `< Cinematic     >`. In other words, replace multiple spaces by just one - that way it should hopefully look good even with proportional fonts?)

Question, when I load a performance I see quite some flickering on the screen. Is there more than one display message sent during the loading?

Are there any other buttons beside Shift and the encoder button that interact with the menu operation in MiniDexed?

soyersoyer:
The problem was that the thickness of '<' and '>' were different. But I found a way to make them the same.
`< Cinematic        >` looks better than `< Cinematic >`
By the way, I didn't miss the arrows at all.

Which screen are you seeing flickering? While loading a performance the menu reloads the screen a lot. I added a debouncer to speed things up. (as a workaround)

> Are there any other buttons beside Shift and the encoder button that interact with the menu operation in MiniDexed?

The encoders, faders, pads show their values on the screen, but they are independent from the menu.

probonopd:
Much better with the arrows! 👍

But now the fist line of the display is entirely missing (empty), I only see the second line.

(At least on my emulated controller which currently still just uses a 16x2 character LCD.)

About the flickering, don't worry about it too much, I just was curious if it might be indicative about more display messages being sent than needed. More a curiosity than a real concern.

soyersoyer:
The display message has changed from
```0xF0, 0x00, 0x20, 0x6B, 0x7F, 0x42, 0x04, 0x02, 0x60, 0x12, 0x01, Line1, 0x00, 0x02, Line2, 0x00, 0xF7```
to 
```0xF0, 0x00, 0x20, 0x6B, 0x7F, 0x42, 0x04, 0x02, 0x60, 0x1f, 0x07, 0x01, 0x00, 0x00, 0x01, 0x00, 0x01, Line1, 0x00, 0x02, Line2, 0x00, 0xF7```

soyersoyer:
Why are these commits good here and in the other PR? :)

probonopd:
> Why are these commits good here and in the other PR? :)

So that I have a way to look at MiniDexed over the network while playing with this stuff. Very helpful for me ;-)

soyersoyer:
News:
- Pressing (harder) one of PAD B pad enables only its TG
- Pressing (harder) Portamento (PAD A 2) enables Portamento Glissando
- The Encoders and Faders now control the first enabled TG, update themself from that TG and send the messages to the TG's channel.
- arrows in the display (smaller fonts with this mode?)

probonopd:
Btw: AnalogLab also shows arrows on the display...

soyersoyer:
I haven't tried, I'm not sure if it does.
https://youtu.be/1IrLYMlyEDU?t=160

Spacepadrille:
Hi there. I don't understand well the reason to show the arrows.. It makes the patch name very tiny on Minilab3. 
But it's OK ;-)
Pressing harder on a B pad does not enable only the concerned TG...
Maybe something different on Minilab3 and other Arturia keyboards ?

probonopd:
@Spacepadrille
The arrows show in which direction you can rotate the knob or which arrow buttons can be used, and what they change.

If we want to save 2 characters, we can remove the spaces between the arrow and the text.

soyersoyer:
> Pressing harder on a B pad does not enable only the concerned TG...
> Maybe something different on Minilab3 and other Arturia keyboards ?

I routed the Pad Aftertouch messages to work as buttons, but only the Aftertouch with value 127 for the Bank B.
My pads aftertouch is in exponential mode, so the maximum value was easyly achievable for me. But this is more difficult in linear or logarithmic mode.
Btw, you can change the mode in MIDI Control Center (Win/Mac) or in Sysex-Controls (Linux).

But now it works on any Aftertouch message, so you don't have to press hard, but it's still easy to use.

Thanks.

soyersoyer:
I looked at the rollable UXs around me. One type is that you can always roll them around. The other type is that you can't always roll them around, but you only know this after rolling, that nothing has changed. If you only have one scroller, I don't think the arrows provide very useful information. If you have arrow buttons, it might be more useful.

soyersoyer:
This is the previous mode, bigger second line, but different looking arrows, (it doesn't seem so here):
Screenshot From 2024-12-22 15-27-02

The actual mde, bigger first line, the arrows are similar:
Screenshot From 2024-12-22 15-26-56

soyersoyer:
It would also be a good idea to have the screen show the current encoder assignment by default and only show the MiniDexed menu as a popup when using the main encoder. Like in arpeggiator mode. Then the built-in arrows would also be usable.

https://youtu.be/ELOOoc9vmOQ?t=123

Spacepadrille:
The new way aftertouch works on pads for solo one TG works like a charm on my Minilab3. 
For the screens displays, what is the more useful information ? The preset name or the upper line ? The differents arrow are not a big deal :-) 
All this stuff is so great !

soyersoyer:
> For the screens displays, what is the more useful information ? The preset name or the upper line ?

I think the preset name.

> The differents arrow are not a big deal :-)

Yes, my OCD just bites me ;)

soyersoyer:
News:
- Longpressing a Bank B Pad enables/disables all TGs on that TG's channel.

soyersoyer:
News:
- Hold mode (If you enable it on MiniLab3, it will also work on other connected keyboards)

soyersoyer:
News:
- handling longpress timeout
- no longer changes the configuration, uses the same click/longpress mechanism for navigation (this means that multiple controllers can be used for the navigation)

GertSchepens:
Wonderful work! Quick little playing around with the new version; I love how this version puts a lot of controls on the keyboard; ready for a lot of fun. I didn't expect to edit all the enabled instruments, but after poking around; I do kinda like it.

To (begrudgingly) weigh in on the < > conversation.. I do like what it does for the usability and consistency of the UI. It's clearly shows where in the menu you are, etc.
However, it hurts my aesthetic sensibilities to see the menu text bounce around.. It looks like all text is centered? Which works nicely if it's just the words; however with the menu elements, that just makes things bounce around.. In the TG menu most of all; "Voice    >" going to "< Bank    >", the word already jumps; and then "< Reverb-Send    >" is longer again than all else; and "< edit voice" jumps out right again.
Maybe ensuring the string is always the same length and format might help? Ie "  Voice     >" with a space instead of the <; and the same on the right; and maybe padding the space on the right dependent on the length of the menu text? Though I suppose the characters aren't fixed width anyway :/

Different question; the display of my pirate audio hat now shows "Minidexed" and the build number. Is this a testing setting? Can I Set a parameter to see the minidexed menu as usual instead?


It's awesome to see this evolve so rapidly!

soyersoyer:
By skipping the Minilab timer, I managed to use good looking arrows.

@GertSchepens: Does your display not work with this version? When did it break?

I have added some menu integration.
After 2 seconds it shows the current encoder assignment.
The current values ​​are displayed with the shift key.
You can change the assignment page with shift+main knob.
The effect,Main/ TG, Voice, and OP menus have pages, which can be accessed by opening the menus.

The encoders on the TG page adjust the enabled TGs that are on the same channel as the first TG. This can be a bit confusing at first.

After a bit of use, I don't really like this continuous encoder assignment display.
Maybe it should only be displayed on shift. Its values ​​should be displayed on long-pressing shift.

soyersoyer:
News:
- Rebased on the latest main
- Use the latest Synth_Dexed

probonopd:
> Different question; the display of my pirate audio hat now shows "Minidexed" and the build number. Is this a testing setting? Can I Set a parameter to see the minidexed menu as usual instead?

That should show up only during boot time. Where and when do you see it?

GertSchepens:
PXL_20241229_215818196 MP

I only ever see the version on my pirate display. The output and buttons work, I'm playing sounds, everything is as expected, except the screen only ever shows the version. Whereas with the same config and the vanilla version, it showed the menu.


On the newest build.

Absolutely loving the arrows, the _ when at the end, and the text. Very elegant. Great update!!

I'm enjoying pressing shift to see what the knob values are.

I'm not so sure I love the by default showing the list of what does what. It looks very busy and jumbled. Though I get it's a lot of info and a very small screen. Do we really need that inélégant reminder of what's what?
And definitely. When I'm trying to change things in tg1, it reverts the text too soon.

Slightly related. I'm still thinking I'd enjoy the behaviour to be context sensitive to where in the menu I am.. on the root, I guess info is better than just showing "TG1" all the time.  But when I'm in TG2 already, I'm doing something there and I'd prefer the behaviour and inputs to support what I'm doing in that menu. I'm probably there to change things in that TG, so I'd prefer seeing that menu instead of the root info; and the inputs to change what I see on my screen and not all active TGs. (Instead of for example changing all voices of all active TGs.) When on the root, it does make sense to change all active ones.

GertSchepens:
After rereading the latest posts, I see there is context aware stuff 🎉 Sorry for missing that. Poking through a bit more.

However, the letter salad gets really heavy in the effects part; and while efficient, I was kinda surprised to be able to edit all in the first level effects menu.
I guess it's a challenge between expert accessibility to settings and simple UX.
I do enjoy that the rotator popup always clearly shows what's being edited, and that gave me a clear idea of what's happening even without that extra overview.

That said. Effects rotator 1 says compress, while 2 doesn't say Reverb, it confusingly says enable instead.

(Not editing the previous post to not risk messing up the photo)

probonopd:
@GertSchepens what do you see on your RPi screen after booting with the normal firmware (not the one from this PR for Arturia controllers), when no USB controller is connected to it at all?

soyersoyer:
> Absolutely loving the arrows, the _ when at the end, and the text. Very elegant.

This is the merit of Arturia, it's the only way to set it up :)
It will work similarly to the arpeggiator, where it shows at the bottom where you are in the menu, but I haven't decided yet how I will query the size of the menu.

>  I guess it's a challenge between expert accessibility to settings and simple UX.

I'm trying to put the encoders on as few pages as possible, maybe this is better than scattering them by type.

> A thought on on/off selectors.. right now it looks like on is 90+ and below, off. I think I'd enjoy the choices to be spread over the range.

I'll check it.

> It feels like it'd be cool if, when editing a voice, pad 1 to 6 would open OP1-6 for editing. It would tremendously speed things up, it seems..

That's a good idea, it can be solved.

> And of course, something more to do for the sliders too, while we're at it.

Is there a better idea than the volume of the first 4 active MIDI channels?:)
The first 4 OP Output levels?

What I also thought about was that there could be encoder pages where there are individual values ​​for the Operators, e.g. 6 output levels, or 6 detunes, etc.

News:
- shift button changes between the encoder display and the menu display,
- the menu display works as before
- you can change the encoder pages with the main knob, it's more convenient and hopefully more understandable
- added all ModWheel/FootController/AfterTouch/BreatController controls
- Handle MenuEventPressAndStepDown/MenuEventPressAndStepUp as the KY040 encoder (the encoder display can't work with it yet)
- Renamed Reverb encoders.

soyersoyer:
News:
- Long press shift to switch between menu display and the encoder/values displays
- short press of shift switches between encoder display and values display

probonopd:
At some point I'm going to need a proper MiniLab 3 👍

soyersoyer:
News:
- the on/off encoder center point is now 64
- the encoders are aware of the TG menu, so only TG1 values will be set in the TG1 Menu
- the main menu encoders use all enabled TGs, which channel is the same as the first active TG channel
- more encoder pages, each op parameter has a separate page with the 6 ops and one encoder that sets all the OPs.

soyersoyer:
News:
- Channel, Volume, Portamento Glissando, Poly/Mono added to encoder pages.

soyersoyer:
Keylab MkII display fix

Spacepadrille:
Hi there ! Today I am unable to download The MiniDexed_2025... I have this error message : "Failed to generate URL to download artifact." I tried on chrome and also on Brave, I was connected to my GitHub account... Any idea ?

Spacepadrille:
Hello, I am very impressed with this développement for the Arturia's keyboards. It is going fast, improvements every week or less, bravo ! 
I just want there report my experience with the MiniDexed/Soyer_mod and my Minilab3 ; I am a little bit lost with all the different pannels we can now display. Surely it is because I don't master MiniDexed's sub menues, and with some practice it will be fine. 
But I must admit I experienced a lot of glitches and delay in the display when I turn the knobs to kraft the sound. Sometimes the MiniDexed freeze and I have to unplug/replug it. Maybe my microSD card is not good enough, or my Rapsberry Pi 3 is to slow, I don't know. But it was not like that before the in depth menu implementation. 
An other thing : I use a lcd display on my MiniDexed, and this display is way bigger than the Minilab3's one. So I look to this screen instead of the minilab one while performing. And all the stuff with the menues and sub-menues is not displayed on this MiniDexed screen...
I wish you all a very good day.

probonopd:
@Spacepadrille thanks for testing. Can you give an example (click path) for a menu that is only available on one but not the other display? Thanks.

Spacepadrille:
> @Spacepadrille thanks for testing. Can you give an example (click path) for a menu that is only available on one but not the other display? Thanks.
 I don't know what is a click path :-), but I try to explain ; The two screens displays the same thing for general navigation, performance banks and preset names. But when you activate by a long press on shift the sub-menus, they are not displayed on the MiniDexed screen, only on the arturia one. From there, if I tweak a little bit to fast some knobs, the display freeze, but the actions are done, just the display have a delay after freezing. This make the whole thing difficult to manipulate, and I wonder if it is because of my hardware or not.

probonopd:
What is "long press on shift the sub-menu"? @soyersoyer do you understand why the two displays might not behave identically?

Spacepadrille:
@probonopd : lol my sentence was badly formulated. I mean activating the display of the sub menu by a long press on shift.

soyersoyer:
> But I must admit I experienced a lot of glitches and delay in the display when I turn the knobs to kraft the sound. 

Sorry, I didn't test a change, and it updated the colors of the pads too many times. That's what caused the glitches. If you experience this too, please let me know.

> What is "long press on shift the sub-menu"?

Holding down the Shift key will bring up an overlay menu where you can select control layouts for what the 2x4 encoders change.

The available overlay menus change depending on where you are in the menu. The Main, TG, Effect, Voice menus have different overlay layers. But all Voice settings are also routed to Main, so that on one page, for example, you can set OP_EG_R1 of the 6 voices (encoder 1-6) and all of them with encoder 8.

I think it's a good idea to have the menu appear on the default screen as well. However, it would have to be integrated into the MiniDexed menu and I don't know if others would think it's a good idea.

Spacepadrille:
Thank you for your hard work !

Spacepadrille:
`I think it's a good idea to have the overlay menu appear on the default screen as well. However, it would have to be integrated into the MiniDexed menu and I don't know if others would think it's a good idea.`

Maybe this can be "transparent", I mean not activated on minidexed without Arturia mod ? If you select "DAWControllerEnabled=1" you have the display of the sub menus on the MiniDexed screen, otherwise not. I'm not a coder, so maybe I tell sh** ?

soyersoyer:
I was thinking that it would be possible to set a MIDI CC for the shift button and 8 for the encoders, and it could work with any configurable controller that has 1 button and 8 encoders.

soyersoyer:

soyersoyer:
#767 couldn't make it to MiniDexed in 2 months. Maybe probonopd doesn't want to see any vendor specific stuff in it? I don't know.

#767 is a small change compared to this, so if this PR could be merged, I think it would take about 1 year to get there.

Since these output zips are only available for 3 days, I'd have to do something with the code every 3 days for the next year so people can try it out here.

This makes no sense.

There's also no point in asking people to try the change #800, but that's not available (for most people).

I don't see the reason for changing the retention days to 3 anymore, because it seems like actions from public github projects wouldn't count towards storage. https://github.com/probonopd/MiniDexed/commit/e9751e62411b49af6823c4eab0411a4c9a2ab143

The linked PR doesn't have a 3-day limit. https://github.com/soyersoyer/MiniDexed/pull/2

I'll always try to rebase on the main from here.

diyelectromusic:
> #767 couldn't make it to MiniDexed in 2 months. Maybe probonopd doesn't want to see any vendor specific stuff in it? I don't know.

Well, to be fair - just /one week/ ago there will still some updates required for this "simple" change.

The hazard of having a larger user-base now for code like this is that changing code and testing new functionality is relatively easy, but reviewing those changes and testing for no side effects for existing platforms (remember this supports Pi V1,2,3,4,5,Zero, Zero 2, 400) takes time.  People always come forward to test new functionality, especially when it meets an unmet need they had.  But very few people come forward and confirm that everything is still fine without side effects on existing platforms.

And you may recall it was I who was concerned about vendor specific code (https://github.com/probonopd/MiniDexed/pull/743#issuecomment-2458176786) not @probonopd who has only been very supportive of these more experimental changes.

But my main concern is about longevity of support for these large changes, hence wanting to understand the implications if asked for my opinions on them.

As I say, initial implementation is one thing, but it took quite a while to build up knowledge of the code base after the initial creators left, so when faced with a large body of changes (https://github.com/probonopd/MiniDexed/pull/743, https://github.com/probonopd/MiniDexed/pull/791, https://github.com/probonopd/MiniDexed/pull/790, https://github.com/probonopd/MiniDexed/pull/778, https://github.com/probonopd/MiniDexed/pull/782, https://github.com/probonopd/MiniDexed/pull/779, https://github.com/probonopd/MiniDexed/pull/777, https://github.com/probonopd/MiniDexed/pull/747, https://github.com/probonopd/MiniDexed/pull/783, https://github.com/probonopd/MiniDexed/pull/755, https://github.com/probonopd/MiniDexed/pull/750, https://github.com/probonopd/MiniDexed/pull/758), some of which didn't seem to be solving an obvious known issue, some of which were updating core libraries, and some of which were adding a significant amount of code, then we must collectively forgive the fact that volunteer maintainers (who also do this for fun) might take quite a while to catch up with what is going on.

Kevin

probonopd:
I am very sorry for the 3 days limit. I have been running out of space that GitHub grants. Still would very much like too see these contributions to be merged!

soyersoyer:
What do public github projects count towards?

probonopd:
Good question, I need to check. I just got a warning from GitHub toward end of December that I was over 90%.

soyersoyer:
News:
- Better KeyLab mkII support
- Fewer unnecessary display/encoder/pad color updates
- Uses less power

It turned out that it doesn't work well with the MiniLab 3 firmware v1.2.0. Please use version v1.1.1.

soyersoyer:
You can get a usage report.https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/viewing-your-github-actions-usage

soyersoyer:
News
 - Minilab 3 firmware v1.2.0 fixes

probonopd:
> You can get a usage report.

Not entirely sure how to read this, but my understanding is that across all my repositories, I only have 0.5 GB-months in the free tier but am at 0.79 already for January.

{8098E217-BC4F-456A-A58F-169EBF379FD1}

soyersoyer:
For me it's 0, even though I already have a few MiniDexed artifacts. I think this doesn't include public projects. If you request the usage report, you'll get a CSV where you can see how much each project is using.

soyersoyer:
Will the 3-day limit remain or will there be something else?

probonopd:
Maybe we can find some alternative service where one can temporarily upload files?
There used to be transfer.sh but it seems to be down...

soyersoyer:
How much space does MiniDexed take up based on the CSV?

probonopd:
I never managed to see the CSV. They don't give it for direct download but send it to some mail address (I don't even know which one, I seem to have too many).

soyersoyer:
https://github.com/settings/emails

soyersoyer:
And if you click on the "Get Usage Report" button, it will show you where it will send the report.

PR #744, opened 2024-11-07
Add network support

### LINKED ISSUE #43: Supporting RTP-MIDI
> [RTP-MIDI](https://en.wikipedia.org/wiki/RTP-MIDI) (also known as AppleMIDI) is a protocol to transport [MIDI](https://en.wikipedia.org/wiki/MIDI) messages within RTP ([Real-time Protocol](https://en.wikipedia.org/wiki/Real-time_Transport_Protocol)) packets over Ethernet and WiFi networks. It is completely open and free (no license is needed), and is compatible both with LAN and WAN application fields.

I imagine this could be quite useful to, e.g., send sysex from a computer to the Raspberry Pi and to have the Raspberry Pi driven by sequencers on a computer, phone, or tablet, especially since USB gadget mode is not available on Circle (although a Raspberry Pi PIco could probably be used to act as a USB MIDI to serial converter).

The [AppleMIDI library](https://github.com/lathoub/Arduino-AppleMidi-Library) contains C++ code for Arduino. It compiles on Arduino, MacOS (XCode) and Windows (MSVS).

I wonder how much effort it would be to port it to Circle.

In case the effort is prohibitive, we might want to use an ESP8266 as a WLAN-RTP-MIDI-to-Serial-MIDI bridge, but using wired Ethernet would be better to [reduce latency](https://github.com/lathoub/Arduino-AppleMIDI-Library/wiki/Keeping-Latency-under-control).

probonopd:
Closes #43
Thanks @omersiar


omersiar:
Tried to resolve the conflict from network branch.

Banana71:
Hello Omersiar,
I'm happy to read that there may soon be a network function for the miniDexed.
What should I test with version 2024_11_08_fc85a46? Have I missed any instructions?

Regards
Peter

omersiar:
Hi,

https://github.com/omersiar/MiniDexed/releases/tag/v0.0.1

Please also follow the instruction there. mt32's wiki has useful information too.

omersiar:
I'm sure you have pretty decent setup, so you can send midi notes for all your tracks to see if it can keep up, the implementation does not take into account the timestamps so notes can arrive late or early.

I mostly wanted to have it to test and tune sounds but full track playback should be the ultimate goal i think

probonopd:
According to https://github.com/rsta2/circle/tree/master/addon/wlan/sample/hello_wlan, this might work with the on-board WLAN device of the Raspberry Pi 3, 4, 5 and Zero W (not sure about the Zero 2).

A `/firmware` directory must be created on the SD card and firmware files for the WLAN modules need to be put in there. I'll try to do this as part of the image building so that users won't have to think about this.

Furthermore a configuration file `wpa_supplicant.conf` is required in the root
directory of the SD card. Please don't forget to edit your WLAN configuration in this file.

In `minidexed.ini` we need

```
[Network]
NetworkEnabled=1
NetworkDHCP=1
NetworkType=wifi
```

Most likely we should ship it by default as

```
[Network]
NetworkEnabled=0
NetworkDHCP=1
NetworkType=wifi
```

and instruct users to change `NetworkEnabled` to `1` if they want to use it.

probonopd:
Merged into a separate branch so that I can work on it; continues in https://github.com/probonopd/MiniDexed/pull/747.

omersiar:
/firmware files are propriety blobs which they may not allowed to redistribute freely I don't know.

https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/copyright

probonopd:
That's what I am using; let's continue in #747.

PR #747, opened 2024-11-09
Add network support by omersiar

probonopd:
image

Continuation from https://github.com/probonopd/MiniDexed/pull/744

This adds wired and wireless network for [supported models](https://github.com/probonopd/MiniDexed/wiki/Networking#ethernet-and-wlan-support):
* rtpMIDI a.k.a. "AppleMIDI" over WLAN
* UDP MIDI
* FTP over WLAN (e.g., to update without takin the microSD card out of the Pi)
* DNS-SD service discovery (a.k.a. "Bonjour", "Zeroconf")
 
Closes https://github.com/probonopd/MiniDexed/issues/43
Thanks @omersiar

This allows us to send MIDI e.g., from DAWs, Dexed on the PC, or even sysex for voices from sites like patches.fm

Documentation:
https://github.com/probonopd/MiniDexed/wiki/Networking#ethernet-and-wlan-support


probonopd:
https://github.com/dwhinham/mt32-pi/wiki/Networking%3A-RTP-MIDI-%28AppleMIDI%29 says

> mDNS/"Bonjour" for zero-configuration is not implemented, so mt32-pi will not be auto-discovered by your RTP-MIDI host software. You must add mt32-pi as a peer manually.

In the meantime, the awesome @rsta2 has implemented [this](https://github.com/rsta2/circle/discussions/427#discussioncomment-9693322). Do you think you could incorporate it?

* https://github.com/rsta2/circle/discussions/427

omersiar:
One can already connect to minidexed by its hostname if the router supports home network domain.

probonopd:
Yes, but that won't cause it to show up in music applications on the PC/Mac automagically.

omersiar:
Oh sorry I meant the circle as a submodule.  Currently minidexed do not checkout to required point for <circle/net/mdnspublisher.h>

probonopd:
Probably have to adjust

https://github.com/probonopd/MiniDexed/blob/main/submod.sh

so that the needed branch/commit is checked out?

omersiar:
Yeah this ⬆️, but changing this can break things. Let me check if latest circle [release](https://github.com/rsta2/circle/releases/tag/Step48) does break anything.

probonopd:
Testing on a Windows 11 PC:

After the usual

```
MiniDexed
Loading...
```

it says

```
Network      IP
192.168.0.122     >
```

(Minor nitpick: The `>` suggests there is a submenu while there isn't; I think we should leave the `>` away here)

Trying to connect via FTP from the FTP client built into Windows, I cannot get a connection using `ftp://minidexed` (which could well be my router's fault). When using `ftp://192.168.0.122` it asks for username and password, I used `admin` for both.

I do see a `SD` directory which contains the whole SD card contents. Cool! 👍 

Now I'll move on to testing the MIDI aspects...

omersiar:
It compiles with Step48 with these warnings:

```bash
net/ftpworker.cpp: In member function 'bool CFTPWorker::_ZN10CFTPWorker4PortEPKc.part.0(const char*)':
net/ftpworker.cpp:464:9: warning: 'char* strncpy(char*, const char*, size_t)' specified bound 512 equals destination size [-Wstringop-truncation]
  464 |  strncpy(Buffer, pArgs, sizeof(Buffer));
      |  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
net/ftpworker.cpp: In member function 'const TDirectoryListEntry* CFTPWorker::BuildDirectoryList(size_t&) const':
net/ftpworker.cpp:363:11: warning: 'char* strncpy(char*, const char*, size_t)' specified bound 6 equals destination size [-Wstringop-truncation]
  363 |    strncpy(VolumeName, VolumeNames[i], sizeof(VolumeName));
      |    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
minidexed.cpp: In member function 'void CMiniDexed::SetVoiceName(std::string, unsigned int)':
minidexed.cpp:1656:9: warning: 'char* strncpy(char*, const char*, size_t)' specified bound 10 equals destination size [-Wstringop-truncation]
 1656 |  strncpy(Name, VoiceName.c_str(),10);
      |  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
../Synth_Dexed/src/dexed.cpp: In member function 'void Dexed::setName(char*)':
../Synth_Dexed/src/dexed.cpp:1699:10: warning: 'char* strncpy(char*, const char*, size_t)' output may be truncated copying 10 bytes from a string of length 155 [-Wstringop-truncation]
 1699 |   strncpy(name, (char*)&data[DEXED_VOICE_OFFSET + DEXED_NAME], 10);
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

omersiar:
> (Minor nitpick: The > suggests there is a submenu while there isn't; I think we should leave the > away here)

I wanted to give impression that this is not stuck and move the encoder one step to see actual menu.  Dirty hack.
I am thinking to add a menu item for it.

probonopd:
I actually like that the IP address shows up automatically and doesn't go away until you press a button. Let's keep that!

Maybe we could do the following:
In the main menu

```
            MiniDexed
TG 1                >
```

we could show

```
        192.168.0.122
TG 1                >
```

instead, until the first button press. What do you think?

probonopd:
...RTP-MIDI on Windows 11 using loopMIDI and rtpMIDI works very nicely as described [here](https://github.com/dwhinham/mt32-pi/wiki/Networking%3A-RTP-MIDI-%28AppleMIDI%29#windows). Playing MIDI from REAPER to MiniDexed over WLAN 👍

probonopd:
> It compiles with Step48 with these warnings

Most likely this points to minor bugs that were there before but didn't trigger a warning so far?

probonopd:
Getting out my trusty old Mac mini to test it there...

probonopd:
Trying to update MiniDexed by FTPing over these files seems to crash the FTP client built into Windows 11. It doesn't even ask whether to overwrite the files, it just seems to hang.

image

Need to check with other FTP clients.

probonopd:
> Directory listing is now working

I can output MIDI to MiniDexed with Logic Pro 9.1.8 on Mac OS X 10.6.8; autodiscovery is working 👍

omersiar:
> Trying to update MiniDexed by FTPing over these files seems to crash the FTP client built into Windows 11. It doesn't even ask whether to overwrite the files, it just seems to hang.

For me it asks for individual file, did not try to upload more than one file.  [advised to use only 1 connection](https://github.com/dwhinham/mt32-pi/wiki/Embedded-FTP-server#supported-featureslimitations)

probonopd:
In rtpMidi, it shows up as "MiniDexed" now (like Apple, let's avoid too many technical words).

image

omersiar:
Also works mdns works briefly after booting, after awhile if it is not cached by the network peers, can not be able to discover minidexed.

https://github.com/rsta2/circle/discussions/427#discussioncomment-11196574

probonopd:
I had mine play some MIDI sent over WLAN from REAPER and after a while (within one hour or so) MiniDexed just rebooted. Might be a power supply issue, but worth watching out for.

Banana71:
FTP connection with Total Commander works.
Every now and then my miniDexed shows the IP address again. The Raspberry Pi4 is built into a metal case, could that be why the WLAN connection is not good enough?

However, the data transfer is not really bad at around 2 MB/s.

I am very happy that this function is finally being introduced into the miniDexed, so you have more flexibility when building the case and no longer necessarily need easy access to the SD card. Thank you @omersiar & @probonopd 

miniDexed FTP

probonopd:
@Banana71 can you send MIDI to it and it plays for longer than half an hour without issues?

Banana71:
> @Banana71 can you send MIDI to it and it plays for longer than half an hour without issues?

After installation I was able to control the miniDexed via WLAN, but then the miniDexed crashed and since then I have no longer been able to establish a connection via WLAN. 🤷‍♂️

omersiar:
Do you have heatsinks for the Pi4?

probonopd:
When mine freezes, power cycling makes it work again. This is on a Pi 3.

omersiar:
Without console logs it is hard to guess what is getting wrong.  Maybe we can put it to SD card so that can be read via FTP. 
https://github.com/rsta2/circle/issues/93#issuecomment-957345526 from this discussion I am thinking WLAN code can be improved further.

Also my RPi 3 was getting lightning icon during WLAN initialization that wasn't there before, so after switching to a new good USB micro cable and power source it got away.
BYW I noticed, during lightning icons (voltage brownouts), the red led indicator turns off for some reason

probonopd:
As it is currently implemented, using this will expose the WLAN password openly to anyone on the network (by being stored in clear text on a FTP server with a default password). While I don't think that a hobby synthesizer needs to be built like an enterprise server regarding security, I do think we should find some solution around that (e.g., hiding the `wpa_supplicant.conf` file in FTP). Possibly this should be taken care of in Circle so that everyone uses it could benefit? (And no, setting an FTP password would not improve the situation much as FTP is unencrypted.)

omersiar:
I wonder if ftpdaemon respects FAT file attributes, if then the files not needed to touched by users can be hidden

probonopd:
error

The red line appears when MiniDexed stalls after ~~40 minutes of playing the mp3 posted above over WLAN.

probonopd:
All that network logging should be optional, @omersiar do you think you could introduce a `VerbosityEnabled` option and only print out any network related messages if it is set to `1`?

https://github.com/probonopd/MiniDexed/wiki/Files#debug

We made all the other messages opt-in because it is known that printing messages to HDMI can cause audio issues on some RPi models.

probonopd:
Right now it seems to have a tendency for stuck notes (notes playing forever) when using an attached MIDI controller (keyboard) while networking is enabled. Maybe caused by the additional logging?

omersiar:
I see, stuck notes did not happen to me, tasks exceed can happen when to much new mdns task created. okay I may have idea for that.  Since your wifi connection is not reliable it creates a new task every time network reconnects. or maybe it is because of TTL setting.

probonopd:
MiniDexed being stuck already was the same before I shortened the TTL.

omersiar:
https://github.com/omersiar/MiniDexed/commit/3e77f72a7ab943edaf399f17e9569a9ac20e5ba3

This may help with the unreliable wlan connections, if my understanding from this discussion is correct > https://github.com/rsta2/circle/issues/93#issuecomment-957427536

omersiar:
> **Update 4:** Updating circle from fff3764 to c243194 may have introduced this issue, as #750 is also affected.

Yeah, this was predicted.  Unfortunately my low level c skills are not up to this extend.

omersiar:
> All that network logging should be optional, @omersiar do you think you could introduce a `VerbosityEnabled` option and only print out any network related messages if it is set to `1`?
> 
> https://github.com/probonopd/MiniDexed/wiki/Files#debug
> 
> We made all the other messages opt-in because it is known that printing messages to HDMI can cause audio issues on some RPi models.

Other than network disconnect and init messages, rest is coming from libs.

probonopd:
> https://github.com/omersiar/MiniDexed/commit/3e77f72a7ab943edaf399f17e9569a9ac20e5ba3
https://github.com/omersiar/MiniDexed/commit/17d905a1ae150d56a57c85486d5d87ea7a27d47f

I am having a hard time applying these cleanly, could you send a pull request based on the `probonopd/MiniDexed` branch `networking` please?

> Yeah, this was predicted.

Can you elaborate on this or point me to where this was discussed/predicted? Thanks!

omersiar:
> I am having a hard time applying these cleanly, could you send a pull request based on the probonopd/MiniDexed branch networking please?

Sure, I'll update you shortly

> Can you elaborate on this or point me to where this was discussed/predicted? Thanks!

https://github.com/probonopd/MiniDexed/pull/747#issuecomment-2466147464

> I wonder if ftpdaemon respects FAT file attributes, if then the files not needed to touched by users can be hidden

I checked this and it seems it is possible.   Let me try.

probonopd:
> Can you elaborate on this or point me to where this was discussed/predicted? Thanks!
> https://github.com/probonopd/MiniDexed/pull/747#issuecomment-2466147464

Indeed, the build from https://github.com/probonopd/MiniDexed/pull/747#issuecomment-2466136275 (before adding mDNS and updating Circle to c243194 for it) does not suffer from the stuck notes. So either the mDNS or updating Circle to c243194 likely causes the stuck notes. Since https://github.com/probonopd/MiniDexed/pull/750 is also affected, I tend to think that updating Circle to c243194 is likely causing the stuck notes. Maybe @rsta2 has an idea?

omersiar:
rebased downstream. now you should be able to merge.

Banana71:
> Build for testing:
> [MiniDexed_2024-11-09-eb96c5e](https://github.com/probonopd/MiniDexed/suites/30700804756/artifacts/2167171933)
> Use at your own risk.

Network not working, the version number of miniDexed is displayed, after a few seconds the display jumps to TG1 and crashes.
If I set ```NetworkEnabled=0``` in the minidexed.ini it works again.

Raspberry Pi 4, unfortunately I have no way to connect a monitor via HDMI

omersiar:
After reverting the changes applied to main branch, resolving ftp compiler warnings and handling wireless disconnections properly, I can say that my setup is now stable.

If I build rebased (main with sustain pr) branch, there are more than one issue which is hard to isolate.

probonopd:
[MiniDexed_2024-11-14-e9373ce](https://github.com/probonopd/MiniDexed/suites/30901680205/artifacts/2189115267) still has the stuck notes issue.

> After reverting the changes applied to main branch, resolving ftp compiler warnings and handling wireless disconnections properly, I can say that my setup is now stable.

Which circle commit/branch are you using @omersiar?

probonopd:
Has anyone tested wireless on a RPi Zero 2 W so far?

Banana71:
> Build for testing:
> [MiniDexed_2024-11-14-e9373ce](https://github.com/probonopd/MiniDexed/suites/30901680205/artifacts/2189115267)
> Has anyone tested wireless on a RPi Zero 2 W so far?

Networking doesn't work on my RPi Zero 2 W in gadget mode. After the version number, TG1        >  is displayed and the Rapsberry crashes.

probonopd:
This one seems to work for me on a RPi Zero 2 W 👍 

* I can send MIDI to it from my desktop using rtpMIDI
* I can transfer files via FTP using WinSCP
* I do not experience the "stuck notes" issue when playing glissando rapidly on an attached MIDI keyboard

Let's test the heck out of this!

omersiar:
Mine is also working straight for 2 hours, I disconnec

> _Sometimes this works 100% perfectly, other times it’s a fucking nightmare. Best of luck."_
> Kevin

Can confirm.  But it is great in MiniDexed context, which makes much easier to interact with it via browser or other tools.

Banana71:
RPi4: rtpMIDI connects quickly and works perfectly.
FTP works. The connection is terminated after about 20 seconds without a transfer, but can be reconnected immediately.

probonopd:
Wired network also tested positively.

probonopd:
I seem to be unable to use Ethernet UDP networking together with the MIDI Button Navigation functionality to simulate keys being pressed on the MiniDexed device (which would be neat for e.g., test automation):

```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import sys
import socket
import time

server_address = "192.168.0.171"

# MIDI message format
# MIDI CC

"""
minidexed.ini has

# MIDI Button Navigation
#  Specify MIDI CC to act as a button
#  NB: Off < 64 < ON
# CC channel: 0=OFF; 1-16 MIDI Ch; >16 Omni
MIDIButtonCh=17
MIDIButtonNotes=0
# Arrow left
MIDIButtonPrev=46
# Arrow right
MIDIButtonNext=47
# Arrow up
MIDIButtonBack=48
# Arrow down
MIDIButtonSelect=49
# Home button
MIDIButtonHome=50
MIDIButtonPgmUp=51
MIDIButtonPgmDown=52
MIDIButtonTGUp=53
MIDIButtonTGDown=54
"""

def playC3():
    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    message = b'\x90\x3C\x7F'
    print("Sending MIDI message: ", message)
    sock.sendto(message, (server_address, 1999))
    time.sleep(0.5)
    message = b'\x80\x3C\x00'
    print("Sending MIDI message: ", message)
    sock.sendto(message, (server_address, 1999))
    sock.close()

def arrowRight():
    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    # MIDI CC 47 value 127
    message = b'\xB0\x2F\x7F'
    print("Sending MIDI message: ", message)
    sock.sendto(message, (server_address, 1999))
    sock.close()

if __name__ == "__main__":
    arrowRight()
    time.sleep(0.1)
    playC3()
    sys.exit(0)
```

I hear the note play but I don't see the right arrow having any effect on the MiniDexed display.

probonopd:
@omersiar works with dns-sd with no stuck notes. Great work👍

probonopd:
Now multiple MiniDexed devices can be distinguished from each other by setting different hostnames.

probonopd:
[MiniDexed_2024-11-16-7198c9d](https://github.com/probonopd/MiniDexed/suites/30974682672/artifacts/2197220296) still shows `wpa_supplicant.conf` and allows it to be downloaded.

probonopd:
@omersiar do you have an idea how to block `wpa_supplicant.conf` from being downloaded?

probonopd:
Strangely, this works

```
			static const u8 SysLogServer[]   = {192, 168, 0, 143}; // FIXME: Don't hardcode this, use m_INetworkSyslogServerIPAddress instead
			static const u16 usServerPort    = 8514;		// standard port is 514
			CIPAddress ServerIP (SysLogServer);
			CString IPString;
			ServerIP.Format (&IPString);
			LOGNOTE ( "Sending log messages to %s:%u",
			(const char *) IPString, (unsigned) usServerPort);

			new CSysLogDaemon (m_pNet, ServerIP, usServerPort);
```

but this does not:

```
			static const CIPAddress ServerIP = m_pConfig->GetNetworkSyslogServerIPAddress().Get();
			static const u16 usServerPort    = 8514;		// standard port is 514
			CString IPString;
			ServerIP.Format (&IPString);
			LOGNOTE ( "Sending log messages to %s:%u",
			(const char *) IPString, (unsigned) usServerPort);

			new CSysLogDaemon (m_pNet, ServerIP, usServerPort);
```

What am I doing wrong?

omersiar:
The type for syslog SysLogServer is not matching what you define in config.  It is not IPAddress. it is char string so m_Properties.GetString would work.

omersiar:
> @omersiar do you have an idea how to block `wpa_supplicant.conf` from being downloaded? My attempts are not working so far.

Yes, found a dirty way

[ftpworker.cpp.patch](https://github.com/user-attachments/files/17791788/ftpworker.cpp.patch)


Command:	RETR wpa_supplicant.conf
Response:	553 Reading this file is not allowed

probonopd:
@omersiar seems like I had to do

https://github.com/probonopd/MiniDexed/blob/46ccb9ca12ae45bf60a93e0cfe6d4eb063fdfe28/src/config.cpp#L213-L217

as advised [here](https://github.com/rsta2/circle/discussions/506#discussioncomment-11282934). I wonder if we would need to do the same thing to the lines directly above it.

probonopd:
I'm pretty happy with this PR overall now.

In the log I see very frequent

````
06:31:23.512 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:23.513 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:24.502 >1 - 192.168.0.60 mdnspub - - - Send failed
06:31:31.512 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:31.516 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:32.507 >1 - 192.168.0.60 mdnspub - - - Send failed
06:31:39.517 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:39.520 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
06:31:40.520 >1 - 192.168.0.60 mdnspub - - - Send failed
```

The advertised service is recognized properly by the host computer.

Is this expected @omersiar?

omersiar:
It should be the other way around, if the syslog class expects a string then config should return a string.

probonopd:
Well, I'm not going to touch this anymore since it's working now :)

omersiar:
> I'm pretty happy with this PR overall now.
> 
> In the log I see very frequent
> 
> ```
> 06:31:23.512 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:23.513 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:24.502 >1 - 192.168.0.60 mdnspub - - - Send failed
> 06:31:31.512 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:31.516 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:32.507 >1 - 192.168.0.60 mdnspub - - - Send failed
> 06:31:39.517 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:39.520 >1 - 192.168.0.60 icmp - - - Time exceeded (0)
> 06:31:40.520 >1 - 192.168.0.60 mdnspub - - - Send failed
> ```
> 
> The advertised service is recognized properly by the host computer.
> 
> Is this expected @omersiar?

This is due if network connection is not reliable, don't you get also the network disconnected messages?

If that is not the case then network stack thinks it is connected but packages may not go through.

probonopd:
Thing is, it's playing MIDI coming in over rtpMIDI nicely, and I can upload firmware no problem. So for me it's purely cosmetic.

probonopd:
@diyelectromusic please review this if you have a chance. I'd like to get it merged before master deviates too much. Alone the possibility to output logs somewhere else than HDMI proves really useful.
Thanks!

diyelectromusic:
> @diyelectromusic please review this if you have a chance.

As I mentioned in the other PR there is far too much code for me to comment on the implications for the MiniDexed codebase.

I'm afraid I'm unable to advise on any of this change other than repeat my general concern - adding a range of scheduled tasks to handle network activity will, by definition, interfere with the audio digital signal processing at some point. I just don't know if it will be significant or not, especially if other Pi versions are factored in.

How has the networking scheduling scheme been designed (and why)?  How will it fit alongside the main ProcessSound loop and architecture of MiniDexed as it stands?  Has anyone looked at the process latency this introduces now?  If FTPing files over, what does run-time updating of the SD card do to the rest of the processing threads?  How will network timeouts be handled and will they suspend processing in odd places?  How is concurrency between menu updates and network updates handled (and how do we ensure there are no race-conditions or consistency problems)?  And so on...

I've not seen any discussion of how the networking stack has been architected into the existing code.  And I don't know enough about the architecture of MT32-Pi to know if a straight port of the same approach works here.

I've feel like I've managed to get to grips with the core codebase (pretty much), but there is no way I can try to get inside this code, so if you plan on accepting it, I think you need to think very carefully about who is going to maintain it and debug it (and the rest of the system now having to cope with it of course) if there are issues further down the line.

Kevin

probonopd:
Yes, I understand your concerns @diyelectromusic. I view this whole networking code as an opt-in feature that _should_, if switched off, have no impact. So that whenever an issue comes up, we can compare with networking switched off. And no, I don't advise on uploading files over FTP during a live performance ;-)

diyelectromusic:
Ok, so it would appear (according to the Scheduler documentation in Circle) that the non-pre-emptive, cooperative multitasking scheduler will only ever work on core 0 in a multi-core system...

So in that case, these networking tasks cannot interfere with the audio DSP side of things.

Any potential issues will be in things like UI handling and handling of non-networked MIDI, so it would be good to know if the USB and serial MIDI handling are affected by the networking; and then how the UI functions whilst the networking is running.

Aside:  I think this call to ->Yield() probably shouldn't be there - this is in the non-multicore (so Pi V1) area of code:
```
void CMiniDexed::Process (bool bPlugAndPlayUpdated)
{
	CScheduler* const pScheduler = CScheduler::Get();
#ifndef ARM_ALLOW_MULTI_CORE
	ProcessSound ();
	pScheduler->Yield();
#endif
```

I'm also not clear why there are so many calls to Yield in the CMiniDexed::Process() - e.g. in every part of the performance file handling.  Was this to fix some issues that have been found with the scheduling?

I really think this needs some developer documentation adding to the wiki to explain how the CTask and CScheduler facilities of circle are being used to support the networking.

I'd also like to see more checks/comments in the networking code to make it clear what is called when and to make sure things will drop out if things aren't properly initialised.  There seems to be a few assumptions about order and call sequences to ensure that the networking isn't enabled when not configured.  Should any of this sequencing change in the future (I don't know why it would, but things do shuffle around with time) then those assumptions would break.

Another Aside: I think printf() was used in mididevice.cpp as LOGNOTE doesn't always work in all contexts - e.g. if called from a serial port interrupt or something like that - I forget exactly... (although I'm not sure those printfs are particularly useful in there anymore anyway...)

I'm not sure I quite follow the logic in CMiniDexed::UpdateNetwork() - is that just performing the network startup? If so, why is it called for every scan of Process()?  I'm not following the interplay between bNetIsRunning, m_bNetworkInit, and m_bNetworkReady - can we make this a proper state machine if these are states the system passes through - and are these states appearing on separate calls to UpdateNetwork() or are they all expected to happen within the same call?

Are those LOGPANIC calls really the most appropriate response to a networking issue?  Also, response to various other bits of the networking failing seem inconsistent (there is no checking at all for m_UDPMIDI.Initialize() for example).

btw - should we be saying "applemidi" - wouldn't it be better to say rtpmidi... (and we'd not be building someone else's company name into the source of an open project...)?

There seem to be some build updates as part of these changes that don't seem specific to this change?

I've mentioned before that I'm really not a fan of on-the-fly patching of major source files.  If you really want to build the git hash into the firmware, can we put it in its own include file or something so we're not trying to patch userinterface.cpp?  Wouldn't need a s// then either - just echo it out to the file. (isn't "." a wildcard anyway?  So wouldn't this actually patch every occurrence of the string "Loading" and any following three characters no matter what they are?  Which could even include a newline and the start of the next line iirc...)

As I say, I've not been able to follow the networking code itself, I've just tried to think through any ramifications for the core MiniDexed functionality...

Kevin

probonopd:
I don't think synth_dexed has been ported to Linux. Might be a fun experiment for someone, though. If it is done in a way that keeps dependencies minimal (e.g., only ALSA, no jackd) it might run on a very stripped down "embedded Linux" type system containing not much more than the kernel, but allow for easier implementation of things like USB gadget mode as a sound device, MIDI device, and at the same time USB host for MIDI controllers, mass storage, etc. (I think) and networking, including ssh. Let me know if you ever start experiments in that direction.

probonopd:
So far I have tested this successfully on RPi Zero 2 W.

We should also test it on the other models:

- [ ] Zero (probably pointless since it has neither Ethernet nor WLAN)
- [ ] Zero W (WLAN only)
- [x ] Zero 2 W (WLAN only
- [ ] 1 (Ethernet only)
- [ ] 2 (Ethernet only)
- [ ] 3 (Ethernet and WLAN)
- [ ] 4 (Ethernet and WLAN)
- [ ] 5 (Ethernet and WLAN)

soyersoyer:
I've a 3b+ and 4 for MiniDexed. I'll check out this PR on those.

omersiar:
> If we already have wifi, ftpd, syslog, I think we should also have sshd :P Has anyone tried to use 8x synth_dexed on Linux? I think it could be interesting with PREEMPT_RT.

Not synth_dexed but 3x [VDX7](https://github.com/chiaccona/VDX7) is achievable on RPi 3A+, this is due to VDX7 being bit accurate emulation, also author made possible to run it in [headless](https://github.com/chiaccona/VDX7/issues/2) setup.

Minigoer:
Hello, thanks a lot for this update.
I tried several time to install MiniDexed_2024-11-19-5aad2ef on RPI 3 model b, but it freeze on the splash screen on the lcd.
I have no hdmi display to read debug message, and may miss something while doing the installation.

Banana71:
Hello, thanks a lot for this update.
I tried several time to install MiniDexed_2024-12-19-ccec4b6 on RPI 4 model b, but it freeze on the splash screen on the lcd.
I have no hdmi display to read debug message.

blixabloxa:
Hi All,

Thanks for working on this enhancement.

However, I too am stuck on the splash screen with Minidexed_2024-12-19-ccec4b6 on a Raspberry Pi 3 A+ wifi. I will try the previous version to see if that works.

Banana71:
Does anyone have a version after MiniDexed_2024-11-16-bfca612 working? https://github.com/probonopd/MiniDexed/pull/747#issuecomment-2480853368

Minigoer:
Hello, thanks for the MiniDexed_2024-11-16-bfca612 build, but unfortunately, it stuck also as MiniDexed_2024-11-19-5aad2ef on the splash screen (showing build number). I plugged an hdmi display (and edited the .ini for debug) but nothing is displayed.

probonopd:
* MiniDexed_2024-12-19-ccec4b6 on a Raspberry Pi Zero 2 W with `NetworkEnabled=1` configured to use WLAN and DHCP stalls at showing `MiniDexed 20241219-ccec4b6` for a long-ish time (around 10 seconds), but then eventually proceeds to showing the IP address and seems to be working for me. 
* When the configured WLAN is not available (e.g., because the router is shut off or the configuration in `wpa_supplicant.conf` is wrong), then it seems to stay on the `MiniDexed 20241219-ccec4b6` screen forever.
* If networking is enabled but the `wpa_supplicant.conf` file is missing, the system seems to just hang or reboot.
* With `NetworkEnabled=0` it boots into the menu without delay.
* Note to self: Retest with Ethernet and update this post accordingly.

So what needs to be changed? (Better ideas welcome!)

* Can it be changed so that it boots into the menu in any case, whether the WLAN can be reached or not, and give a clear message on the screen after a reasonable timeout (e.g., 20 seconds) if the configured WLAN cannot be reached?
* When WLAN is enabled but  `wpa_supplicant.conf` is missing, it should imho still boot into the menu, albeit without WLAN, or show a message on the screen.

omersiar:
Please see #783

Mopsteri:
> Does anyone have a version after MiniDexed_2024-11-16-bfca612 working? [#747 (comment)](https://github.com/probonopd/MiniDexed/pull/747#issuecomment-2480853368)

Hello!,
Tested with Raspberry 3B v1.2 / Zero2W
1. Download this version MiniDexed_2025-01-03-6800987 "Latest".
2. Install it to SDCard and make necessary modifications to files, eg. wpa_supplicant.conf, minidexed.ini and so on...
3. Now download this version MiniDexed_2025-01-03-0ec4b55.
4. Extract kernel8.img to the SDCard and overwrite existing file.
5. Boot MiniDexed.
It takes about 15 seconds to show IP-Address.

I know this is somekind "workaround" but hey! everything works just fine.

Tested with (rtpMIDI+Hatari+Twenty-Four 3.0 by Steinberg).

Br.
Marko

blixabloxa:
Hi All,

Just confirming that I followed Marko's procedure and that I got it working too with a RPi 3A+ wifi. Loaded Filezilla on Linux and FTP works to the SD card.

Enzo

omersiar:
@Mopsteri It is not a workaround actually, under the hood wifi chip tries to associate with the wireless access point which usually require some time which would be longer or shorter depending on how two devices are configured. Also built-in antenna is not the best available so keep it away from metal shielding and cables.

The difference now is Minidexed synth engine does not wait for network to be running so it is immediately available after boot.

omersiar:
> * Can it be changed so that it searches for the WLAN in the background without being stuck on the boot screen? In other words, boot to the MiniDexed main and only then try to (optionally) configure the network and just continue without the network if for any reason it cannot be reached.

It is now.

> * It should imho boots into the MiniDexed main screen **first** , without network. Imagine you have configured WLAN but then you are in a different location (traveling) in which your configured WLAN is not available. MiniDexed should still be usable (without WLAN) in that case. After some timeout, it should imho  just give up searching for the WLAN so that it doesn't continue to search all the time.

Network functionality now should not interfere with the synth engine

> * When WLAN is enabled but `wpa_supplicant.conf` is missing, it should imho still boot into the MiniDexed main screen, albeit without WLAN, or show a message on the screen.

It is now emitting some log (HDMI) and continue booting, wpa_supplicant.conf can easily be wrapped with a handler which then can be configured via menu in the future.

> * Wired Ethernet needs to be fixed. Of course there is no `wpa_supplicant.conf` in this case. _This worked in earlier builds, e.g. 20241116-973008d_.

I do not have any pi's to test wired network, so any feedback is welcomed.

probonopd:
My results with MiniDexed_2025-01-03-6800987:

* WLAN on RPi Zero 2 W does not work, last message on HDMI is "minidexed: WLAN initialized" (Picture 1)
* WLAN on RPi 3B does not work, last message on HDMI is "minidexed: WLAN initialized" (Picture 2)
* NOTE: I did not upgrade from any other version but just used the build from this PR directly
* Wired Ethernet works (also when there is no `wpa_supplicant.conf`) (Picure 3)
* In order not to run the risk of any trademark issues, I suggest to use the term "WLAN"/"wlan" instead of "wifi" everywhere (also in `minidexed.ini`)


Picture 1:

{F3B68F53-5930-47B4-89EB-1311BD6769B2}

Picture 2:

{B5431B81-26B1-48B6-AC4E-FCCAE9C7E43D}

Picture 3:

{33B84267-D16C-44B8-80E6-2DBA6C0F96CE}

probonopd:
@Mopsteri 

> Extract kernel8.img to the SDCard and overwrite existing file

This makes me curious which kernel you actually end up running. Does your installation still work when you move all files starting with `kernel*` EXCEPT `kernel8.img` into a subdirectory?

probonopd:
Closing in favor of https://github.com/probonopd/MiniDexed/pull/783. Dicussion continues there.

Issue #624, opened 2024-03-11
Sostenuto pedal does nothing


Kirtai:
I just tried the latest release with my dpiano and while the sustain pedal works the sostenuto one doesn't.

Any thoughts on adding support for this?

The MIDI CC# for sostenuto is 66.

diyelectromusic:
Ok, so I've had a bit of a look at this and I'm not sure this is implemented in the Synth_Dexed synth engine MiniDexed uses.

I've found references online to sostenuto being "key hold" which can be set to be controlled by the second foot controller (FC2) in a DX7ii:

![image](https://github.com/probonopd/MiniDexed/assets/68612569/424c4145-c988-4f38-9772-f8ff4c3b2b52)

But can't see where this maps into Dexed itself - I'm not sure it is part of the original DX7?  There are a number of API calls in Synth_Dexed to set what the foot controller will do for voices (choosing combinations of pitch, amplitude, EG and the range) and there is a global sustain function.  But I can't see a mention of sostenuto or key hold in the original DX7 manual and I'm not sure the original DX7 recognised CC66 either from the MIDI implementation chart I've found - again I'm wondering if this might be a DX7ii thing...

Kevin

Kirtai:
"Key Hold" sounds like the right thing though it's sad that it's not in the original DX7. :(

Kirtai:
Should I suggest this to the upstream Synth_Dexed or the original Dexed?

diyelectromusic:
Hmm, good question.  Not sure.  I guess ultimately it would have to be in Synth_Dexed regardless even if the original change is further upstream, so I'd probably start there.

I can see that each note has a "sustain" flag but that seems to always be set according to the voice's global sustain status.  I don't know how the two would interact so it might be necessary to introduce a "hold" flag for notes too to allow either sustain or hold to hang on...

I don't follow the Dexed code as much as Synth_Dexed, but I couldn't obviously see it implemented there either.  It certainly didn't look like it processed CC 66, only CC 64 - this is in PluginProcessor/processMidiMessage, which appears to have most of the logic that became dexed.cpp in Synth_Dexed.

Given how the code is quite diverged anyway, it could perhaps be added to Synth_Dexed and not worry about the original...
I don't know how much time Holger has to think about such things in Synth_Dexed these days though...

Kevin

soyersoyer:
https://codeberg.org/dcoredump/Synth_Dexed/pulls/7

probonopd:
[...]

PR #750, opened 2024-11-09
Update Synth_Dexed and support Sostenuto pedal

probonopd:
Hopefully closes https://github.com/probonopd/MiniDexed/issues/624

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ jobs:
     - name: Get specific commits of git submodules
       run: |
            sh -ex ./submod.sh
+           sed -i -e 's|int E;|int32_t E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9
     - name: Install toolchains
       run: |
            set -ex
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Get specific commits of git submodules
       run: |
            sh -ex ./submod.sh
-           sed -i -e 's|int E;|int32_t E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9
+           sed -i -e 's|int32_t E;|int E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9
     - name: Install toolchains
       run: |
            set -ex
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,10 @@ jobs:
     - name: Get specific commits of git submodules
       run: |
            sh -ex ./submod.sh
-           sed -i -e 's|int32_t E;|int E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9
+    - name: Apply patches
+      run: |
+           # Put git hash in startup message
+           sed -i "s/Loading.../$(date +%Y%m%d)-$(git rev-parse --short HEAD)/g" src/userinterface.cpp
     - name: Install toolchains
       run: |
            set -ex
--- a/src/effect_compressor.cpp
+++ b/src/effect_compressor.cpp
@@ -12,6 +12,7 @@
    MIT License.  use at your own risk.
 */
 
+#include <algorithm>
 #include <circle/logger.h>
 #include <cstdlib>
 #include "effect_compressor.h"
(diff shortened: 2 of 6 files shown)


probonopd:
Do you have an idea what is going wrong here @soyersoyer? Thanks for your help.

soyersoyer:
I think this one is the culprit:
https://codeberg.org/dcoredump/Synth_Dexed/commit/54e543143f2d0c89e6ba7254c8cfccc87b65c32a

Probably the int is 16 bit in teensy and they want  to use the 32 bit one? And on rpi2 the int32_t and the int is different type.
But the TEENSYDUINO doesn't use the compressor, so I think we can revert the types in that file.

soyersoyer:
https://codeberg.org/dcoredump/Synth_Dexed/pulls/9

probonopd:
Hmmm. The error slightly changed. Possibly this needs to be applied in more than one place? Does it build for you?

soyersoyer:
hmm, yes it worked for me.

sed -i -e 's|int E;|int32_t E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9
this should be:
sed -i -e 's|int32_t E;|int E;|g' Synth_Dexed/src/compressor.h # https://codeberg.org/dcoredump/Synth_Dexed/pulls/9

probonopd:
Stupid me!

probonopd:
@Kirtai does the above work for you?

soyersoyer:
https://github.com/soyersoyer/MiniDexed/commit/882d6c0bcd74b8e948ff327d7927c68b8befd458 this still needs
If you merge this, I can create a PR with this

probonopd:
@Kirtai please try with the version linked above - does it work for you?

diyelectromusic:
I would strongly recommend that we don't include changes that mean that the system can no longer be built using the build.sh and make in the src area.  Patching source code "on the fly" is a really bad idea in my view.

I know that change to build.yml is described as a temporary fix, but as I say having external dependencies outside of the main build/make system to my mind is storing up dependency problems for the future.  It will certainly make developing and building changes to the code messier and less obvious what is going on.

So, please can this not be merged in until that patching is not required.

Kevin

probonopd:
Note to self: only merge after the temporary fix can be removed once https://codeberg.org/dcoredump/Synth_Dexed/pulls/9 is merged.

We can nevertheless test and polish this PR in the meantime 👍

Kirtai:
> @Kirtai please try with the version linked above - does it work for you?

The initial sostenuto works but it stops as soon as I play another key and turns into sustain.

What happens.

1. Play a chord.
2. Press pedal.
3. Release chord.
4. Chord is correctly sustained.
5. Play another note.
6. Chord is silenced.
7. New note and following ones are sustained.
8. Release pedal, all sounds end.

What *should* happen.
1. Play a chord.
2. Press pedal.
3. Release chord.
4. Chord is correctly sustained.
5. Play another note.
6. Chord remains sustained.
7. New note and following ones are *not* sustained.
8. Release pedal, all sounds end.

Another problem is that pressing the pedal *then* playing a chord causes it to sustain. A sostenuto pedal is only supposed to sustain notes that are played when it is pressed.

probonopd:
So this would probably need to be addressed in [Synth_Dexed]([https://codeberg.org/dcoredump/Synth_Dexed](https://codeberg.org/dcoredump/Synth_Dexed/pulls/9))?

probonopd:
Right now it seems to have a tendency for stuck notes (notes playing forever) when using an attached MIDI controller (keyboard).

To reproduce, just play a glissando up and down (put finger on keyboard and slide left and right all over the range of the keyboard).

Might updating circle be causing this?

soyersoyer:
> The initial sostenuto works but it stops as soon as I play another key and turns into sustain.

Thank you for your invaluable review @Kirtai ! I did this based on the "hold" function of one of my controllers, which I thought was the same as sostenuto. By the way, that function is also fun, just not sostenuto. (And not all of my controllers know it.) I will update Synth_Dexed soon to make it work correctly.

dcoredump:
> Note to self: only merge after the temporary fix can be removed once https://codeberg.org/dcoredump/Synth_Dexed/pulls/9 is merged.
> 
> We can nevertheless test and polish this PR in the meantime 👍

Done

soyersoyer:
https://codeberg.org/dcoredump/Synth_Dexed/pulls/10

dcoredump:
> https://codeberg.org/dcoredump/Synth_Dexed/pulls/10

PR merged.

probonopd:
Updated Synth_Dexed. New build should appear here soon for testing. Thanks!

soyersoyer:
This would be good without the circle update.

probonopd:
https://github.com/probonopd/MiniDexed/pull/871 is meant to fix Sostenuto a.k.a. Hold. Please comment here if you think that something is still wrong. Thanks!

soyersoyer:
#878 is still needed. Sostenuto and Hold are different, but they work well.

probonopd:
Merged:

* #878 878

PR #761, opened 2024-11-16
Backport mdnspublisher from Circle

omersiar:
This re-adds mdns functionality back to MiniDexed while it is still on circle Step47.


omersiar:
Just realized there is no need to patch TTL

https://github.com/omersiar/MiniDexed/commit/ea338ff87154a274fb4b2fc29d13f5acd3f3264b

PR #768, opened 2024-11-28
Add hardware configuration for genXnoise devices

probonopd:
Add hardware configurations for
* [genXnoise dXeus machina | MiniDexed in eurorack format](https://www.genxnoise.com/product-page/dxeus-machina-minidexed-eurorack-format)
* [genXnoise MiniDexed MIDI Tone Module](https://www.genxnoise.com/product-page/minidexed-midi-tone-module)

(still to be tested)


Lonolo:
'USBGadget=1' is not appearing in the desktop_module.ini of this build.

probonopd:
According to https://github.com/probonopd/MiniDexed/wiki/Hardware#usb-gadget-mode

> In USB Gadget mode the Pi will obtain its power from the USB host it is plugged into and it MUST NOT be powered independently.

On the device as it is shipped, is there no way to inadvertently power it on another way than the USB host? (E.g., does the housing prevent from this?) Otherwise I'd be hesitant to enable this by default, in order not to have people who don't read the wiki brick their devices.

Lonolo:
The desktop module comes with a USB cable and no other PSU. Instructions explicitly say: 

> Configured as a _USB gadget_. Use the included USB cable and connect the micro side to the port labeled _USB_. Do not connect a power supply to the _PWR_ port.

I've considered plugging the PWR port with these, to solve that posiibility of user-error I think I will:
Screen Shot 2024-12-08 at 5 43 46 PM

But that is the simplicity of the desktop module, powered and MIDI controlled by the _USB_ port. That and an audio cable are the only two connections: 
IMG_3530

Lonolo:
Thanks again for your time spent on this, I understand the option to enable networking could also complicate including USBGadget in the default configuration. Perhaps adding a clear warning in the #comments above the USBGadget parameter would help caution users about potential risks. Users of both the desktop and dXeus modules already have flexibility with multiple connectivity options, including USB-MIDI, USB-Gadget, and soon, networking access, allowing them to tailor their setup to their needs.

probonopd:
The ini file for the desktop module now contains:

```
# CAUTION: To prevent hardware damage, DO NOT use the port labeled 'PWR'
# (the microUSB port near the edge of the device) when USBGadget is set to 1!
# You need to disable USBGadget if you would like to use that port!
# See https://github.com/probonopd/MiniDexed/wiki/Hardware#usb-gadget-mode for more information
USBGadget=1
```

That doesn't guarantee that anyone will ever read it there though...

Lonolo:
Thanks - works great!

These will now ship with a plug in the _PWR_ port:

desktopPlug

soyersoyer:
`MIDIThru=ttyS1,ttyS1` looks wrong. Is it?

Lonolo:
> `MIDIThru=ttyS1,ttyS1` looks wrong. Is it?

What looks wrong?

Using the RPI's Tx/Rx pins to a midi DIN circuit. MIDI chaining them all and works perfectly:

dXeus-racked

soyersoyer:
Shouldn't it forward MIDI IN to the THRU port?

I assume that ttyS0 is the IN port and ttyS1 is the THRU port.
And with this configuration, it forwards ttyS1 to ttyS1.

Issue #773, opened 2024-12-12
Manipulating Master Volume makes System stop playing sounds. E-MU Xboard 25


RichyT:
System is working well and playing sounds but moving the master volume at all and everything just stops.

Powering off/replugging the keyboard doesn't seem to work at all, only a full reboot. None of the other knobs or the pitch-bend have a similar effect, just the master volume as far as I can tell.

I removed the usbspeed=full from cmdline.txt to see if that was the issue but the problem persists.

I don't have a rotary encoder/LCD to test anything with that.

RichyT:
Sure. The picture is not great as it's on a low-res display and I had all sorts of issues with glare. Let me know if you need more.

![midi](https://github.com/user-attachments/assets/02ec0ccd-30dc-48a2-bf1c-dea11944ddc8)

diyelectromusic:
Hmm.  I've not really done much with master volume... I am intrigued by some of the code though (and have mentioned it before: https://github.com/probonopd/MiniDexed/issues/502).

Any idea what that stream of what looks like B0 5D xx messages before it relates to?  That looks like continually sending MIDI CC message for controller 93...

I'll try to take a look and see if I can simulate something that reproduces the issue here...

Kevin

RichyT:
For this run, I just did played some random notes to ensure it was working then moved the slider a little to cause the issue. I am not sure what the other numbers are as playing the notes does not seem to correlate with values being displayed though I am not sure if there is some buffering going on. (Turning the knobs does produce an obvious correlation).

I'm not really concerned about getting the master volume working as much as having it not cause an unrecoverable error. 

I do notice that the LED sometimes displays CC along with some numbers but I'm not sure if that correlates with what you mention. I can look harder and see if it does.

diyelectromusic:
Ok.  Anyone spot the mistake here?
```
		float32_t nMasterVolume=((pMessage[5] & 0x7c) & ((pMessage[6] & 0x7c) <<7))/(1<<14);
```

It turns out that if you take the logical AND of a LSB and a MSB you get... guess what?  zero :)

So on the first SysEx master volume, it zeros everything and can never recover.

Also - no idea why that is taking 0x7C - that is grabbing bits 0b01111100 from each byte...

Kevin

diyelectromusic:
Ok, give that PR a go - hopefully that now handles Master Volume correctly!

Kevin

PR #775, opened 2024-12-13
Fix Device Master Volume Control handling

### LINKED ISSUE #502: SysEx Issues: Master Volume and Device ID
The master volume SysEx System Common message is only valid if used with the special manufacturers ID of 0x7F, so this should be included in the test in mididevice.cpp (https://github.com/probonopd/MiniDexed/blob/main/src/mididevice.cpp#L178).

Also, in the TG specific SysEx handling later on, there is no confirmation that the SysEx message is from Yamaha, although this happens within Synth_Dexed itself so is probably ok.

I am also now wondering about the device ID in the SysEx messages.

The device ID is assumed to be a MIDI channel, so is checked against the MIDI channel for the TG.  This is probably an assumption as Device IDs are meant to be something different I think?  I don't know how the TX816 uses them, but it might make more sense to check the Device ID against the TG number itself rather than MIDI channel?  But it probably depends on how a DX editor would use Device ID (Dexed itself only seems to have the concept of "sysex channel")?

The DX7II Advanced MIDI Application note says this about device IDs:
image

I've not found any mention of them for the original DX7 so far (did that even have them in the UI?).

Device IDs also appear in the Universal System Exclusive Messages (0x7F messages) too and it seems other manufacturers often have a device ID after the manufacturer's ID too, but I guess that isn't mandatory.  Here is what the MIDI spec says about Universal SysEx Device IDs:
image

So is assuming it is the same as MIDI channel the simplest way to go for someone using MiniDexed?  Maybe it is... but we perhaps ought to have a note about it somewhere and it might make exchanging data with a real DX7II problematic if the UI on the real synth is used to set Device ID to something else (does it default to 1?)

Kevin

diyelectromusic:
And also comment out low-level MIDI dump code that really should only be there if someone is messing around with the MIDI handler itself.
-- actually no, that is already commented out in the serial driver.  But this is very unresponsive due to line-buffer flushing, etc, so switched to fprintf to stderr instead.

This is a fix for Issue #773 and partly fixes #502 - at least the Master Volume Control part - the message is now explicitly checking all SysEx bytes and the message length to ensure it really is a MIDI SysEx realtime device control message for the Master Volume Control.

Kevin

### DIFF
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -136,6 +136,8 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 	// The packet contents are just normal MIDI data - see
 	// https://www.midi.org/specifications/item/table-1-summary-of-midi-message
 
+	// Additional byte-level MIDI dump for extra debugging
+/*
 	if (m_pConfig->GetMIDIDumpEnabled ())
 	{
 		switch (nLength)
@@ -177,6 +179,7 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 			break;
 		}
 	}
+*/
 
 	// Only for debugging:
 /*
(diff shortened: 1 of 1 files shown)


probonopd:
@richyT can you please test the build above and let us know whether it fixes the issue for you? Thanks.

RichyT:
Sorry, no sound at all with that one. Verified that I changed the output to hdmi too.

RichyT:
OK, turns out the usbspeed thing _was_ needed. Works good now, including full volume control while playing.

PR #779, opened 2024-12-30
pass strings by const reference and use find_last_not_of

soyersoyer:
const reference opportunities were advised by cppcheck to skip copy construction

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1968,9 +1968,9 @@ std::string CMiniDexed::GetNewPerformanceDefaultName(void)
 	return m_PerformanceConfig.GetNewPerformanceDefaultName();
 }
 
-void CMiniDexed::SetNewPerformanceName(std::string nName)
+void CMiniDexed::SetNewPerformanceName(const std::string &Name)
 {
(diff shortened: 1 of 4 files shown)


probonopd:
If I understand it right:

*   It improves the efficiency of string handling by using `const std::string &` instead of `std::string` as function parameters. This avoids unnecessary string copies and reduces memory allocation.
*   It simplifies the code in `CPerformanceConfig::SetNewPerformanceName` by using the `find_last_not_of` method to remove trailing spaces from the input string. This replaces the original loop that iterated over the string to find the last non-space character.
*   The changes are consistent across multiple files (`minidexed.cpp`, `minidexed.h`, `performanceconfig.cpp`, and `performanceconfig.h`), ensuring that the improvements are applied uniformly throughout the codebase.

Thanks @soyersoyer. Were you able to test it thoroughly so it's low risk to merge?

soyersoyer:
Yes.

diyelectromusic (approved):
Remember I'm not a C++ expert, but from what I think I know of references, I can't see any issues with this.

PR #786, opened 2025-01-07
Add hardware configuration for Serdaco MP32L device

probonopd:
https://github.com/probonopd/MiniDexed/discussions/785#discussioncomment-11765623

@BobanSpasic it would be great if you could test the configuration in this build. 

Need to replace the default `minidexed.ini` in the root of the microSD card with the appropriate one from inside the `hardware` folder.

PR #787, opened 2025-01-08
Add hardware configuration for mt32-Pi-Midi-Hat device

probonopd:
Based on https://github.com/probonopd/MiniDexed/pull/786#issuecomment-2576758227

@logomancer can you please give this build a try and please tell me the exact version of your board (it is printed on the PCB). Thanks!


logomancer:
The configuration works. I am using version 1.7.1 of the board. I should point out that the board is designed to be assembled with either the 128x32 or the 128x64 display, so people may have to do some configuration regardless. But at least the display is in the correct orientation now.

PR #790, opened 2025-01-09
Always show both arrows in the bank selector menu

soyersoyer:
It wraps around.


soyersoyer:
What about this? It's really a simple change.

soyersoyer:
What about this? It's really a simple change.

soyersoyer:

PR #793, opened 2025-01-12
Fx updated by @jnonis

probonopd:
Continuation of https://github.com/probonopd/MiniDexed/pull/765


probonopd:
Hi @jnonis, happy new year!

I really think FX would be a great addition to MiniDexed. But I can also understand @diyelectromusic's concerns that we add a lot of code here which may or may not interfere with the performance of MiniDexed overall.

So I'd like to have, in minidexed.ini, a "master switch" with which all FX related code can be enabled or disabled, `EffectsEnabled=1`. So that whenever we run into some (seemingly unrelated) issue, we can switch off the whole FX code path. Do you think this is possible, or is it even implemented this way?

Also, coming back to https://github.com/probonopd/MiniDexed/pull/765#issuecomment-2491643223, FX should be entirely disabled when MiniDexed is in  8-output mode.

Do you think you could make these changes?

Also I notice the PR contains changes regarding `m_nButtonsDebounceTime`, what's up with these? If they are unrelated to FX, they'd better be handled separately.

@diyelectromusic do you have other requirements/considerations to make this ready for being merged (after intensive testing)?

probonopd:
Notes to self (to be compared with main):
- [ ] On Raspberry Pi Zero 2 W, I have to replug my Keystation Mini32 MK3 for it to work

Questions:
* Can there be more than 1 "Insert FX"? (e.g., "Insert FX 1"... "Insert FX 4")
* Would it be useful to have "Global FX" that affect all TGs at once as well? Like "Insert FX" but for all TGs? Or a quick way in the menu to set all TGs to the same "Insert FX" (which then can be fine-tuned by going into each TG individually)?
* Can we make all of this controllable via MIDI? In a way similar to how the MPX90 is controllable over MIDI?

Banana71:
Unfortunately, this miniDexed version does not start on my Raspberry Pi 4. :worried: The display freezes after the version number is displayed. So I guess I have to play the sound like this first. :trollface:
TX816-Symphonic

Banana71:
This version runs on my Raspberry Pi 3 A+. I sent the E. Piano from the TX816 to the A Phaser and assigned the insert FX of the 8 TGs to the TalRvrb3. After a short time the sound starts to crackle. My Raspberry Pi 3 has no additional heat sinks. How well does it work for you?
[000118_TX816 EP FX.zip](https://github.com/user-attachments/files/18390062/000118_TX816.EP.FX.zip)

probonopd:
[000118_TX816 EP FX.zip](https://github.com/user-attachments/files/18390062/000118_TX816.EP.FX.zip) plays well for me with no crackle even after 2 minutes on the Raspberry Pi Zero 2 W. Maybe a power supply or similar issue?

jnonis:
> Hi @jnonis, happy new year!
> 
> I really think FX would be a great addition to MiniDexed. But I can also understand @diyelectromusic's concerns that we add a lot of code here which may or may not interfere with the performance of MiniDexed overall.
> 
> So I'd like to have, in minidexed.ini, a "master switch" with which all FX related code can be enabled or disabled, `FXEnabled=1`. So that whenever we run into some (seemingly unrelated) issue, we can switch off the whole FX code path. Do you think this is possible, or is it even implemented this way?
> 
I think it is possible to add feature switch to disable all the FX.
Should this switch also includes current Reverb on MInidexed?

> Also, coming back to [#765 (comment)](https://github.com/probonopd/MiniDexed/pull/765#issuecomment-2491643223), FX should be entirely disabled when MiniDexed is in 8-output mode.
> 
> Do you think you could make these changes?
> 
I think that FX are not processed in case of 8-Output mode, I will double check. I made this when updated the branch.

> Also I notice the PR contains changes regarding `m_nButtonsDebounceTime`, what's up with these? If they are unrelated to FX, they'd better be handled separately.
> 
I will remove that change from fx branch, it is something that I need for one of my builds, I was having a lot of bouncing with buttons.
PXL_20250112_163735268

> @jnonis do you plan other changes before you'd consider this ready for integration?
> 
Yes, I have another branch :)
This branch includes a second send fx and a master fx. But it is pushing the limits of the RPi 1, so.. I had to add some changes to improve the performance. I noticed a lot of glitches and noise when I dive in the menus, they are related with the LCD display update.
https://github.com/jnonis/MiniDexed/tree/fx-full

probonopd:
Know what, I'd say disable FX on RPi altogether.

For me RPi1 = 1 DX7. Mono. :)

jnonis:
@probonopd You have done a great work creating this project.
Minidexed is simple to build, work on all RPis, and sounds great.

About your questions:

> * Can there be more than 1 "Insert FX" (e.g., "Insert FX 1"... "Insert FX 4")

I was think into have FX combinations, by example Distortion+Chorus instead of adding more Insert FX Slots. That allows us to create combinations that actually works based on the processing load that each effect adds to the chain.

> * Likewise, can there be more than 1 "Send FX" (e.g., "Send FX 1"... "Send FX 4")?

I'm testing other branch with 2 Send FXs and a Master FX, It struggle a bit on the RPi 1, but it works.
https://github.com/jnonis/MiniDexed/tree/fx-full

> * Does it make sense to have all FX types available both as "Insert FX" and "Send FX"?

I think it doesn't but at this moment all the implementation was more a research and it was easy to have all the FX available to do some tests.

> * Can we make all of this controllable via MIDI? In a way similar to how the MPX90 is controllable over MIDI?
> 

I guess it is possible with some limitations, the main issue are the amount FX parameters which are different for each FX. I need to think about how the Midi CC will be assigned to the FX parameters.

Something I would like to add the the FX is Preset Support.
- Fist version can be  just a hardcoded list of presets
- Then we can add file support to Load and Save presets for each FX.

jnonis:
> Unfortunately, this miniDexed version does not start on my Raspberry Pi 4. 😟 The display freezes after the version number is displayed.

I have a RPi 4 but I didn't test Minidexed on it, I will test it.

probonopd:
So... let me test 
* https://github.com/probonopd/MiniDexed/pull/795

Issue #563, opened 2023-11-19
Reverb time depends on the sample rate


Banana71:
I wanted to improve the latency of the miniDexed and experimented with the SampleRate and ChunkSize in the miniDexed.ini. I noticed that the ReverbTime behaves anti-proportionally to the SampleRate.

Is it possible to change the programming so that the reverb time is independent of the sample rate or should we just document the behavior?

Why all this: I noticed an audible latency between the miniDexed and the Roland RD-300 for sounds with a short attack time. With the following values, the latency can no longer be determined and even a Raspberry Pi 2 runs stable:
`SampleRate=48000` (default)
`ChunkSize=128` (default 256)

With a sample rate of 96000, the latency is measurably better but the reverb time is halved.

probonopd:
Good observation! In other words, you are saying that we need to adjut the reverb time like this?

```
AdjustedReverbTime = ReverbTime / (DefaultSampleRate / SampleRate)
```

probonopd:
@Banana71 can you confirm the above? Then we could go about implementing it.

Banana71:
```
AdjustedReverbTime = DesiredReverbTime * correctionFactor

correctionFactor = SampleRate / DefaultSampleRate
```

The formula describes the correct solution approach.

But the code for the reverb effect already works with a sample rate. Is the code in the effect incorrect? Are the variables not assigned correctly? What sample rate is mentioned in the code for the effect?


search for Samplerate:
![Samplerate](https://github.com/probonopd/MiniDexed/assets/104296205/f005cf17-4e6f-4237-bdcb-742c56dddb9d)

Banana71:
Thanks alot @probonopd.  🥇

Banana71:
I was getting too excited.

I've now reloaded the version MiniDexed_2025-04-18-b15ffad. At sample rates < 48000, the reverb time gets shorter. Now the behavior is exactly the opposite.

At a sample rate of 96000, miniDexed crashes.

When I first tested it, I was probably just listening to see if the reverb time increased at 44,100 Hz. The difference between 44,100 and 48,000 is small. I hadn't noticed that then.

PR #795, opened 2025-01-12
Fx full by @jnonis

probonopd:


probonopd:
Working well so far on my Raspberry Pi Zero 2 W.

Banana71:
> Build for testing:
> [MiniDexed_2025-01-12-7175505](https://github.com/probonopd/MiniDexed/suites/32963937526/artifacts/2419139796)

Testet with my Raspberry Pi 3 Model A+
- If I set the sample rate to 22000Hz the sound output works reliably.
- The SSD1306 display seems slower, the display is no longer built up as quickly
- The "Save as default" command does not work correctly. The first performance of the current bank is overwritten. (Probably a bug that applies to all versions)
- The sample rate set in the minidexed.ini affects the reverb time of the standard Reverb effect. Higher sample rates mean shorter reverb times and correspondingly lower sample rates mean longer reverb times.
- The DAC crashes during many performances. When loading a performance, there is a loud cracking sound and the audio output can only be revived by restarting. I have not yet found a common cause for this.

probonopd:
Thanks for your testing @Banana71, this is really a big help.

> If I set the sample rate to 22000Hz the sound output works reliably.

So far I have not tested other sample rates. Do you know sample rates that work with the normal build but not with this one?

> The SSD1306 display seems slower, the display is no longer built up as quickly

I noted the same with a 16x2 character display. I wonder what additional processing is going on even when no sound is playing?

> The "Save as default" command does not work correctly. The first performance of the current bank is overwritten. (Probably a bug that applies to all versions)

Can you please retest this with the normal version and open a bug if it is there too?

> The sample rate set in the minidexed.ini affects the reverb time of the standard Reverb effect. Higher sample rates mean shorter reverb times and correspondingly lower sample rates mean longer reverb times.

Is this only the case for the (old) standard Reverb effect or also for the new ones? (Should we even keep the old one once we have the new ones?)

> The DAC crashes during many performances. When loading a performance, there is a loud cracking sound and the audio output can only be revived by restarting. I have not yet found a common cause for this.

Can you please give a few examples so that we can replicate?

probonopd:
Noticed the following:

When I switch on MiniDexed, play a bit, and then just let it sit for a while (e.g., half an hour to an hour) and then try to play again, it stays silent. (Not sure whether this is a hardware issue and/or whether it also happens with the normal MiniDexed version. Just something to watch out for.)

Banana71:
> So far I have not tested other sample rates. Do you know sample rates that work with the normal build but not with this one?

A sample rate of 48000 and the use of effects on all TGs causes crackling in the sound output. I haven't tested other sample rates up to 22000. I haven't tried fewer TGs either.
##


> The "Save as default" command does not work correctly. The first performance of the current bank is overwritten. (Probably a bug that applies to all versions)
> Can you please retest this with the normal version and open a bug if it is there too?

The main version is also affected.
##
I only noticed the dependence of the reverb time on the sample rate with the old standard reverb.
##
e.g. the DAC crashes with the following performances: 001:041 Barock, 001:065 1984, 002:001 mDx-Strg.AT (which is why I can't test the Greenhouse bank)
##
The feedback parameter for the delay effect is incorrect. If the feedback value is set to zero, the delay should work like an echo. Unfortunately, there is then no echo at all. I hope you understand what I mean

probonopd:
> I only noticed the dependence of the reverb time on the sample rate with the old standard reverb.

Ah right, that's a known issue of the old reverb, so unrelated to this change.

* https://github.com/probonopd/MiniDexed/issues/563

Banana71:
The effects chain has been changed more often now and I think it has gotten better and better. But as I would like it to be.

In each TG the following: Send FX 1, Send FX 2, EQ
under Effects: FX1, FX2, Compressor, The parameters of the compressor should be changeable (threshold, ratio, input, output, gain or whatever the compressor can do). The compressor should always be at the last position in the chain.

Banana71:
It would be sad if nothing was done at this point.

probonopd:
@jnonis: Please let us know when this is function-wise ready for being merged from your perspective.

We then all will need to test it on all RPi models we have access to.

I would be fine with the entire FX code path being disabled on RPi 1 and RPi Zero (but not RPi Zero 2).

jnonis:
Hi @probonopd,
I won't me able to work on this next 2 weeks.
At this point I have 3 things that needs work:

- Configuration property to disable all The FXs
- Remove or make optional the chages to get performance for single core RPis, this makes the display slower but give priority to sound processing.
- Test in RPi 4

I think I can do this in a weekend when I will be back.

jnonis:
@probonopd

I've added a new property **FXEnabled** to disable all the FXs.

I've reverted 2 changes non related with FX, I will send a PR for each one.
- Configurable debounce time for buttons.
- Avoid to send to much updates to display, this improve the performance for single core RPis, but also can make the UI smoother for the rest of RPis (with the proper configuration)

I found the problem with RPi 4/5, it is related with the configurable amount of tone generators, there were several places in the code where I was using CConfig:AllToneGenerators constant instead of use m_nToneGenerators member attribute.
I've fixed the problem partially, the last build should work but if you try to edit an invalid tone generator the application will crash.

probonopd:
Hi @jnonis, I'd like to get this merged soon.

> I've added a new property FXEnabled to disable all the FXs.

How does this work? When I set `FXEnabled=0` in `minidexed.ini`, I still see `#120` and all the other FX-related menu entries. Can you make it so that when it is set to `0`, none of the FX-related codepaths run?

> I've fixed the problem partially, the last build should work but if you try to edit an invalid tone generator the application will crash.

Is this something you think you can figure out, or would you need some help on specific parts of the existing code (which ones)?

Thanks a lot!

jnonis:
Hi @probonopd 

> > I've added a new property FXEnabled to disable all the FXs.
> 
> How does this work? When I set `FXEnabled=0` in `minidexed.ini`, I still see `#120` and all the other FX-related menu entries. Can you make it so that when it is set to `0`, none of the FX-related codepath runs, and no FX-related menu entries show up?

I've made some changes to show the FXs disabled in de UI.
Now the FXs are disabled by default, they need to be enabled with  `FXEnabled=1`.

> > I've fixed the problem partially, the last build should work but if you try to edit an invalid tone generator the application will crash.
> 
> Is this what [4bc87fd](https://github.com/probonopd/MiniDexed/pull/795/commits/4bc87fd7773ba534e6535ce8a61dfb2b5f1df67d) fixes?
> 
Yes, it is fixed in that commit.

probonopd:
Thanks a lot @jnonis. Even with `FXEnabled=0` I see menu entries e.g., for "Send FX 1", even though its only option is "Disabled". Wouldn't it be better to not show any FX-related menu entries when `FXEnabled` is not set to `1`?

jnonis:
@probonopd 
The same happens with the disabled TGs, if set 8 TGs in a RPi4, you see the 16 TGs in the UI menu.
The menu implementation is mostly static so, to remove items from the menu needs to reimplement MenuHandler method, and how the menu is declare.
What I am trying to say is that looks like a major refactor.

probonopd:
OK, thanks @jnonis, in this case it's probably not worth the hassle.

Do you think it's otherwise ready for being merged?

jnonis:
I think it is ready to be merged

probonopd:
@diyelectromusic wdyt? Do you have any objections?

diyelectromusic:
This is an awful lot of code to include, so its a little hard to say.

A couple of things I have noticed that I'll mention:
* A lot of the files don't have headers, is that an issue?
* mididevice.cpp is no longer ignoring 1 byte messages, so I don't know what that will do performance wise if used with a device that does Active Sensing for example, as a lot more code appears to run every time now.
* In minidexed.cpp I see the channels has been changed to 2 by default but previous it was only 2 when using a multi-core Pi.
* In MiniDexed::Run for cores 2 and 3 I notice that there are now additional spinlocks being acquired to support MidiArp either side of the GetSamples call - I've no idea if this will affect audio processing performance in the multi-core system.
* In MiniDexed::ProcessSound there are a lot of changes to the audio processing that I'm afraid I don't understand (it took me long enough to work out the previous behaviour I described here: https://github.com/probonopd/MiniDexed/discussions/818#discussioncomment-12405615).
* In particular there is now a stereo effects chain that seems to happen for non-multicore Pis, whereas previously, for performance reasons as I understand things, non-multicore Pis had no effects, had lower polyphony and only a mono output.
* Also, like in Run there are now more spinlocks being processing around the getSamples, so the same concern applies here.
* I'm guessing this is a "breaking change" for all the existing performances too (at least, I see the Reverb settings appear to have gone)?
* uibuttons I see the debounce time has increased, but doesn't way why (but I guess is performance related)?
* uimenu.cpp has changed the main menu handling (switching from MenuHandler to a new MainMenuHandler) in a way I don't quite understand.  I can't really comment on the other uimenu changes as the comparison gets out of sync with all the changes.
* userinterface.cpp has changed how the rotary encoder is handled, and it isn't clear to me why.  Is this performance related again?

In short, this seems to be a fundamental change from effects being something that happens during TG mixing to produce the final audio, to being something that is now on a per TG basis and I don't know what that will mean for the system.

And the changes to the rest of the code seem quite significant to me, and I don't really feel qualified to be able to comment on the implications, but I'm certainly nervous that this might affect performance on a non-multicore (V1) Pi.

Kevin

Banana71:
Please don't merge.

- I can't save a performance with my RPi 4. The little Pi crashes immediately.
- The names of some parameters are not consistent.
- The functions of some effects don't work.
- The flanger doesn't do what a flanger should do at all.
- The parameter feedback of the delay effect still doesn't work properly.
- The menu navigation has been changed, why?
- and a lot more...

I've only tested it for half an hour and found a lot of little things.
These effects have a big impact on the sound creation and should only be released when they work 100%. Maybe the effects should be integrated in small steps.
The YKCorus, TALReverb3, the 3Band EQ, A Phaser work quite well.

But it's a lot of fun to experiment with it. The sound is always of high quality. It's really amazing what you can do with a Raspberry.

Peter

jnonis:
@Banana71 

> * I can't save a performance with my RPi 4. The little Pi crashes immediately.
I can look on this, I guess is something related with the disabled TGs.
Do you have all 16 TGs enabled?

> * The names of some parameters are not consistent.
> * The functions of some effects don't work.
Most of the effects are ports of LV2 plugins, I guess some paramters are broken on the originla code.

> * The flanger doesn't do what a flanger should do at all.
Flanger port is a bit tricky to setup.

> * The parameter feedback of the delay effect still doesn't work properly.
What do you mean? The delay works fine for me.

> * The menu navigation has been changed, why?
How? I didn't change the navigation.

jnonis:
> This is an awful lot of code to include, so its a little hard to say.
> 
> A couple of things I have noticed that I'll mention:
> 
> * A lot of the files don't have headers, is that an issue?
> * mididevice.cpp is no longer ignoring 1 byte messages, so I don't know what that will do performance wise if used with a device that does Active Sensing for example, as a lot more code appears to run every time now.
> * In minidexed.cpp I see the channels has been changed to 2 by default but previous it was only 2 when using a multi-core Pi.
> * In MiniDexed::Run for cores 2 and 3 I notice that there are now additional spinlocks being acquired to support MidiArp either side of the GetSamples call - I've no idea if this will affect audio processing performance in the multi-core system.
> * In MiniDexed::ProcessSound there are a lot of changes to the audio processing that I'm afraid I don't understand (it took me long enough to work out the previous behaviour I described here: [stereo fx processing #818 (comment)](https://github.com/probonopd/MiniDexed/discussions/818#discussioncomment-12405615)).
> * In particular there is now a stereo effects chain that seems to happen for non-multicore Pis, whereas previously, for performance reasons as I understand things, non-multicore Pis had no effects, had lower polyphony and only a mono output.
> * Also, like in Run there are now more spinlocks being processing around the getSamples, so the same concern applies here.
> * I'm guessing this is a "breaking change" for all the existing performances too (at least, I see the Reverb settings appear to have gone)?
> * uibuttons I see the debounce time has increased, but doesn't way why (but I guess is performance related)?
> * uimenu.cpp has changed the main menu handling (switching from MenuHandler to a new MainMenuHandler) in a way I don't quite understand.  I can't really comment on the other uimenu changes as the comparison gets out of sync with all the changes.
> * userinterface.cpp has changed how the rotary encoder is handled, and it isn't clear to me why.  Is this performance related again?
> 
> In short, this seems to be a fundamental change from effects being something that happens during TG mixing to produce the final audio, to being something that is now on a per TG basis and I don't know what that will mean for the system.
> 
> And the changes to the rest of the code seem quite significant to me, and I don't really feel qualified to be able to comment on the implications, but I'm certainly nervous that this might affect performance on a non-multicore (V1) Pi.
> 
> Kevin

I give up you. I don't know what is your problem

diyelectromusic:
> I give up with you. I don't know what is your problem

I was asked if I thought this was ready to merge, so spent a couple of hours attempting to understand the changes.  I've asked questions and raised possible concerns with the parts of MiniDexed I believe I understand (not being one of the original authors of the code) but have had to admit there are a lot of changes here, especially to some of the core, that I don't understand.

Ultimately, I don't feel able to comment if this is ready for merging or not.  I don't understand enough about what has changed to do that, so someone else will have to take a look.

Kevin

diyelectromusic:
> * Once there are no more hard showstoppers (like crashes, broken sound processing, etc.), we merge this code but ship it disabled by default

The issue here is that whilst the FX processing might be disabled from the menus, the changes to MIDI (now processing 1 byte messages), to V1 Pis (now having a stereo path), to the UI (button press times, encoder timings, etc), and to the audio processing chains are still all there - so even with no additional FX processing enabled there are some changes to the core code that I can't personally assess that they won't impact on the final system.

As I say, it needs either someone more knowledgeable than me to review it all and say it's fine; or some significant regression testing on a range of Pis and configurations whilst FX processing is disabled to give some confidence it will all be ok.  Preferably both.

It isn't as simple as having FX disabled by default I'm afraid. That won't turn off the other changes in this PR.

It might all be fine.  As I keep repeating, I only know enough to see there are changes here to the core that could cause issues.  I don't know enough to know if they will.

So how to get to this point I'm not sure (it's one of the reasons I started the conversation about a possible "dev" branch - to allow changes like this to be accepted on a trial basis until everyone's shaken the issues out).

Kevin

Issue #820, opened 2025-03-09
Performance: "Save as default" overwrites the first performance of the current bank


Banana71:
Performance: "Save as default" overwrites the first performance of the current bank but i think the command "Save as default" should only affect the Performance.ini from the root directory.

diyelectromusic:
I think I know what the issue is.  Give this PR a try!

Note: I've not had time to test it myself...

Banana71:
@diyelectromusic Kevin, Thank you very much.

probonopd:
You guys rock!

PR #822, opened 2025-03-10
Potential fix for Issue #820

diyelectromusic:
Potential fix for issue #820 where saving default performance always overwrites the first performance in the current bank, but should act on bank 0, performance 0.

Note: untested at this time by me.  If someone fancies trying it - please do!

Kevin

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1483,6 +1483,7 @@ bool CMiniDexed::DoSavePerformance (void)
 
 	if(m_bSaveAsDeault)
 	{
+		m_PerformanceConfig.SetNewPerformanceBank(0);
 		m_PerformanceConfig.SetNewPerformance(0);
 		
 	}


Banana71:
Hey @diyelectromusic Kevin, thank you very much, you were successful. On my setup "Save as default" now works great. ❤️

Banana71:
One more note: the miniDexed switches to the "default" bank after "save as default", which is logical.

diyelectromusic:
It would be more complicated to do otherwise as MiniDexed has a pretty short-term memory when it comes to performance banks - it basically has a current directory concept which is managed within the performance code, but the choice of saving as default or not is within the synth code (which doesn't know anything about the directories).

I could probably do something like:
* Remember the current bank/performance
* Select bank 0 and save as performance 0
* Restore previous bank/performance

That would probably work, but yes, I think leaving the system set to what you've just saved does largely seem right to me.

Kevin

probonopd:
Thanks @diyelectromusic for fixing this so swiftly and thanks @Banana71 for bringing this to our attention and for testing it.

Issue #11, opened 2022-02-20
Support selecting patches using MIDI Bank Change and Program Change messages


probonopd:
Many USB keyboards can send Bank LSB and Bank MSB Changes, and Program Changes. If we put several `.syx` files on the microSD card, then we might use Bank LSB and Bank MSB Changes to select those `.syx` files, and Program Changes to select patches.

(Once #8 is implemented I can say which MIDI messages we need to react to.)

probonopd:
Since our ultimate goal here is to run multiple instances of Dexed (e.g., up to 8 instances akin to the [TX816](https://www.vintagesynth.com/yamaha/tx816.php) and the [TX802](https://www.vintagesynth.com/yamaha/tx802.php), we should also think of a way to define and load "performance" patches (basically a collection of up to 8 patches).

dcoredump:
Will do this with adding MIDI processing support for Synth_Dexed (see also #9, #10, #11)

probonopd:
@dcoredump in the meantime I've been trying to add an interim solution for loading different default voices but I can load only voice 0. I must be doing a stupid C mistake, can you spot it?

https://github.com/probonopd/MiniDexed/commit/b0038a750a29ae6bfa83815cbe0227638f6f48aa

dcoredump:
Currently not much free time for me, sorry.

I think you forgot to decode the 138 bytes of voice data:

uint8_t decoded_voice[156];
dexed.decodeVoice(progmem_bank[b][v],decoded_voice);

probonopd:
Decoded them using `sysex2c.py --decode´. Voice 0 works. Must be something else?

dcoredump:
> Decoded them using `sysex2c.py --decode´. Voice 0 works. Must be something else?

This is right, but you get 128 bytes per voice and some bytes are having more than one parameter stored. You have to decodem them before (as I wrote above).

probonopd:
Thanks. Just to be sure @dcoredump. By using `--decode´  get more than 128 bytes C code. Do I have to decode them _again_ in MiniDexed? Or should I _not_ use  `--decode´ in Python and _just_ do the decoding in MiniDexed?

probonopd:
I mean, in the end we may want to load voices from `.syx` files stored on SD anyway, most likely by porting
https://codeberg.org/dcoredump/MicroDexed/src/branch/master/dexed_sd.cpp - maybe @rsta2 could help with this?

dcoredump:
[...]

PR #830, opened 2025-03-27
MIDI CC 11 (Expression)

diyelectromusic:
Initial implementation of MIDI CC 11 (Expression) implemented as a "live" performance parameter - i.e. MIDI only, not settable via the UI.

This wasn't originally supported in the DX7 or DX7II but is described as setting the percentage of the channel volume.  It is implemented so that the MIDI range 0..127 maps onto 0% to 100% of the channel volume.

The current setting is preserved through voice and performance changes.  There could be an option to reset it back to the default (max, 127) on voice or performance change if required.

Kevin

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -112,6 +112,7 @@ void CConfig::Load (void)
 	m_nMIDISystemCCVol = m_Properties.GetNumber ("MIDISystemCCVol", 0);
 	m_nMIDISystemCCPan = m_Properties.GetNumber ("MIDISystemCCPan", 0);
 	m_nMIDISystemCCDetune = m_Properties.GetNumber ("MIDISystemCCDetune", 0);
+	m_nMIDIGlobalExpression = m_Properties.GetNumber ("MIDIGlobalExpression", 0);
 
 	m_bLCDEnabled = m_Properties.GetNumber ("LCDEnabled", 0) != 0;
 	m_nLCDPinEnable = m_Properties.GetNumber ("LCDPinEnable", 4);
@@ -351,6 +352,11 @@ unsigned CConfig::GetMIDISystemCCPan (void) const
 unsigned CConfig::GetMIDISystemCCDetune (void) const
 {
 	return m_nMIDISystemCCDetune;
+}
+
+unsigned CConfig::GetMIDIGlobalExpression (void) const
+{
+	return m_nMIDIGlobalExpression;
 }
 
 bool CConfig::GetLCDEnabled (void) const
--- a/src/config.h
+++ b/src/config.h
@@ -131,6 +131,7 @@ public:
 	unsigned GetMIDISystemCCVol (void) const;
 	unsigned GetMIDISystemCCPan (void) const;
 	unsigned GetMIDISystemCCDetune (void) const;
+	unsigned GetMIDIGlobalExpression (void) const;
 
 	// HD44780 LCD
 	// GPIO pin numbers are chip numbers, not header positions
@@ -267,6 +268,7 @@ private:
 	unsigned m_nMIDISystemCCVol;
 	unsigned m_nMIDISystemCCPan;
 	unsigned m_nMIDISystemCCDetune;
+	unsigned m_nMIDIGlobalExpression;
 
 	bool m_bLCDEnabled;
 	unsigned m_nLCDPinEnable;
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -95,6 +95,15 @@ CMIDIDevice::CMIDIDevice (CMiniDexed *pSynthesizer, CConfig *pConfig, CUserInter
 	m_MIDISystemCCBitmap[2] = 0;
 	m_MIDISystemCCBitmap[3] = 0;
 
+	m_nMIDIGlobalExpression = m_pConfig->GetMIDIGlobalExpression();
+	// convert from config channels 1..16 to internal channels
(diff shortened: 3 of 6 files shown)


soyersoyer:
Shouldn't LoadPerformanceParameters reset it to 127?

GordonGunzelman:
Thank you for your work on this.  It works fine on Channel 1 but there is no expression on the other channels.  You're right saying that Y never put this in the early models since it was not in the original MIDI specification.  It was only added in after poly timbral synths were developed.  This allowed the musician to set relative levels on each channel (CTRL 07) and then add real time expression (CTRL 11) without changing the balance between voices.  This is exactly what I want to do in the organ as each manual sends on a different channel but for some sounds I would be layering more than one part (TGn).  As an aside, I do not know of any modules that implemented CTRL 07 as a global (OMNI mode) to allow the entire module to change it's output at once.  Both Controllers should operate independently on each part, multiplied together as you did.  (Back when I was working in Fortran and COBOL we would have made the final division by 16,384 in the source code,  The machines were slow enough that eliminating that second multiply would make a noticeable difference in execution time.)
 
I don't know why it is working only on MIDI CH1 but at least I know it's possible.  The MiniDexed sounds are a great addition to the (4 Manual) organ which, IMHO, is severely lacking in solo tones. 

Gordon Gunzelman

Banana71:
I haven't tested anything yet. Just my thoughts.

I assume that if you use the miniDexed in a church, you want to control the master volume. Changing the volume via the CC sets the volume of all TGs to the same value, and the volume balance in a performance is lost. Instead, the MIDI CC (#7 or #11) should be routed to the master volume (Sys ex). This function should be configurable, e.g., in the minidexed.ini file.

```
# Mastervolume
Mastervolumecontrol=On 
MIDICCtoMastervolume=#07
# Mastervolumecontrol=On/Off enable/disable, Mastervolume on or off
# MIDICCtoMastervolume=#07   Configueres which MIDI CC changes the master volume (e.g., #07, #11, ...)
```

Peter

diyelectromusic:
> Shouldn't LoadPerformanceParameters reset it to 127?

I wasn't sure. I did think that initially but then I thought if this is (for example) being controlled by an expression pedal resetting it would give a disconnect between the position of the pedal and the setting, which I thought would make for a really clunky experience...

diyelectromusic:
> I assume that if you use the miniDexed in a church, you want to control the master volume.

That may be true, but I was attempting to implement what the MIDI CC seesm to be meant for.

> Changing the volume via the CC sets the volume of all TGs to the same value, and the volume balance in a performance is lost.

Actually I don't think that would be true.  The Expression is combined with the master volume, and I've deliberately not included it in the UI/performance settings.

I read it as a modifier to the existing volume settings, so at full (127) all channel volumes will be whatever the mix in the performance settings say.  At 64, it will be half of whatever the mix settings are.  So actually I think it is a way for live adjustments of all TGs on that channel that respects the mix?

We could allow for it to be routed, but actually I think using 11 rather than 7 works better for performances.  But I haven't tried it myself yet as I need to set up a MIDI CC test controller :)

The key here is that Expression isn't part of the configuration - it is a way for live modification of the settings without changing the intent so I guess it was an early attempt at an alternative to touch sensitivity?

But feel free to correct me if I've misunderstood.  As it wasn't originally supported I was mostly making it up :)

Kevin

diyelectromusic:
> This allowed the musician to set relative levels on each channel (CTRL 07) and then add real time expression (CTRL 11) without changing the balance between voices.

I'm not clear how this was intended then as by definition this is a MIDI channel message.

> Back when I was working in Fortran and COBOL we would have made the final division by 16,384 in the source code,

I've used 127*127 as that is the maximum value it will have, and gain should be 1.0 at full volume (give or take floating point errors).

> The machines were slow enough that eliminating that second multiply would make a noticeable difference in execution time.)

If I was using 16384 I'd just >> 14 :)

> I don't know why it is working only on MIDI CH1 but at least I know it's possible. The MiniDexed sounds are a great addition to the (4 Manual) organ which, IMHO, is severely lacking in solo tones.

As I say, by definition, a MIDI channel message, so I'm not sure what to do about this.

We could add it in so that CC 11 is received by OMNI regardless of the TG channel setting but whilst that satisfies what it seems the CC is meant to be for, that is taking liberties a little with the MIDI spec and configuration...

Maybe we need a
```
MIDIGlobalExpression=True
```
configuration parameter?

But it will make a bit of a mess of the MIDI handling... but not by much...

Kevin

probonopd:
Does anyone know an existing instrument (ideally from big Y) that we could compare our implementation with?

Banana71:
As I understand Gordon, he doesn't want an expression, just a way to change the volume. This means that MIDI CC #11 is routed to the volume in miniDexed. Otherwise, I like my idea as described above. 😂

diyelectromusic:
To be honest, I'm trying to solve it for what it looks like MIDI has intended it to do and then hopefully we can find a way to map it onto the organ.  I don't want a "this specific organ mode" :)

But happy to be guided to what we think is most useful!

diyelectromusic:
Added:
```
MIDIGlobalExpression=
```
as an option in minidexed.ini.

Set to 1 to 16 to specify a MIDI channel on which to receive a global Expression CC (11) which will then be applied to all tone generators.  All other CC Expression messages on any other channels are ignored.

Set to 0 (or >16 or omit altogether) to revert to MIDI Expression CC (11) to be channel only.

See how that works...?

probonopd:
Thanks @diyelectromusic. Please don't forget to document it in the wiki.

PR #833, opened 2025-04-09
Initial attempt at caching performances to improve performance load times

diyelectromusic:
Looking at the performance issues when loading performances.

This change adds a "caching layer" between minidexed.cpp and the circle properties files, so that if a file has already been loaded then it doesn't need to be grabbed off the SD card a second time.

This has the following properties:
* First time a performance is selected it will need to get it from the card which will be slower.
* All subsequent times the performance is loaded it is loaded from memory.
* When the banks are changed the cache is cleared so all performances will once again have to be obtained from the card.
* I believe save as new, overwrite and delete all work, but it needs some serious ttesting.
* I think this makes `PerformanceSelectToLoad=0` much more feasible to use now, meaning selecting performances can be a lot more similar to selecting voices.
* When first loading performances it is still possible for the menu to lag compared to the selected performance due to loading from the card.
* It should be possible to include a start-up mode to preload all performances in a bank - tbd.  But it isn't clear what to do on bank changes - should it load all performances in that bank too?  That will be slow I'd imagine.
* If using PCCH and MIDI and your programmer sends BANKSEL+PC each time, then the cache will be invalidated on every BANKSEL.  But if your programmer only sends Program Change, the cache will work as expected.

I don't know how much memory this takes up, so it might not work on smaller PIs... I need to have a look at that.  I'm running on a PI 3A+ fine as far as I know so far.

Please give it a go and let me know any issues :)

If this works, then I hope it essentially fixes #694 and similar issues (tbc).

Kevin

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1913,55 +1913,57 @@ bool CMiniDexed::DoSavePerformanceNewFile (void)
 void CMiniDexed::LoadPerformanceParameters(void)
 {
 	for (unsigned nTG = 0; nTG < CConfig::AllToneGenerators; nTG++)
+	{
+
+		BankSelect (m_PerformanceConfig.GetBankNumber (nTG), nTG);
+		ProgramChange (m_PerformanceConfig.GetVoiceNumber (nTG), nTG);
+		SetMIDIChannel (m_PerformanceConfig.GetMIDIChannel (nTG), nTG);
+		SetVolume (m_PerformanceConfig.GetVolume (nTG), nTG);
+		SetPan (m_PerformanceConfig.GetPan (nTG), nTG);
+		SetMasterTune (m_PerformanceConfig.GetDetune (nTG), nTG);
+		SetCutoff (m_PerformanceConfig.GetCutoff (nTG), nTG);
+		SetResonance (m_PerformanceConfig.GetResonance (nTG), nTG);
+		setPitchbendRange (m_PerformanceConfig.GetPitchBendRange (nTG), nTG);
+		setPitchbendStep (m_PerformanceConfig.GetPitchBendStep (nTG), nTG);
+		setPortamentoMode (m_PerformanceConfig.GetPortamentoMode (nTG), nTG);
+		setPortamentoGlissando (m_PerformanceConfig.GetPortamentoGlissando  (nTG), nTG);
+		setPortamentoTime (m_PerformanceConfig.GetPortamentoTime (nTG), nTG);
+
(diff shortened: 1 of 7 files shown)


diyelectromusic:
> CPropertiesFatFsFile also has a cache...

I was going on what Rene has said in the past about file system access (https://github.com/rsta2/circle/discussions/357#discussioncomment-5537091).

Of course, once loaded in, the properties are not read from disk again, but MiniDexed treats each performance as a separate Properties file and so when changing performances the circle Properties object is recreated on each new load and so is re-read from disk each time.

I believe this could be what is slowing down the access, especially if using a slower SD card.  Well, it seems to be doing that for me anyway but happy to be shown to be wrong if the problems lie elsewhere and this doesn't help with the originally reported issue.

Kevin

soyersoyer:
For me, the problem no longer occurs even when using only the 2nd commit.

I didn't know there was no fat buffer cache. Maybe it speeds it up if that performance was already loaded before.
By the way, the code leaks in a few places, it might be worth fixing. ("= nullptr" does not delete the object)

diyelectromusic:
I wondered about that, but the original code seemed to let m_Properties go out of scope quite happily and then recreate a new object when loading in the new file, so I figured there must be some garbage collection going on...

But as I say, it is experimental at the moment so my priority was to see if it actually seems to fix the original issue rather than robustly worry about things like that right now.

Kevin

diyelectromusic:
> For me, the problem no longer occurs even when using only the 2nd commit.

Were you seeing 3 second delays?  If not, then you might have not been seeing the original problem anyway, just the menu updating one.

I'm waiting for confirmation from some of the original posters, but as I say, I think SD card performance has a lot to do with it, so I was trying to take that "out of the loop".

Kevin

soyersoyer:
This is a placement new, it doesn't allocate new memory, it uses the existing one.
```
new (&m_Properties) CPropertiesFatFsFile(nFileName.c_str(), m_pFileSystem);
```

soyersoyer:
> Were you seeing 3 second delays? If not, then you might have not been seeing the original problem anyway, just the menu updating one.

yes

diyelectromusic:
Ok, then maybe this is unnecessary.  I've not seen any significant delays myself, so was only going on what has been reported and attempting to think through the issues.

(That would have been useful to have said on the original discussion of course...)

Kevin

probonopd:
Thanks @diyelectromusic. Please feel free to reopen should you or anyone determine that this is, in fact, necessary.

PR #834, opened 2025-04-09
Optimise performance menu

diyelectromusic:
Alternative possible fix for #694 to replace #833 on suggestion of @soyersoyer .

If this works and solves the original problem then it is a lot less risk to include.

Kevin

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1932,55 +1932,57 @@ bool CMiniDexed::DoSavePerformanceNewFile (void)
 void CMiniDexed::LoadPerformanceParameters(void)
 {
 	for (unsigned nTG = 0; nTG < CConfig::AllToneGenerators; nTG++)
+	{
+
+		BankSelect (m_PerformanceConfig.GetBankNumber (nTG), nTG);
+		ProgramChange (m_PerformanceConfig.GetVoiceNumber (nTG), nTG);
+		SetMIDIChannel (m_PerformanceConfig.GetMIDIChannel (nTG), nTG);
+		SetVolume (m_PerformanceConfig.GetVolume (nTG), nTG);
+		SetPan (m_PerformanceConfig.GetPan (nTG), nTG);
+		SetMasterTune (m_PerformanceConfig.GetDetune (nTG), nTG);
+		SetCutoff (m_PerformanceConfig.GetCutoff (nTG), nTG);
+		SetResonance (m_PerformanceConfig.GetResonance (nTG), nTG);
+		setPitchbendRange (m_PerformanceConfig.GetPitchBendRange (nTG), nTG);
+		setPitchbendStep (m_PerformanceConfig.GetPitchBendStep (nTG), nTG);
+		setPortamentoMode (m_PerformanceConfig.GetPortamentoMode (nTG), nTG);
+		setPortamentoGlissando (m_PerformanceConfig.GetPortamentoGlissando  (nTG), nTG);
+		setPortamentoTime (m_PerformanceConfig.GetPortamentoTime (nTG), nTG);
+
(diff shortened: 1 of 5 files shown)


Banana71:
I tested the version on my Rpi4 and noticed a clear improvement.
Load, Save, and Delete works as expected.

The behavior I demonstrated in the video is still there:https://github.com/probonopd/MiniDexed/discussions/694#discussioncomment-11217007
During the wait time between a single click and a double click, scrolling in the menu is not blocked. Is this a problem that needs to be addressed? 
I set the DoubleClickTimeout to 200ms, and the usability has improved even further.

```
# Timeouts in milliseconds for double click and long press
DoubleClickTimeout=200 #(default=400)
LongPressTimeout=400
```
Peter

Banana71:
@diyelectromusic @probonopd , I would add the following lines to the [wiki](https://github.com/probonopd/MiniDexed/wiki/Files#timeouts) on the topic of timeout:


**For optimal responsiveness and seamless operation, the following configuration values have proven to be effective:**
```
DoubleClickTimeout=200
LongPressTimeout=300
```
**This configuration requires a well-installed encoder.**



Is that okay with you?

diyelectromusic:
> The behavior I demonstrated in the video is still there:[#694 (comment)](https://github.com/probonopd/MiniDexed/discussions/694#discussioncomment-11217007)

The problem here I think is that the selection of performances is an asynchronous activity.  I'm guessing that is something to do with not blocking for card reads or something.  It certainly makes the processing paths more complex, so I'm not sure how easy that would be to sort out.

Does the other change (https://github.com/probonopd/MiniDexed/pull/833) also have the issue?  I'm guessing it would, but it would be interesting to see if it is any better.

Kevin

diyelectromusic:
> @diyelectromusic @probonopd , I would add the following lines to the [wiki](https://github.com/probonopd/MiniDexed/wiki/Files#timeouts) on the topic of timeout:
> 
> **For optimal responsiveness and seamless operation, the following configuration values have proven to be effective:**
> Is that okay with you?

If these are preferred and well-tested values (on all versions of Pi) then I see no reason not to just change the defaults in the code/ini file?

Kevin

probonopd:
Thanks Kevin, I was to ask exactly that :-)

soyersoyer:
Can't you cherry-pick my commit instead? That would be more clean.

probonopd:
Are parts of this still needed with #841?

diyelectromusic:
> Are parts of this still needed with #841?

As per my comments in #841 I think this fixes the issue whilst still retaining the option to use the older style loading if the new style causes issues, which as I've stated I think has benefits and see no reason to remove.

I also think this is a more elegant way to solve it than the 500mS, so would prefer this approach.

The only thing missing, as I asked for opinions, was changing the default in minidexed.ini.

@soyersoyer I'll leave it to @probonopd to decide how best to incorporate this change, but the fact you hadn't submitted it directly, left me to assume you wanted someone else to do it on your behalf.  My apologies if that wasn't the case and you'd rather submit it directly.

Kevin

probonopd:
> When I use the MiniDexed with a sequencer, it's useful that it doesn't switch to another performance right away

...assuming you are not using the sequencer to also do the switching. Is that a real use case of yours?

And wouldn't the same be true for the voices as well?

diyelectromusic:
Now superseded by https://github.com/probonopd/MiniDexed/pull/874 so can be closed.

soyersoyer:
> ......assuming you are not using the sequencer to also do the switching. Is that a real use case of yours?

I'm just jamming sometimes.
So in this case the sequencer runs continuously and at the end of the sequence (e.g. after 4x64 steps) I would like to change the performance.  So I choose the performance in advance, and when the time comes, I simply change it with a click.

> And wouldn't the same be true for the voices as well?

I usually only change the performances, not the voices.

PR #837, opened 2025-04-14
Handle CC 124, 125, 126, 127 (Omni Mode Off, Omni Mode On, Mono Mode On, Poly Mode On)

probonopd:
Try to implement some the functionality requested in https://github.com/probonopd/MiniDexed/discussions/836 and below:

1. Omni Mode On/Off via MIDI CC: MIDI CC **124** and **125** are now handled to switch Omni Mode Off and On, respectively.
2. Mono Mode On, Poly Mode On via MIDI CC: MIDI CC **126** and **127** are now handled to switch Mono Mode On, Poly Mode On, respectively.
3. If `MIDIDumpEnabled=1` and syslog is configured, then MIDI messages are now also sent to syslog over the network. This helps to see what is going on without the need for a HDMI screen.

The following are __not__ implemented here and will need to be handled in another pull request:

1. Voice Dump Request via SysEx: MiniDexed should listen for the specific SysEx message (`F0 43 20 00 F7`) and send the voice dump for the requested channel. (Reference: https://github.com/asb2m10/dexed/issues/141) - Possibly should be handled by Synth_Dexed?
1. Mute Operator via SysEx: MiniDexed should handle the SysEx message for muting operators (`F0 43 11 01 1B 2F F7`) and toggle the mute state of the specified operator. - Possibly should be handled by Synth_Dexed?


diyelectromusic:
Are you sure that is doing the voice dump correctly?  The DX7 manual states there are four formats to be supported, but this seems to just be calling getSysExVoiceDump() in miniDexed which is a bit of a hack we had, which only supports one option as far as I can see - non-packed voice as per below and I seem to recall it was pretty indiscriminate about which MIDI channels/ports is spammed out on - in fact I think it sent them to all?

image

Also, as these are Yamaha specific, they should probably be included in the TG specific sysex handling further down in mididevice.cpp where the DX device number (which we kind of use as a sysex proxy for channel) is checked.

I can't find any mention of Mute in the DX MIDI specs I have - do you have a proper reference?  It looks like it is only checking device (channel) 1...

That call to setOperatorMute is passing in two fixed message bytes - they have just been checked to be 0x1B and 0x2F yet they are passed in as an operator number and TG number... I'm pretty sure we don't have 47 tone generators or 27 operators...

As far as I can see (trying to decode the sysex from https://www.synthmanuals.com/manuals/yamaha/tx816/owners_manual/tx816e1.pdf):
f0 Start SysEx
43 Yamaha
11 Substatus 1 = parameter change; channel=1 (i.e. MIDI channel 2)
01 g=0; h=1 so table 4.1
1b p=27 so operator on/off (parameter 155, yes)
2f - presumably the data?  So if so, this is turning on operators 1,3,4,5 and 6 - i.e. muting operator 2...?
f7

image
image

So if all that is the case then this function will need to check for MIDI channel/device; should allow the 5th byte to be a parameter and then use the TG from the for loop in the MIDI handler to know which synthesizer to send this to, and then use the parameter as a bit mask for the operators... I think?

Also, if this is muting TGs should it be toggling them or setting them to zero?  Again where is the spec for this mute function?  Or is it just the specific voice parameter for operator on/off?

If the latter, then we already have a function to handle voice parameters - SetVoiceParameter() but it doesn't look like it is tied up to SysEx at the moment - just to the UI...

As we are supporting Omin On/Off presumably we ought to also support Poly On/Off too:
image

(Aside, if we're going to start letting copilot in to mess with the code, then someone else will have to pick up the reviewing.  It's hard enough reviewing code with human intent let alone guessing what "AI" may or may not have hallucinated...)

Kevin

probonopd:
> Are you sure that is doing the voice dump correctly?

No... I tried to re-use what we already have.

The current implementation in `getSysExVoiceDump()` method only supports a single format (the voice edit buffer format, f=0). This doesn't match the DX7 specification that shows four different dump request formats:

- Voice edit buffer (f=0)
- Supplement edit buffer (f=5)
- Packed 32 supplement (f=6)
- Packed 32 voice (f=9)

In `mididevice.cpp`, in the `MIDIMessageHandler` method, there's code that detects a voice dump request with this pattern:
```
if (nLength == 5 &&
    pMessage[0] == MIDI_SYSTEM_EXCLUSIVE_BEGIN &&
    pMessage[1] == 0x43 &&
    (pMessage[2] & 0xF0) == 0x20 && // Check for Yamaha SysEx with channel // 0010nnnn n= MIDI ch nr
    pMessage[3] == 0x00 && // 0fffffff f=0 Transmits voice in Edit Buffer; see above for f=0,5,6,9
    pMessage[4] == MIDI_SYSTEM_EXCLUSIVE_END)
```

Currently this only detects format 0 requests (voice edit buffer) and doesn't properly handle the other formats.
Ideally we would support all 4 formats, but I am not sure whether we have methods to get the other 3 yet.

The current implementation calls `SendSystemExclusiveVoice` which indeed seems to send to all MIDI outputs. It should probably only respond on the MIDI channel encoded in `pMessage[2]`.

> I can't find any mention of Mute in the DX MIDI spec

Same here. It  was mentioned in #837 that "Parameter number 155, Operator on/off" is `f0 43 11 01 1b 2f f7` but I could not find that sysex documented in DX/TX instruments. Maybe @Skerjanc can point us to some official documentation?

> As we are supporting Omin On/Off presumably we ought to also support Poly On/Off too

Good point.

diyelectromusic:
Well I still think it ought to be part of the HandleSystemExclusive() mechanism that already exists, but I've not really looked at that before...

But that already takes care of the Yamaha specific device number in a consistent (although, still debateable if correct) way and then checks a whole pile of other parameters by calling into Synth_Dexed.

I just don't know without working through it all a bit more which ones are supported and which aren't, but if these are not supported then they ought to be added to Synth_Dexed rather than doing something special in MiniDexed in my view...

The TX816 manual has a fairly complete, albeit hard to work through, description of its MIDI support, so I figure what is described as "mute" is just poking the voice parameters directly...

Kevin

diyelectromusic:
Hmm.  This looks like it ought to handle voice parameters anyway...
https://github.com/probonopd/MiniDexed/blob/main/src/mididevice.cpp#L690

As part of this:
https://codeberg.org/dcoredump/Synth_Dexed/src/branch/master/src/dexed.cpp#L990

So it would need some figuring out to see why it isn't already handled.

Kevin

diyelectromusic:
> Good point, doesn't this look like it should already honor the voice dump request?

It does, except that it doesn't seem to be matched with the functionality in Synth_Dexed: https://codeberg.org/dcoredump/Synth_Dexed/src/branch/master/src/dexed.cpp#L953

The whole thing is a bit of a mess tbh...

I'm trying to see where the operator enabling occurs, and it is listed in all the manuals as voice parameters 155 and 156, yet our data structures only go up to the end of the VOICE NAME (154)...

But in Synth_Dexed, operator status seems to be a controller function (see setOPAll())... so it isn't clear how this is meant to be handled - is it part of a "voice" or not?  You'd think having operators on or off would be part of the voice specification, but I've never had a DX7 to programme myself :)

Either way, I think this ought to be split out from the omni/poly stuff which is a much simpler change.  We need to properly figure it out how we can support:
* sysex voice dump requests
* sysex voice parameter changes

But it will need a bit of investigation first to work out what exactly is already going on and what is and isn't supported...

Kevin

Skerjanc:
Dear all, what great response to my request!
Yes, there are four kinds of Dump Requests. The most interesting for me is the f=0 1 voice bulk data, the second is f=9 32 voice bulk data. All from the TX816 perspective.

To program and play eight modules simultaneously, each of them is addressed by different MIDI channels, but you have to switch them all to OMNI if you want to listen to them together. Setting OMNI on/off is luckily standard MIDI CC messages 124 and 125.

The mute implementation on the DX7 is strange: it is part of (Chap 4-1) Voice Parameter Change (g=0) as parameter 155, but not stored in the patch. Directly after the voice name.

Skerjanc:
Dear all, what great response to my request!
Yes, there are four kinds of Dump Requests. The most interesting for me is the f=0 1 voice bulk data, the second is f=9 32 voice bulk data. All from the TX816 perspective.

To program and play eight modules simultaneously, each of them is addressed by different MIDI channels, but you have to switch them all to OMNI if you want to listen them together. Setting OMNI on/off is luckily standard MIDI CC messages 124 and 125. Sending up to 8 Note on/off messages is not an option.

The mute implementation on the DX7 is strange: it is part of (Chap 4-1) Voice Parameter Change (g=0) as parameter 155, but not stored in the patch. Directly after the voice name.

diyelectromusic:
If you are talking about the buttons on the front of a TX816, as far as I recall that isn't a true OMNI setting, it is more about routing the physical MIDI connections around the system?  They allow the whole rack to be treated, well, like MiniDexed in performance mode :)

Using a microcontroller for the panel IO, when I was looking at my own pseudo-TX816, I used that to handle the MIDI routing to MiniDexed - see https://diyelectromusic.com/2023/01/22/minidexed-tx816/.

Kevin

Skerjanc:
> Build for testing: [MiniDexed_2025-04-14-a383775](https://github.com/probonopd/MiniDexed/suites/37165490227/artifacts/2943197411) Use at your own risk.

Tried it out today. OMNI On works, but when switch back to OMNI Off, all modules are then set to MIDI Ch 1. They do not store their channels they had before.

Skerjanc:
> If you are talking about the buttons on the front of a TX816, as far as I recall that isn't a true OMNI setting, it is more about routing the physical MIDI connections around the system? They allow the whole rack to be treated, well, like MiniDexed in performance mode :)
Yes, the +1 button at the front only switches single MIDI Out lines from the modules to the front MIDI Out jack. The front MIDI In goes into all modules. Via the MIDI OMNI command the modules can be set to respond to an individual MIDI channel or all cannels.

Skerjanc:
> If you are talking about the buttons on the front of a TX816, as far as I recall that isn't a true OMNI setting, it is more about routing the physical MIDI connections around the system? They allow the whole rack to be treated, well, like MiniDexed in performance mode :)

Yes, the +1 button at the front only toggles the MIDI Out lines from the modules to the front MIDI Out jack. The front MIDI In goes into all modules, if they are set to Common.

Over the Common MIDI In you can set each module individually to OMNI mode and back with the MIDI CC command.

Skerjanc:
> If you are talking about the buttons on the front of a TX816, as far as I recall that isn't a true OMNI setting, it is more about routing the physical MIDI connections around the system? They allow the whole rack to be treated, well, like MiniDexed in performance mode :)

Yes, the +1 button at the front only toggles the MIDI Out lines from the modules to the front MIDI Out jack. The front MIDI In goes into all modules, if they are set to Common.
Over the MIDI Common jack, you can set each module individually to OMNI and back with the CC command.

Skerjanc:
> If you are talking about the buttons on the front of a TX816, as far as I recall that isn't a true OMNI setting, it is more about routing the physical MIDI connections around the system? They allow the whole rack to be treated, well, like MiniDexed in performance mode :)

Yes, the +1 button at the front only toggles the MIDI Out lines from the modules to the front MIDI Out jack. The front MIDI In goes into all modules, if they are set to Common.
Over the MIDI Common jack you can set each module individually to OMNI and back with the CC command.

probonopd:
Ignore the "Implement Poly Mode (CC 127)" in the commit message above, it was a mistake (it will be squashed before merging anyway). It only implements OMNI mode channel memory. Still untested.

Skerjanc:
> Build for testing: [MiniDexed_2025-04-15-1040214](https://github.com/probonopd/MiniDexed/suites/37223683284/artifacts/2951048709) Use at your own risk.

Tested, works perfect now.

P.S.:
- How are the SYSEX messages implemented for performance? It does not respond to g=1,h=0: DX performance parameter change. 
  E.g. f0 43 10 04 40 32 f7 // Automated detuning is must-have for Omni sounds :-)

- Have you tried to improve the noticable latency? I measured more than 20ms, something under 10ms is a must for synthesizers.

probonopd:
> Tested, works perfect now.

Which of the 3 features did you test, and which work, which don't?

1. Voice Dump Request via SysEx
2. Mute Operator via SysE
3. Omni Mode On/Off via MIDI CC

Skerjanc:
> Thanks. Which of the 3 features did you test, and which work, which don't?
> 1. Voice Dump Request via SysEx
**does not work**
> 2. Mute Operator via SysEx
**does not work**
> 3. Omni Mode On/Off via MIDI CC
**fixed**

probonopd:
@Skerjanc if you have some time, could you please re-run all tests, ideally with a HDMI display connected to see the messages before it possibly crashes? Thanks.

probonopd:
I reduced the scope of this pull request and updated the description in the first post of this page accordingly.

probonopd:
Hello @Skerjanc, does this version work properly regarding CC 124, 125, 126, 127 (Omni Mode Off, Omni Mode On, Mono Mode On, Poly Mode On)? If yes, I'd like to merge it into the main branch. Thanks!

Skerjanc:
OMNI mode is working now. I could not test Mono/Poly over CC, as I am using the SYSEX messages.

probonopd:
I tested Mono/Poly over CC.

diyelectromusic:
I'm pretty sure fprintf was used for MIDIDump rather than LOGNOTE as circle doesn't always allow logging in some contexts and I think serial interrupt/MIDI handling was one of them.

I haven't tested it, but suspect MIDI dump will no longer work to HDMI now...?

PR #838, opened 2025-04-18
Normalize reverb time to sample rate

### LINKED ISSUE #563: Reverb time depends on the sample rate

probonopd:
For consistent decay across different sample rates

Closes #563

### DIFF
--- a/src/effect_platervbstereo.cpp
+++ b/src/effect_platervbstereo.cpp
@@ -201,8 +201,7 @@ void AudioEffectPlateReverb::doReverb(const float32_t* inblockL, const float32_t
     }
     cleanup_done = false;
 
-    rv_time = rv_time_k * (m_samplerate / 48000.0f);
-    float32_t rv_time_scaler_scaled = rv_time_scaler * (m_samplerate / 48000.0f);
+    rv_time = rv_time_k * (48000.0f / m_samplerate);
 
     for (uint16_t i=0; i < len; i++) 
     {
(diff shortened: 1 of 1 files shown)


probonopd:
@Banana71 if you have some time to test this, I'd appreciate it. Thanks!

Banana71:
Thanks @probonopd,

Is it correct that the basis for the calculation is 44.1 kHz?
Then the reverb tails will be a bit longer in the future, since in the past we always worked with 48 kHz. I might adjust the reverb times in the performances.

probonopd:
Excellent question. How could we know for sure? I can change the calculation to 48 kHz if you think that is the correct base.

Banana71:
Yes,
Test successful at 48,000 Hz, the reverb now sounds longer, as expected.
What would be correct?
With 48,000 Hz as the basis for calculation, the performance would sound unchanged.

probonopd:
Thank you @Banana71. Please test my latest change and then let me know if you think this is ready for being merged. Thanks!

probonopd:
This build should hopefully work as expected?

probonopd:
[Does not work as expected](https://github.com/probonopd/MiniDexed/issues/563#issuecomment-2817028777): At sample rates < 48000, the reverb time gets shorter. Now the behavior is exactly the opposite.
At a sample rate of 96000, miniDexed crashes.

probonopd:
Didn't have a chance to test this yet, but if you have some extra time, maybe you'll want to give this a try?

Banana71:
This doesn't work yet.
Higher sample rates result in shorter reverb times.
Lower sample rates than 48000 result in longer reverb times.
Peter

probonopd:
Running out of ideas. https://github.com/probonopd/MiniDexed/pull/838/files seems simple enough; there must be more to it than meets the eye.

probonopd:
@diyelectromusic do you have any idea?

Issue #718, opened 2024-09-07
Midi - Programm change doesn't Change patchname at display


Gitarrenzupfer:
Hello, i just got started to fiddle with this awesome project. I took a raspberry 3 and everything worked quite good, but if I send a program change message in Performance-Mode, the sound is changing, but not the patchname/number on the display. 

Any ideas/ hints?

greetings 
Gitarrenzupfer

diyelectromusic:
If I understand you correctly, that is a consequence of how performances are chosen in the menu.  The "[L]" which indicates which performance is loaded, will change but it doesn't automatically step-through the display to the new chosen performance.

There is a whole pile of considerations about what the display ought to show when MIDI messages are received that we need to work through at some point.

So this isn't an issue as such, it is as currently designed.  As I say, assuming I'm understanding you correctly.

Kevin

Gitarrenzupfer:
Hello Kevin,

Thanks for your quick reply. I understand, it is not an error or a bug, but "by design". I would be very pleased if it were changed in a software update and thus had the same behavior as all the other synths that I own or know.

To be honest, I don't see any point or advantage in the current behavior of the display when receiving MIDI program changes.

Regards

diyelectromusic:
> To be honest, I don't see any point or advantage in the current behavior of the display when receiving MIDI program changes.

"There is a whole pile of considerations about what the display ought to show when MIDI messages are received that we need to work through at some point."

I didn't say I agreed with the behaviour either - it's on the todo list ;)

(refs: https://github.com/probonopd/MiniDexed/issues/647 https://github.com/probonopd/MiniDexed/issues/549 https://github.com/probonopd/MiniDexed/pull/604 https://github.com/probonopd/MiniDexed/discussions/659 and probably quite a few more - as I say, it's on the todo list...)

Kevin

PR #841, opened 2025-04-18
Load performances like voices

probonopd:
Load performances in the same way voices are loaded (without requiring to select/click) to increase consistency.

Performances requiring an extra step has been bothering me for a while.

TODO:
* Retest https://github.com/probonopd/MiniDexed/issues/718 with this
* Remove `PerformanceSelectToLoad` configuration parrameter, since this PR makes that behavior the default


probonopd:
Now, in `CUIMenu::PerformanceMenu`, we call `SetNewPerformance(nValue)` immediately on every scroll event, causing a lag when scrolling through the performances fast.
To speed up scrolling through performances, we could delay the actual loading until the user has stopped scrolling for at least 500ms.

probonopd:
@diyelectromusic @Banana71 what do you think? 

As for me, I'm loving it. Makes it so much easier to go through the performances.

diyelectromusic:
I'm a little confused by this one I have to admit.  This is what PerformanceSelectToLoad=0 enables, and then with the more optimised use of the menus (as per https://github.com/probonopd/MiniDexed/pull/834), doesn't that give you this anyway?

Or have I missed something?

I'm a little nervous about assuming this will always be ok as performances are so SD card bound, unlike voices, so if someone has a slower card or lots of entries in a directory, I can see that it is useful for them to have the option of select-then-load rather than auto-selection on every turn of the encoder?  Or at least, that is what I'd assumed for it being the way it was anyway...

Kevin

probonopd:
Have you tried it out? This loads performances in the same way like voices were loaded all the time. Finally it is consistent.

diyelectromusic:
> Have you tried it out? This loads performances in the same way like voices were loaded all the time. Finally it is consistent.

No, but I know that the minidexed setting I mentioned and the menu update for performance in general (which affects both modes), which was required anyway, does the same thing already? Did you try that?

probonopd:
Thing is, I think we should have only one mode (consistent with what we do for the voices), and in that case we don't need to optimize the mode we are not using anymore.

(In case we really, really need to support 2 modes, at least they should be supported for both voices and performances with the same setting.)

Banana71:
I would always turn off Performance Autoload.

When I create performances, I load a template from the Labatory bank, for example. Then I tweak the voices on my PC using Dexed. If I accidentally turn the encoder while I'm doing this and another performance loads, it would ruin my mood.

But as Kevin @diyelectromusic already mentioned, we've been able to configure this function in the minidexed.ini file for a long time.

```
# Performance
PerformanceSelectToLoad=1
```

diyelectromusic:
> Thing is, I think we should have only one mode (consistent with what we do for the voices), and in that case we don't need to optimize the mode we are not using anymore.

Ironically, if we don't have the ability to choose, then I think we'd need something like I tried here https://github.com/probonopd/MiniDexed/pull/833 to allow your required single mode to work across all SD cards otherwise the "auto load" functionality (to match voices) could be quite unusable for some people.

So, lets change the default, and ensure it is documented that if the auto-loading is not working well for someone, then they can turn it off?  We can add that to the fault finding pages probably?

Kevin

probonopd:
Gosh, maybe I forgot something...

>  the "auto load" functionality (to match voices)

what is this? Do we already have it? What does it do? (Hard to keep everything in mind, sorry)

diyelectromusic:
Well considering this option has been in pretty much since performances were introduced, and for the reasons I've already mentioned (re SD card speed etc) there are reasons to have it configurable...

But as I say, with the menu fixes from the other patch (https://github.com/probonopd/MiniDexed/pull/834), I think we can change the default to make them behave the same.

Yes there is inconsistency, but as I said, the only way to make them the same would be to stop having to go to the SD card for every performance load, which is what I tried to do in https://github.com/probonopd/MiniDexed/pull/833.

They are functionally very different "under the hood" as voices are loaded a bank at a time, but performances are loaded individually.  We can't get away from that without some major rewriting of the performance system.  But change #833 gets us closer but would need significant testing.

It is good to have the option to still have a usable system but with clunkier performance loading - but I see no reason to make loading voices a lot worse (and really complicate the voice menus) just to make them the same... that would be a real step backwards imo.

But with change #834 we can make the default behaviour the same - so surely that is the solution?

Kevin

probonopd:
I fail to see why SD card speed would be an issue with my proposed solution. The SD card is only accessed 500ms after the user having stopped scrolling. Why is one extra klick going to help SD card speeds?

What bothered me from the beginning is that we are loading all voices at boot time rather at the time when they are needed.
This is why MiniDexed can have long boot times currently. 

In my opinion, everything should be read from SD card only at the moment when it is needed. That way, we could keep tens of thousands of voices (and performances) on the device without any slowdown.

diyelectromusic:
> I fail to see why SD card speed would be an issue with my proposed solution. The SD card is only accessed 500ms after the user having stopped scrolling. Why is one extra klick going to help SD card speeds?

It isn't about speeding up SD card access, it is about not letting SD card access slow down the responsiveness of the UI to the user.

> What bothered me from the beginning is that we are loading all voices at boot time rather at the time when they are needed. This is why MiniDexed can have long boot times currently.

Does it have long boot times?  I thought that was solved when I implemented directories to split up voice banks and performances?

> In my opinion, everything should be read from SD card only at the moment when it is needed. That way, we could keep tens of thousands of voices (and performances) on the device without any slowdown.

Well I'm pretty sure that reading from the SD card is the slowest operation MiniDexed will do so we ought to be really careful about when we do it.

Kevin

probonopd:
In this PR, the performance only gets loaded _after_ the user has stopped scrolling through the menu (500ms), So loading from SD should not slow down scrolling fast through the menu.

Yes, boot times are still long-ish imho - especially when you have really, really many voice syx.

diyelectromusic:
But a common usage is to listen to the voices as they are being selected to find one that sounds interesting to use.  To do that you have to now scroll, wait, listen, scroll, wait, listen.  That is then no different really to scroll, wait (for SD card), listen.  And actually isn't that much better than scroll, select [L], listen with the previous mode.

So it doesn't actually solve the main speed issue, just hides it behind the display updating.

It is fine if you know exactly which sound you are wanting to select (and by know I mean "remember the name or number") but doesn't help if you are selecting sounds based on what they are sounding like.

Kevin

probonopd:
With "wait" you mean the 500ms delay?

diyelectromusic:
If the default setting for `PerformanceSelectToLoad` is changed to 0, then I believe the main branch now has the desired behaviour thanks to https://github.com/probonopd/MiniDexed/pull/874

Kevin

probonopd:
Almost. We should rename `PerformanceSelectToLoad` to `SelectToLoad` and also apply the same behavior for voices.

probonopd:
In any case, it seems like we are loading performances so fast now that we don't need the 500ms trick anymore.

diyelectromusic:
> Almost. We should rename `PerformanceSelectToLoad` to `SelectToLoad` and also apply the same behavior for voices.

I really don't see the need.  The default now can be that they are the same.  This now only exists in the case of performance issues in my view, so is a lesser (compromised) experience for cases when someone isn't able to use performances a different way.

I think we just leave it as is, but change the default, then I expect most people won't even know it is there.

Kevin

probonopd:
>  I expect most people won't even know it is there.

You are probably right there - even I had totally forgotten it was there ;-)

PR #842, opened 2025-04-18
Introduce MasterVolume

probonopd:


probonopd:
Now, in `CUIMenu::PerformanceMenu`, we call `SetNewPerformance(nValue)` immediately on every scroll event, causing a lag when scrolling through the performances fast.
To speed up scrolling through performances, we could delay the actual loading until the user has stopped scrolling for at least 500ms.

diyelectromusic:
Moving to logarithmic is a good idea.
But just out of interest, why do you think we need to preset the master volume in the minidexed.ini?

Kevin

diyelectromusic:
Well that will only take effect until a MIDI update changes it back... surely the issue is more that you are presumably using a line out audio output with headphones!? :)

probonopd:
> surely the issue is more that you are presumably using a line out audio output with headphones!? :)

There might be truth in that ;-)

diyelectromusic:
> > surely the issue is more that you are presumably using a line out audio output with headphones!? :)
> 
> There might be truth in that ;-)

Seriously then, should this be more of a scaling factor so it applies to the UI and MIDI?  To set what "full volume" actually means?  e.g. set to 50% or 75% or something?

Kevin

Banana71:
I'm absolutely in favor of being able to adjust the master volume.
In my opinion, the default in the minidexed.ini file should be 100% (127).
Most users have probably set it up so that 100% is fine for them.

It would be really good if you could adjust the master volume via the UI. Bernhard (Lex) had a solution at the end of 2023 that I really liked: it allowed you to adjust the master volume in the "Load - Performance" menu by pressing and turning the encoder. Do you remember that?

Peter

probonopd:
> Seriously then, should this be more of a scaling factor 

Yes, a scaling factor. Isn't it one now? What do you mean by "so it applies to the UI and MIDI"? When I set it to 64, then everything should be scaled down by 50% globally, no matter what you do via the UI.

> It would be really good if you could adjust the master volume via the UI. 

Yes, but that would require us to rewrite minidexed.ini on the fly, and we don't have that functionality and I am also considering it as somewhat risky.

> Lex) had a solution at the end of 2023 that I really liked: it allowed you to adjust the master volume in the "Load - Performance" menu by pressing and turning the encoder. Do you remember that?

No... do you have a link?

Banana71:
> Yes, but that would require us to rewrite minidexed.ini on the fly, and we don't have that functionality and I am also considering it as somewhat risky.

That's probably a misunderstanding.
I would set the master volume in the miniDexed.ini file just like you did. And after each start of the miniDexed, the master volume is reset to the value defined in the miniDexed.ini. But while the miniDexed is running, you should also be able to adjust the master volume (for example, in the "Load Performance" menu by pressing and turning the encoder).
Nothing needs to be saved back to the miniDexed.ini file.

> No... do you have a link?

#551 
https://github.com/probonopd/MiniDexed/discussions/551#discussioncomment-7363295

probonopd:
Agree. That's how it should be done.

diyelectromusic:
Everything uses the same setMasterVolume call you added to the startup of Minidexed... so as soon as a MIDI command comes in, it will change it to 0..127 again.

To make this work, we need to add a new scaling factor that comes from minidexed.ini that is then multiplied by the volume required in setMasterVolume.  Or actually, to be honest, it should probably go in the section that processes the mix...

So I would say something like:

MasterVolumeScalingFactor = 0.0 to 1.0  or maybe 0% to 100%...

Then setMasterVolume will be something like:

```
void CMiniDexed::setMasterVolume(float32_t vol)
{
    if (vol < 0.0)
        vol = 0.0;
    else if (vol > 1.0)
        vol = 1.0;

    // Apply logarithmic scaling to match perceived loudness
    vol = powf(vol, 2.0f);

    nMasterVolume = scalingfactor * vol;
}
```

That kind of thing...

Then MIDI will still be able to adjust 0 to 127, but 127 will always only be the scaled version set in minidexed.ini?

Yes, we don't have a UI master volume - my mistake.  But we could (as per the code @Banana71 has found) and it should do the same by using set/getMasterVolume.

Kevin

Banana71:
> Then MIDI will still be able to adjust 0 to 127, but 127 will always only be the scaled version set in minidexed.ini?

That means you want to limit the maximum volume of miniDexed!?

```
# Master Volume (0-127)
MasterVolume=64
```
As far as I understand, it would be more likely these variable names, for example.
```
# Max Master Volume (0-127)
MaxMasterVolume=64
```
No matter what changes to the MasterVolume come from outside via MIDI SysEx, it will be limited (scaled) to the maximum volume from minidexed.ini.
If I understand correctly, I like that.

probonopd:
For me it would be ok of MasterVolume is how the synth starts up, and if someone wants to change the master volume over MIDI they should be able to, including the ability to set it _higher_ than the MasterVolume in the ini file. Wouldn't that be best? And isn't that what we have now?

Banana71:
Limiting the maximum volume protects people who have headphones plugged directly into the line output. 😄 
Whether we define a starting value or a limit in the miniDexed.ini is a matter of opinion. If we can't agree, a combination would also be conceivable.

```
#Master Volume (0-100%)
maxMasterVolume=80
StartMasterVolume=50
```

Personally, I would be satisfied with the current state with the additional option to change the volume in the UI.
I see it as absolutely necessary to be able to change the volume directly on a device at any time.

probonopd:
Would everyone be happy if we renamed `MasterVolume=` to `StartMasterVolume=`, and have it editable via MIDI and in the GUI, but not saving changes back to `minidexed.ini`? (We could always add a MaxMasterVolume=` later if it turns out to be needed.)

Banana71:
Yes, I would be happy with that. :heart_eyes:

diyelectromusic:
> Would everyone be happy if we renamed `MasterVolume=` to `StartMasterVolume=`, (to make clear that it is only the initial value) and have it editable via MIDI and in the GUI, but not saving changes back to `minidexed.ini`? (We could always add a MaxMasterVolume=` later if it turns out to be needed.)

I guess that is up to you as you're the one mostly affected as I understand things.  This won't stop you playing a MIDI file and suddenly getting unexpectedly blasted in your headphones if it sets the volume back to max...

Whereas having it as a global scaling factor would mean you never exceed that factor.

So as I say, it's up to you.

Has anyone put in a feature request for a system master volume in the UI?

Kevin

Banana71:
> Has anyone put in a feature request for a system master volume in the UI?

I actually never expressed this wish directly, but it was there from the beginning. My fault.
Bernhard, @lexonic implemented it in his fork once, and I really appreciated it.
@diyelectromusic Kevin, do you know an electronic musical instrument without the ability to adjust the volume on the device? That's simply essential for me.

I assume there are several users here who play the miniDexed as a standalone synthesizer in a setup, and in that case, influencing the volume is important.

I don't understand where you see the problem either? What's wrong with a menu item where we can adjust the master volume?

Alternatively, I would be happy with a noise-free, external circuit with a potentiometer. However, we should give users the ability to adjust the overall volume of the miniDexed.

Peter

Banana71:
> Oh I agree - I see no problem at all - it simply had never occurred to me, but then I tend to mess around with it more as 8 TGs than a single "performance" instrument :)

Yes, the miniDexed is used in a variety of ways. The trick is to get everything under one roof.

As far as I understand, an initial master volume `StartMasterVolume=xx` configured in the minidexed.ini file and the ability to change the volume via the UI are sufficient. Saving the settings back to the minidexed.ini file isn't necessary.

probonopd:
Well. Click and turn is already taken for Shortcut iirc.

So we'd need to have a "Master Volume" menu entry (on the same level as TG1 ... Performance is today).

Banana71:
> Well. Click and turn is hard to discover.

Yes, I agree. But that's where it's most needed.
1. Idea: Integrating the Master Volume into the Effects menu.
2. Idea: Click and Turn in the Performance Load Menue
3. Idea: MasterVolume in the same Leve as TG1 - TG8 - Effects - and Performance
```
  TG1-TG8         +----------------+      +---------------+   
                  |       MiniDexed|______|            TG1|___...
                  |<TG1           >|      |Voice         >|
                  +----------------+      +---------------+ 
                           |
                          ...
                           |             
                  +----------------+      +---------------+   
                  |       MiniDexed|______|            TG8|___...
                  |<TG8           >|      |Voice         >|
                  +----------------+      +---------------+ 
  Effects                  |                                                                      1. Idea
                  +----------------+      +---------------+      +---------------+      +---------------+   
                  |       MiniDexed|______|        Effects|______|        Effects|______|        Effects|
                  |<Effects       >|      |Compress      >|      |<Reverb       >|      |<MasterVolume  |
                  +----------------+      +---------------+      +---------------+      +---------------+ 
Performance                |
                  +----------------+      +---------------+   
                  |       MiniDexed|______|    Performance|___...
                  |<Performance   >|      |Load          >|
                  +----------------+      +---------------+ 
Master Volume              |                      | Click and Turn      2. Idea
                  +----------------+      +---------------+   
3. Idea           |       MiniDexed|______|   MasterVolume|
                  |<MasterVolume  >|      |<0-100%       >|
                  +----------------+      +---------------+
```
Does anyone have any other ideas? Otherwise, I'll vote for ideas 2 and 3.

soyersoyer:
I would prefer to use click+turn to change Bank in the Performance Load menu.

4. Idea: A configuration menu in the same level as TG1 - TG8 - Effects - Performance with options
MasterVolume, EngineType, MIDIRXProgramChange, ExpandPCAcrossBanks, PerformanceSelectChannel, PerformanceSelectToLoad
or a subset of them.

Banana71:
> I would prefer to use click+turn to change Bank in the Performance Load menu.

A very good idea. I also change the Performance Bank more often than I probably adjust the Master Volume.

PR #843, opened 2025-04-18
Prevent power save from kicking in

probonopd:
Similar to https://github.com/soyersoyer/MiniDexed/pull/5


soyersoyer:
I don't mind if you use my commits, but please cherry pick them instead of copy and add some braces to them. ..

probonopd:
It's a good practice (at least I am convinced) to always enclose them in braces. I have seen (and produced) way too many bugs due to them being left away. It's a matter of preference in coding style. In any case, full credit for the fix goes to you (as in the commit message). Thank you very much!

PR #844, opened 2025-04-18
use SoundFormatSigned24_32 format instead of SoundFormatSigned16

soyersoyer:
More detailed, and not much slower.


probonopd:
Thanks @soyersoyer

PR #848, opened 2025-04-19
Networking by @omersiar, new try

probonopd:
Unfortunately this PR is mixed up with 3 unrelated commits which need to be removed from it. I can't seem to get that done...


probonopd:
Closed in favor of 849.

Issue #23, opened 2022-02-20
Build for Raspberry Pi 2 fails


probonopd:
Trying to build a 32-bit version for Raspberry Pi 2 too but running into

```
config.status: executing depfiles commands
make[2]: Leaving directory '/home/runner/work/MiniDexed/MiniDexed/circle-stdlib/build/circle-newlib'
make[1]: Leaving directory '/home/runner/work/MiniDexed/MiniDexed/circle-stdlib/build/circle-newlib'
make[1]: *** [Makefile:879: all] Error 2
make: *** [Makefile:17: newlib] Error 2
Error: Process completed with exit code 2.
```

https://github.com/probonopd/MiniDexed/runs/5265872462?check_suite_focus=true#step:6:1184

It looks like some issue with circle-newlib. @smuehlst can it be built for 32-bit systems such as Raspberry Pi 2?

Maybe I am missing some necessary "cleanup" steps after having built the 64-bit versions?

smuehlst:
Sure, circle-stdlib can be built for all platforms that Circle supports. Very likely the cleanup is incomplete. Try running `make mrproper` from the circle-stdlib root directory, this should clean up all build results.

probonopd:
Still getting `make: *** [Makefile:17: newlib] Error 2`.

> this should clean up all build results

Does it do this recursively, also for newlib?

smuehlst:
> Still getting `make: *** [Makefile:17: newlib] Error 2`.

Can you show the complete sequence of commands that leads to this error? And what command invocation causes make to stop with the error?
 
> > this should clean up all build results
> 
> Should it do this recursively, also for newlib?

Yes, it should. See the circle-stdlib Makefile:

https://github.com/smuehlst/circle-stdlib/blob/47d9deb580f3d91201c2e275d3412abfbafdf156/Makefile#L77-L78

```
	-$(MAKE) -C $(NEWLIB_BUILD_DIR) clean
	-test -n "$(NEWLIB_INSTALL_DIR)" && rm -rf "$(NEWLIB_INSTALL_DIR)"/*
```

probonopd:
Can you show the complete sequence of commands that leads to this error?

* https://github.com/probonopd/MiniDexed/blob/main/.github/workflows/build.yml which invokes
* https://github.com/probonopd/MiniDexed/blob/main/build.sh

smuehlst:
You are doing the `make mrproper` in the wrong directory. I wrote:

[...]

PR #849, opened 2025-04-19
Networking by @omersiar

### LINKED ISSUE #43: Supporting RTP-MIDI

probonopd:
Network support by @omersiar, closes #43 (and much more).

Features as described on https://github.com/probonopd/MiniDexed/wiki/Networking (note to self: **update** the page once this is merged).

---

The networking has lingered around unmerged for way too long. Time to finally get it merged.

---

**Current showstopper: Sound crackles on RPi Zero 2, even if Networking is disabled.**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Get specific commits of git submodules
         run: sh -ex ./submod.sh
-      - name: Create kernels directory
-        run: mkdir -p ./kernels/
+      - name: Create sdcard directory
+        run: mkdir -p ./sdcard/
       - name: Put git hash in startup message
         run: |
           sed -i "s/Loading.../$(date +%Y%m%d)-$(git rev-parse --short HEAD)/g" src/userinterface.cpp
@@ -39,21 +39,21 @@ jobs:
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/bin):$PATH
           RPI=5 bash -ex build.sh
-          cp ./src/kernel*.img ./kernels/
+          cp ./src/kernel*.img ./sdcard/
 
       - name: Build for Raspberry Pi 4 (64-bit)
         run: |
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/bin):$PATH
           RPI=4 bash -ex build.sh
-          cp ./src/kernel*.img ./kernels/
+          cp ./src/kernel*.img ./sdcard/
 
       - name: Build for Raspberry Pi 3 (64-bit)
         run: |
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/bin):$PATH
           RPI=3 bash -ex build.sh
-          cp ./src/kernel*.img ./kernels/
+          cp ./src/kernel*.img ./sdcard/
 
       - name: Prepare SD card content for 64-bit
         run: |
@@ -63,15 +63,11 @@ jobs:
           make
           make armstub64
           cd -
-          mkdir -p sdcard
           cp -r ./circle-stdlib/libs/circle/boot/* sdcard
           rm -rf sdcard/config*.txt sdcard/README sdcard/Makefile sdcard/armstub sdcard/COPYING.linux
           cp ./src/config.txt ./src/minidexed.ini ./src/*img ./src/performance.ini sdcard/
           cp ./getsysex.sh sdcard/
           echo "usbspeed=full" > sdcard/cmdline.txt
-          cd sdcard
-          cp ../kernels/* . || true
-          cd -
           
       - name: Upload 64-bit artifacts
         id: upload64
@@ -89,8 +85,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Get specific commits of git submodules
         run: sh -ex ./submod.sh
-      - name: Create kernels directory
-        run: mkdir -p ./kernels/
+      - name: Create sdcard directory
+        run: mkdir -p ./sdcard/
       - name: Put git hash in startup message
         run: |
           sed -i "s/Loading.../$(date +%Y%m%d)-$(git rev-parse --short HEAD)/g" src/userinterface.cpp
@@ -107,14 +103,14 @@ jobs:
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-arm-none-eabi/bin):$PATH
           RPI=2 bash -ex build.sh
-          cp ./src/kernel*.img ./kernels/
+          cp ./src/kernel*.img ./sdcard/
 
       - name: Build for Raspberry Pi 1 (32-bit)
         run: |
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-arm-none-eabi/bin):$PATH
           RPI=1 bash -ex build.sh
-          cp ./src/kernel*.img ./kernels/
+          cp ./src/kernel*.img ./sdcard/
 
       - name: Upload 32-bit artifacts
         id: upload32
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,8 +29,6 @@ jobs:
            tar xf *-aarch64-none-elf.tar.xz
            wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-arm-none-eabi.tar.xz
            tar xf *-arm-none-eabi.tar.xz
-           # ls -lh ./circle-stdlib/install/aarch64-none-circle/include/stdint.h # For debugging (can be removed later)
-           find . -name 'stdint.h'
            mkdir -p kernels
     - name: Build for Raspberry Pi 5
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,29 +25,29 @@ jobs:
     - name: Install toolchains
       run: |
            set -ex
-           wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf.tar.xz
-           tar xf *-aarch64-none-elf.tar.xz
+           #wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf.tar.xz
+           #tar xf *-aarch64-none-elf.tar.xz
            wget -q https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-arm-none-eabi.tar.xz
            tar xf *-arm-none-eabi.tar.xz
            mkdir -p kernels
-    - name: Build for Raspberry Pi 5
-      run: |
-           set -ex
-           export PATH=$(readlink -f ./gcc-*aarch64-none*/bin/):$PATH
-           RPI=5 bash -ex build.sh
-           cp ./src/kernel*.img ./kernels/
-    - name: Build for Raspberry Pi 4
-      run: |
-           set -ex
-           export PATH=$(readlink -f ./gcc-*aarch64-none*/bin/):$PATH
-           RPI=4 bash -ex build.sh
-           cp ./src/kernel*.img ./kernels/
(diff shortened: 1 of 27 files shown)


probonopd:
I am lost. Why can't I get it to compile anymore?

The networking has lingered around unmerged for way too long. Time to finally get it merged.

probonopd:
Why does the build for RPi 2 fail now?

```
2025-04-19T13:56:24.7540844Z + cd src
2025-04-19T13:56:24.7541097Z + make clean
2025-04-19T13:56:24.7724440Z   CLEAN  /home/runner/work/MiniDexed/MiniDexed/src
2025-04-19T13:56:24.7877155Z + make -j
2025-04-19T13:56:25.2022613Z   CPP   main.o
(...)
2025-04-19T13:56:25.2728193Z   CPP   midipin.o
2025-04-19T13:56:25.2741275Z make: *** No rule to make target '../circle-stdlib/install/aarch64-none-circle/include/stdint.h', needed by 'net/ftpworker.o'.  Stop.
```

probonopd:
It searches for the non-existing

```
./circle-stdlib/install/aarch64-none-circle/include/stdint.h
```

but the follwing do exist:

```
./gcc-arm-10.3-2021.07-x86_64-arm-none-eabi/arm-none-eabi/include/stdint.h
./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/lib/gcc/aarch64-none-elf/10.3.1/include/stdint.h
./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/aarch64-none-elf/include/c++/10.3.1/tr1/stdint.h
./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/aarch64-none-elf/include/stdint.h
./circle-stdlib/libs/circle/addon/lvgl/lvgl/src/libs/thorvg/rapidjson/msinttypes/stdint.h
./circle-stdlib/libs/circle-newlib/newlib/libc/include/stdint.h
./circle-stdlib/libs/circle-newlib/newlib/libc/sys/linux/include/stdint.h
./circle-stdlib/libs/circle-newlib/newlib/libc/sys/linux/sys/stdint.h
```

probonopd:
Clearly it didn't like this. @smuehlst, maybe you have an idea why `./circle-stdlib/install/aarch64-none-circle/include/stdint.h` is searched but not found?

smuehlst:
> It searches for the non-existing
> 
> ```
> ./circle-stdlib/install/aarch64-none-circle/include/stdint.h
> ```

This file exists after a successful build of circle-stdlib. Maybe the build of `circle-stdlib` failed but the parent build did not detect that.

> 2025-04-19T13:56:25.2728193Z   CPP   midipin.o
> 2025-04-19T13:56:25.2741275Z make: *** No rule to make target '../circle-stdlib/install/aarch64-none-circle/include/stdint.h', needed by 'net/ftpworker.o'.  Stop.

This looks like there's a dependency in the Makefile from `net/ftpworker.o` to the `../circle-stdlib/install/aarch64-none-circle/include/stdint.h` header file. Why is there an explicit dependency to a system include file?

> Clearly it didn't like this (the build now also fails for aarch64 RPis for which it worked before). @smuehlst, maybe you have an idea why `./circle-stdlib/install/aarch64-none-circle/include/stdint.h` is searched but not found?

My current theory is that the build of `circle-stdlib` failed. Do you have a full log of the configure and make output for the build of `circle-stdlib`?

diyelectromusic:
I want to once again raise a concern that a lot of fundamental changes are happening all at once on the codebase and a lot of extra code is about to be added with this change too.

It is no wonder we're seeing weird crashes and breakdowns in audio quality.  Unless we find people more willing to regression test and maintain existing code than create new code, I fear this will only get worse.

Kevin

probonopd:
Hi @diyelectromusic. I hear your concerns but I think that this really needs to be merged before the code base deviates any further.

We can 1.) ship this disabled by default 2.) ensure that when it is disabled the code path doesn't run 3.) revert in case it turns out to cause major issues even if disabled.

But if we don't merge it now, the code base will deviate to the point that this can't be applied cleanly anymore. We have already let this linger for too long.

soyersoyer:
Why is this so important?

probonopd:
Hi @smuehlst, thanks for your help.

You can see the full build log [here](https://github.com/probonopd/MiniDexed/actions/runs/14551433963/job/40823071283). I am looking at it now, given your advice.

probonopd:
> Why is this so important?

Because with every unrelated merge, this becomes harder and harder to apply. It's been lingering for too long already.

smuehlst:
> Hi @smuehlst, thanks for your help.
> 
> You can see the full build log [here](https://github.com/probonopd/MiniDexed/actions/runs/14551433963/job/40823071283). I am looking at it now, given your advice.
> 
> Actually I think it is wrongly looking in `aarch64-none-circle` (the builds that are running before Raspberry Pi 2 are using that) when it should be looking in `arm-none-circle` (for Raspberry Pi 2)?
> 
> Maybe similar to [#23 (comment)](https://github.com/probonopd/MiniDexed/issues/23#issuecomment-1046315470)? Am I doing the `make mrproper` wrong?

When I look at the most recent build from the GitHub action, it seems that you were able to fix the build problem: https://github.com/probonopd/MiniDexed/actions/runs/14561379012. Am I understanding this correctly?

probonopd:
I am now building the 32-bit and the 64-bit versions in separate containers, to avoid "crosstalk". The builds are now succeeding. The issue may well be a missing step in `make clean` on my side, but I now have at least a workaround for it.

probonopd:
Hey hey, it's building now 👍 

Now we just need to test the hell out of it and make sure that it doesn't cause any atastrophic failures or sound degredation, at least when _turned off_.

probonopd:
Hangs on boot even with network disabled.

soyersoyer:
There is no scheduler?
add scheduler to the kernel private members:
```
	CScheduler	m_Scheduler;
```

probonopd:
@soyersoyer that was totally it. How did you find this so quickly? I need to learn your debugging skills :-)

soyersoyer:
Donno, I just read the source of [CScheduler](https://github.com/rsta2/circle/blob/master/lib/sched/scheduler.cpp).

probonopd:
Now even with Networking disabled I get crackling sound. Can it be that merely having the scheduler causes that?

soyersoyer:
And if you comment out the m_Scheduler?

probonopd:
Still crackles with Networking disabled. This really puzzles me - I am not even calling the networking code now?

soyersoyer:
From https://github.com/rsta2/circle/blob/master/doc/realtime.txt:
```
* If you rely on a small IRQ latency, USE_SDHOST should be disabled in include/
circle/sysconfig.h. WLAN access is not possible parallel to SD card access then
on Raspberry Pi 3 and Zero W.
```

If you disable the USE_SDHOST in build.sh?

soyersoyer:
Is your cable or headphones ok?:) Does the main branch work?

probonopd:
Still crackles. Running out of ideas...

Does anyone have any further ideas what might be causing the crackle even though Networking is disabled and largely commented out/removed from the code now?

soyersoyer:
And what about the latest CI build? MiniDexed_2025-04-20-5bc4895.zip

probonopd:
Does not crackle

soyersoyer:
maybe these libs can run something at startup (static stuffs, etc), but if you remove them you have to remove the other network stuff too.

$(CIRCLEHOME)/addon/wlan/hostap/wpa_supplicant/libwpa_supplicant.a \
$(CIRCLEHOME)/addon/wlan/libwlan.a \
$(CIRCLEHOME)/lib/net/libnet.a

diyelectromusic:
I see there are three network objects instantiated in MiniDexed, so the object constructors for these will be called regardless of any configuration settings:
	CBcm4343Device m_WLAN;
	CWPASupplicant m_WPASupplicant;
	CUDPMIDIDevice m_UDPMIDI;

So I'd start looking in those to see what they do.  I see that UDPMIDIDevice calls AddDevice ("udp") for example.

My experience of adding networking to embedded systems is that there are a lot of system level things that have to be done to maintain the network regardless of its use by application code.

But I've not looked into network support in circle to know if that applies here.

But historically, networking has not been particularly compatible with real time systems, but these days I'd imagine faster devices and multiple cores means that the responsivity to a network can be maintained with sensible use of tasks across cores. But again I don't know what this has implemented.

Is all networking on the same core as all the UI processing? If so, I believe that ought to be ok as I think all audio processing happens on cores 1,2,3.  But if there are no lengthy interrupt processing happening they will interrupt (as I understand things) all normal processing on cores.

If you Pi is not actually plugged into a network do you still get crackling?  I'm wondering if just having a UDP device enabled means it is spending time processing packets and rejecting them?  I don't know what background wireless processing would occur even if Wi-Fi isn't enabled... I guess we need to wade through those constructors to see what is actually going on...

Kevin

probonopd:
Yes, crackling even with network disabled, largely commented out, and not connected. 

> I see there are three network objects instantiated in MiniDexed, so the object constructors for these will be called regardless of any configuration settings:
CBcm4343Device m_WLAN;
CWPASupplicant m_WPASupplicant;
CUDPMIDIDevice m_UDPMIDI;

We need to find a way to instantiate these only if networking is actually enabled via the config file.

soyersoyer:
in the .h
```
CBcm4343Device *m_WLAN;
CWPASupplicant *m_WPASupplicant;
CUDPMIDIDevice *m_UDPMIDI;
```
in the constructor
```
m_WLAN(nullptr),
m_WPASupplicant(nullptr),
m_UDPMIDI (nullptr),
```
in the InitNetwork:
```
m_WLAN = new CBcm4343Device(WLANFirmwarePath);
m_WPASupplicant = new m_WPASupplicant(WLANConfigFile);
m_UDPMIDI = new CUDPMIDIDevice (this, m_pConfig, &m_UI);
```

but please read a C++ quickstart in the future at least, because this is really at the beginning :)

probonopd:
Tried that; also get the crackle.

At this point, I have spent the better part of the Easter holidays on this. And am out of ideas why I am getting crackling sound (that hadn't been there months ago - we should have merged it back then).

Maybe it's time to ask @omersiar to have a look at how to get his networking support cleanly folded into MiniDexed main.

soyersoyer:
Could you still try the main branch with this?
 
build.sh
```
cd ../wlan/
make clean || true
make -j
````
src/Rules.Mk:
```
	$(CIRCLEHOME)/lib/libcircle.a \
	$(CIRCLEHOME)/addon/wlan/hostap/wpa_supplicant/libwpa_supplicant.a \
	$(CIRCLEHOME)/addon/wlan/libwlan.a \
	$(CIRCLEHOME)/lib/net/libnet.a
```

diyelectromusic:
> At this point, I have spent the better part of the Easter holidays on this. And am out of ideas why I am getting crackling sound (that hadn't been there months ago - we should have merged it back then).

Just because this worked then and all the changes since then on their own haven't introduced an issue doesn't mean that the combination when properly integrated wouldn't have pushed the system into unstable territory.

Kevin

probonopd:
Yes, this is why I want to finally get this merged  - so that when another change causes issues with the networking, we can catch it immediately.

probonopd:
@soyersoyer 

> Could you still try the main branch with this?

Finally, we are getting somewhere... this produces crackle!

probonopd:
Turns out that merely adding one include path with `-I` in Rules.mk and _no code changes_ vs. `main` causes the crackle; see #867.

probonopd:
fatal: unable to access 'https://codeberg.org/dcoredump/Synth_Dexed.git/': The requested URL returned error: 502

Probably an intermittent error, so try again lager.

probonopd:
Consider to merge https://github.com/probonopd/MiniDexed/pull/864 instead of this one.

PR #852, opened 2025-04-19
use SoundFormatSigned24_32 with NEON

soyersoyer:


probonopd:
Does not crackle in my test on RPi Zero 2 W with

```
# Sound device
SoundDevice=i2s
#SoundDevice=pwm
#SoundDevice=hdmi
SampleRate=22000
#ChunkSize=256
DACI2CAddress=0x0
#DACI2CAddress=0x98
ChannelsSwapped=0
MasterVolume=32
```

I am a bit puzzled. More testing needed on more Raspberry Pis and in more configurations.

Banana71:
I've tested some frequencies with an Raspberry Pi 4, I²S, and EngineType=1.

SampleRate= 48000 Hz and 44100 Hz work.
My default performance (001:041 Baroque) works at 36000 Hz, but after changing bank to Greenhouse, the Raspberry Pi crashes.

If this helps with troubleshooting, I'll say it again: I had the same problem with older versions of miniDexed.

soyersoyer:
Do you see anything in the log? Does it work with another SD card? Could this be some kind of file system corruption?

probonopd:
> I had the same problem with older versions of miniDexed.

@Banana71 was there ever a version that did __not__ have this issue?

probonopd:
No crackling neither on PWM nor i2s.

soyersoyer:
If you squash the commits, you lose the point that they are separated in such a way that they can be easily understood. And if someone wants to see what changed in the future, it will be harder for them.
And if you haven't written any code for it, then you don't have to write co-authored-by either.

PR #853, opened 2025-04-19
move MIDI defines to midi.h

soyersoyer:
The MIDIi defines would also be needed in uibuttons.cpp.
Create a common midi.h with the definitions.

### DIFF
--- a/src/midi.h
+++ b/src/midi.h
@@ -1,0 +1,48 @@
+//
+// midi.h
+//
+// MiniDexed - Dexed FM synthesizer for bare metal Raspberry Pi
+// Copyright (C) 2025  The MiniDexed Team
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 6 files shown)


Banana71:
@soyersoyer 
What should I look out for when testing?

soyersoyer:
Oh sorry, I forgot to mention that this is just a code cleanup.

soyersoyer:
If you squash the commits, you lose the point that they are separated in such a way that they can be easily understood. And if someone wants to see what changed in the future, it will be harder for them.
And if you haven't written any code for it, then you don't have to write co-authored-by either.

probonopd:
> If you squash the commits, you lose the point that they are separated in such a way that they can be easily understood. And if someone wants to see what changed in the future, it will be harder for them.

If you look at my typical PRs, they contain lots and lots of individual commits, which not always really work. My workflow is that i use a PR as a means to keep working on something until it is ready to be merged, at which point I want only one entry in the history for the PR. I understand that other projects may have more discipline and handle this differently.

> And if you haven't written any code for it, then you don't have to write co-authored-by either.

I don't write this, GitHub (sometimes!) seems to do it automatically. I will pay more attention to this. The last thing I want is credit for other people's work ;-)

PR #855, opened 2025-04-20
Always show both arrows in the bank selector menu

soyersoyer:
It wraps around.


probonopd:
Hi @soyersoyer. At some point we should decide if we want to wrap around _all_ menus or not, what do you think?

soyersoyer:
It's fine the way it is, but where you can go around, there should be arrows.

probonopd:
Great point!

PR #858, opened 2025-04-20
Confirmation for Overwrite

probonopd:
Like for Delete

### DIFF
--- a/src/uimenu.cpp
+++ b/src/uimenu.cpp
@@ -992,23 +992,76 @@ void CUIMenu::EditOPParameter (CUIMenu *pUIMenu, TMenuEvent Event)
 
 void CUIMenu::SavePerformance (CUIMenu *pUIMenu, TMenuEvent Event)
 {
-	if (Event != MenuEventUpdate)
-	{
+	// Add confirmation before overwriting a performance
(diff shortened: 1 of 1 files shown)


soyersoyer:
For consistency, could bOverwriteMode and bConfirmOverwrite be private members?

probonopd:
Great point @soyersoyer. It should. Will change it once it is confirmed that functionality-wise this fulfills the requestor's requirement.

Hello @Octet-nl, would you like to test the build above? Thanks.

Banana71:
It doesn't work.

I can save a new performance. It automatically uses the next available slot for the new performance.
Overwriting doesn't work. Save - Overwrite - "Overwrite? the performance name is in the second line"  after clicking to confirm, the UI crashes. The sound still works, but the controls freeze.
No changes to the performance are saved either.
Peter

probonopd:
Argh. Will need to dig deeper. Thanks for testing though (currently away from my hardware).

Octet-nl:
Same for me. Back button still works and UI stays active. Voice functions still work, sound ok. Selecting Overwrite no longer works (does not show Overwrite? again). Performance is not saved.

PR #862, opened 2025-04-21
Build 32-bit and 64-bit in parallel

probonopd:
* Decreases build time
* Avoids mixing up 32-bit and 64-bit files (as was the case in #849)


soyersoyer:
Why is this good?

probonopd:
I wanted to check whether this introduces crackle - it doesn't.

Issue #846, opened 2025-04-19
Crackling sound


probonopd:
On RPi 3, this crackles in the main branch as of [0583421](https://github.com/probonopd/MiniDexed/commit/05834211383ae98e37c7a58f2ffec792e57d57af)

```
# Sound device
#SoundDevice=i2s
SoundDevice=pwm
#SoundDevice=hdmi
SampleRate=22000
#ChunkSize=256
DACI2CAddress=0x0
#DACI2CAddress=0x98
ChannelsSwapped=0
MasterVolume=32
```

The build in 
* https://github.com/probonopd/MiniDexed/pull/843
does not crackle.

We need to isolate the merge that caused the crackle to start.

probonopd:
6cbdab5f895a1810cd922d079a6c56d45a3fcf2b (Introduce MasterVolume, #842) also doesn't crackle.

So I suspect the crackle comes from #844, #845,

soyersoyer:
Could you try it without this commit?
https://github.com/probonopd/MiniDexed/pull/844/commits/c918b5985213340ed8b43e598fbf7fe378333e35

probonopd:
For now I have restored main to a non-crackling state. Let's add back changes slowly, testing each change with the config above. Thanks!

(I am right now working on merging the networking stuff and was surprised to hear crackle even though networking was not even switched on ;-))

Banana71:
Just curious.
Why are you using such a strange sample rate of 22000?
A common sample rate in that range would be 22050.

In direct comparison, 48000 sounds much better. Doesn't that work for me? Or why don't you set it to 48000?

Banana71:
I also wanted to mention here again, just like in issue #846, that my Raspberry Pi 4 crashes with a configured sample rate of 22000.

soyersoyer:
I don't notice any difference in the PWM output of my raspberry pi 3b+. Just as noisy. How is it crackling for you?

probonopd:
@soyersoyer have you used the settings I pasted above? Maybe something to do with the sample rate and/or in combination with the Master Volume? For me it crackled really bad to  the point of completely being unusable.

soyersoyer:
Yeah, I tried that too. It's quite strange.

probonopd:
So for you it is not crackling?

soyersoyer:
With my 3b+, no.

probonopd:
Thanks for your individual pull requests. It is strange - so far I don't get the crackle. anywhere..

PR #865, opened 2025-04-21
Test build with networking libs

probonopd:
**Do not merge, this is only a test!**

---

Test for https://github.com/probonopd/MiniDexed/pull/849#issuecomment-2818453197 to see whether this introduces crackle.

[ci skip]

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -45,10 +45,6 @@ cd libs/circle/addon/display/
 make clean || true
 make -j
 
-cd ../wlan/
-make clean || true
-make -j
-
 cd ../sensor/
 make clean || true
 make -j
--- a/build.sh
+++ b/build.sh
@@ -20,11 +20,6 @@ if [ "${RPI}" -gt "1" ]; then
     OPTIONS="${OPTIONS} -o ARM_ALLOW_MULTI_CORE"
 fi
 
-# For wireless access
-if [ "${RPI}" == "3" ]; then
-    OPTIONS="${OPTIONS} -o USE_SDHOST"
-fi
-
 # USB Vendor and Device ID for use with USB Gadget Mode
 source USBID.sh
 if [ "${USB_VID}" ] ; then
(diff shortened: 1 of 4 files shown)


probonopd:
Crackles!

soyersoyer:
I'm betting on libwpa_supplicant.a.

probonopd:
Also crackles with Circle downgraded and `libwpa_supplicant.a` removed.

diyelectromusic:
Sorry, this is probably a silly question, and I must confess that I'm just not able to follow what is changing when and causing what issue with what code, so I'm largely staying out of things to avoid confusing the issue, but...

... how does this relate (if at all) to https://github.com/probonopd/MiniDexed/issues/846?

Kevin

probonopd:
[MiniDexed_2025-04-21-fce23d9](https://github.com/probonopd/MiniDexed/actions/runs/14576194747/artifacts/2979871729) does NOT crackle.

probonopd:
> how does this relate (if at all) to #846?

Excellent question. Maybe not at all. Maybe in a way I don't understand yet. Lastly, there is a very slight possibility that I may have messed up something during testing 846.

probonopd:
2b5d80e crackles.
So I am coming to the conclusion that merely adding `-I $(NET_DIR)` causes the crackle.

probonopd:
Turns out that my testing may have been tainted:

I found out that when I open a downloaded zip file from my browser's download menu, then it shows a pre-existing zip window (showing in fact the wrong content) rather the one I intended to open.

Argh!

soyersoyer:
Adding the new include DIR would be strange because I think it searches them in order and if it finds something, it stops there.
Does it work just by adding libnet.a?

probonopd:
How so? `-I $(NET_DIR)/libnet.a`?

probonopd:
@soyersoyer solved this mystery in https://github.com/probonopd/MiniDexed/pull/867#issuecomment-2819056408. 💯

PR #870, opened 2025-04-21
fix op ordering and reenabling

soyersoyer:
The ops are in reverse order in the OPMask as well

If you change a voice, Dexed enables all operator.
Do the same in MiniDexed.


probonopd:
@Banana71 does this solve it for you?

Banana71:
Very good, now the operators are reactivated after a voice or performance change. Even after a voice MIDI dump via Dexed, deactivated operators become active again.

PR #883, opened 2025-04-23
Cleanup reverb state on reverb enable

soyersoyer:
Since doReverb was not called after reverb was turned off, so the cleanup was missed.

Call the cleanup explicitly when enabling reverb.

The cleanup() resets all values ​​associated with the previous run to zero.

This change also removes the unnecessary bypass variable, as doReverb is only called when it is enabled (m_nParameter[ParameterReverbEnable]).

### DIFF
--- a/src/effect_platervbstereo.cpp
+++ b/src/effect_platervbstereo.cpp
@@ -156,6 +156,62 @@ AudioEffectPlateReverb::AudioEffectPlateReverb(float32_t samplerate)
     reverb_level = 0.0f;
 }
 
+void AudioEffectPlateReverb::cleanup(void)
+{
+        memset(in_allp1_bufL, 0, sizeof(in_allp1_bufL));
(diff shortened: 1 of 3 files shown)


soyersoyer:
This needs some testing. Is this really useful?
Not all the noise I don't like about the performance change has gone away.

probonopd:
It seems to abruptly stop any still-running reverb tail when a new performance is loaded.

I think it gives a small "click" artefact due to the abrupt cutoff. Maybe we could use a very short fade out instead to avoid that?

probonopd:
It also seems to cut off any still-playing notes. Is that intended? I don't think e.g., the Montage does it that way ;-)

soyersoyer:
> It also seems to cut off any still-playing notes. Is that intended? I don't think e.g., the Montage does it that way ;-)

That's really not good then.

I saw on some big brand synths that you can set whether the existing sounds should continue to play when changing performance.

PR #884, opened 2025-04-23
Handle TX216/TX816 style performance sysex messages

### LINKED ISSUE #854: Receive SYSEX messages for Performance Data
MiniDexed currently does not respond to Performance SYSEX messages, what is important if you edit them remotely.

Performance data are using group=1 (2 bits shifted, thus 0x04)

e.g. Master tuning of module 1 (par=64 (40h), val=64 (32h) means 440Hz)
results in  { 0xF0, 0x43, 0x10, 0x04, 0x40, 0x32, 0xF7 };
Image

Image

probonopd:
Closes #854

Format for TX816/TX216 style performance SysEx messages:

| Byte | Value | Description                                              |
|------|-------|----------------------------------------------------------|
| 0    | F0    | SysEx start                                              |
| 1    | 43    | Yamaha manufacturer ID                                   |
| 2    | 10    | Channel number (0x10 = ch 1, 0x11 = ch 2, etc.)          |
| 3    | 04    | Group (0x04 = performance parameter change)              |
| 4    | p     | Parameter number (see table below)                       |
| 5    | val   | Value for parameter                                      |
| 6    | F7    | SysEx end                                                |

| p  | Parameter                     | val   | Notes                                           |
|----|-------------------------------|-------|-------------------------------------------------|
| 0  |                               |       |                                                 |
| 1  | Source Select                 | 1–16  | Selects MIDI receive channel 1 to 16            |
| 2  | Poly/Mono                     | 0–1   |                                                 |
| 3  | Pitch Bend Range              | 0–12  |                                                 |
| 4  | Pitch Bend Step               | 0–12  |                                                 |
| 5  | Portamento Time               | 0–99  |                                                 |
| 6  | Portamento/Glissando          | 0–1   |                                                 |
| 7  | Portamento Mode               | 0–1   | Sustain-key pitch 0: retain, 1: follow          |
| 8  |                               |       |                                                 |
| 9  | Modulation Wheel Sensitivity  | 0–15  |                                                 |
| 10 | Modulation Wheel Assign       | 0–7   | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias  |
| 11 | Foot Controller Sensitivity   | 0–15  |                                                 |
| 12 | Foot Controller Assign        | 0–7   | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias  |
| 13 | After Touch Sensitivity       | 0–15  |                                                 |
| 14 | After Touch Assign            | 0–7   | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias  |
| 15 | Breath Controller Sensitivity | 0–15  |                                                 |
| 16 | Breath Controller Assign      | 0–7   | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias  |
|    |                               |       |                                                 |
| 26 | Audio Output Level Attenuator | 0–7   |                                                 |
|    |                               |       |                                                 |
| 64 | Master Tuning                 | 0–127 | Concert Pitch at 64                             |


probonopd:
@skerjanc could you please test this version regarding performance sysex? Thanks.

probonopd:
I can set detune, poly/mono, MIDI channels with this. Please let me know if it is working for you, then I will merge this into the main branch. Thanks!

Skerjanc:
I tested the build from yesterday:
+ Mono/Poly: ok
+Portamento Time ok
+ PB Range: ok
+PB Step ok
- Master tuning works, but with some quirks. Its range is about +/- 100 cent, while TX816 goes +/-75 cent. Also at the value 0 tuning jumps back to 64.
- Glissando on/off does strange things. Tone gets totally cross modulated.
- OMNI message is no more working.

probonopd:
Thanks for your patience and feedback @Skerjanc. I really appreciate it!

> Master tuning works, but with some quirks. Its range is about +/- 100 cent, while TX816 goes +/-75 cent. Also at the value 0 tuning jumps back to 64.

Changed this now as per your feedback.

> Glissando on/off does strange things. Tone gets totally cross modulated.

Which exact sysex are you sending that doesn't behave like it should? `F0 43 10 04 07 01 F7`?

-- | -- | -- | --
-- | -- | -- | --
5 | Portamento Time | 0–99 |  
6 | Portamento/Glissando | 0–1 |  
7 | Portamento Mode | 0–1 | Sustain-key pitch 0: retain, 1: follow

> OMNI message is no more working

That feature is not part of this pull request but was developed in another pull request. I have now updated this branch so that it also contains the latest other feature that were recently added.

Skerjanc:
Would be great, if these settings were recognized by SYSEX as well, as these are important performance utilities. 
Currently all messages for xyAssign > 4 result in silencing the output.

9 | Modulation Wheel Sensitivity | 0–15 |  
10 | Modulation Wheel Assign | 0–7 | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias
11 | Foot Controller Sensitivity | 0–15 |  
12 | Foot Controller Assign | 0–7 | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias
13 | After Touch Sensitivity | 0–15 |  
14 | After Touch Assign | 0–7 | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias
15 | Breath Controller Sensitivity | 0–15 |  
16 | Breath Controller Assign | 0–7 | Bit 0: Pitch, bit 1: Amplitude, bit 2: EG bias

probonopd:
> Glissando SYSEX catches the right function, but the glissando effect seems to affect all operators differently, thus it sounds detuned in steps. An issue of Dexed also?

I think that is a separate topic that should be handled in a separate issue. I opened this one for it:

* https://github.com/probonopd/MiniDexed/issues/889

I am looking into the Assign issue.

probonopd:
If my understanding is correct, then the values for the Assign should be:

- Bit 0: Pitch modulation (vibrato or pitch bend)
- Bit 1: Amplitude modulation (tremolo or volume change)
- Bit 2: EG (Envelope Generator) bias (shifts the envelope, affecting timbre or dynamics)

Value | Binary | Pitch | Amplitude | EG Bias
-- | -- | -- | -- | --
0 | 000 | Off | Off | Off
1 | 001 | On | Off | Off
2 | 010 | Off | On | Off
3 | 011 | On | On | Off
4 | 100 | Off | Off | On
5 | 101 | On | Off | On
6 | 110 | Off | On | On
7 | 111 | On | On | On

Are you saying that up to a value of 4 this works correctly, but for values 5-7 it doesn't?

probonopd:
I think this PR is now routing TX216/TX816 style performance sysex messages to the Synth_Dexed engine correctly.
Whether the Synth_Dexed engine interprets them correctly (produces the correct sounds) is a different question. But unless someone can spot mistakes in the implementation of this pull request, I'd like to merge it.

probonopd:
Is the 3rd byte (byte 2) supposed to mean the MIDI channel or the physical slot in the TX816 in which the module is inserted?

Currently we are treating it as the MIDI channel. But maybe it would make more sense to treat it as the Tone Generator (TG; "module") number in out case?

Skerjanc:
found the culprit for the mutes at modulation settings. SYSEX works correctly, but the range of TX816 is different.
Could you remap the modulation range from [0..15] to [0..99]?

Skerjanc:
> If I am not mistaken, on the real TX816 there is a common MIDI in jack and a per-module MIDI in jack. Whereas we don't have that luxury. This might make it seem useful to interpret the 3rd byte as the physical slot in the TX816 (a.k.a. Tone Generator, TG in MiniDexed) rather than as the MIDI channel, at least for the Source Select message - even if this would mean that we deviate a little bit from the original.

It has a common MIDI in that is routed to all modules. Each module also has a separate MIDI in and may be toggled between common/individual MIDI in. I guess, nobody today is using 8 MIDI cables, but only the common one. In that case, all modules will be set-up to MIDI Ch 1..8.

probonopd:
So you are saying what is incoming for Sensitivity parameters is 0-15 (as in the table), but the DX7 (and hence probably the Synth_Dexed engine) expects 0-99. So we need to scale the 0-15 range to 0-99?

I found a comment in the code that I don't fully understand but makes me wonder... is there a bug in Synth_Dexed?

https://github.com/probonopd/MiniDexed/blob/988f08c36da61f758c9d0bf5cf09dca16c923c80/src/minidexed.cpp#L1686

Skerjanc:
> So you are saying what is incoming for Sensitivity parameters is 0-15 (as in the table), but the DX7 (and hence probably the Synth_Dexed engine) expects 0-99. So we need to scale the 0-15 range to 0-99?

Yes, that is the difference between DX7 (and TX816) first version and DX7ii Mark 2. You have to decide if you want to simulate the Mark 1 behaviour (what I hope) or not.

Skerjanc:
Tested and working with sensivity15. Also, in OMNI mode as well.

probonopd:
Thanks for testing. So the only thing that is not working as it should is Glissando now?

Skerjanc:
3 things left:

1. Glissando on directly influences the detune parameter, i.e. they are no longer usable then
2. Master Tuning range is now twice as high than TX816 (+/- 150ct instead of +/+75ct)
3. If you want, you can also implement Output level [0..7]

probonopd:
- Is 1. also true when you are not using Performance Sysex but the menu? (Then that is a bug unrelated to the Performance Sysex handling.)
- Hopefully I have fixed 2. and 3. (new build should appear here shortly)

Skerjanc:
Tested and working with sensivity15. Also in OMNI mode as well. Wonderful.

> * Is 1. also true when you are not using Performance Sysex but the menu? (Then that is a bug unrelated to the Performance Sysex handling.)

Yes it is.

probonopd:
Glissando bug will be handled here:

* https://github.com/probonopd/MiniDexed/issues/889

Skerjanc:
1. Sysex for Performance Output Level: at 7, extreme loud, but Values 0..6 is only silence. Further, it changes the master volume of all modules, not the individual volume.
2. Master tuning is ok.

Skerjanc:
Output Level now works on each individual channel.
But there is a small difference to TX816, where the volume changes exponentially, ie. 
7 full level
6 half level
5 quarter level
etc.

probonopd:
@Skerjanc: Thank you very much. This kind of feedback is priceless for us who don't have access to real hardware to compare with. Thanks a lot!

probonopd:
So if everything went well, then in the latest build above the fix for the Glissando/Portamento detune issue should be working also in conjunction with MIDI Performance SysEx.

Remaining thing to fix before this is ready to be merged:
Scaling of the per-TG volume.

Skerjanc:
> Build for testing:
> [MiniDexed_1145_2025-04-29-fd7786d_32bit](https://github.com/probonopd/MiniDexed/suites/37868405457/artifacts/3033103803)
> [MiniDexed_1145_2025-04-29-fd7786d_64bit](https://github.com/probonopd/MiniDexed/suites/37868405457/artifacts/3033110917)

Test passed
 :-)

probonopd:
> Output Level now works on each individual channel.
But there is a small difference to TX816, where the volume changes exponentially, ie.
7 full level
6 half level
5 quarter level

Are you _sure_ it should be exponential on the TX816?

I am asking because hearing loudness is logarithmic, not exponential.

Skerjanc:
> Build for testing:
> 
> **Exponential** (a Logarithmic version is further below for comparison):
> 
> [MiniDexed_1148_2025-05-01-873d8a8_32bit](https://github.com/probonopd/MiniDexed/suites/37943392416/artifacts/3042987196) [MiniDexed_1148_2025-05-01-873d8a8_64bit](https://github.com/probonopd/MiniDexed/suites/37943392416/artifacts/3042989781) Use at your own risk.

Is the correct one, perfect. Only difference is at 0, in your version it is totally quiet, but audible on the TF1 (same as set to 1. I guess this is a bug in the Yamaha HW)

Skerjanc:
And hey, operator mute messages are now working!! Test passed.
 Did I miss this change in the last builds?

probonopd:
@Skerjanc, thank you so much, this is why it is so awesome to have someone with a real TX816. I would never have assumed that exponential was correct and would never have known that 0 should result in the same volume as 1 (I thought they did this to avoid unintentional silence). There may be other areas in MiniDexed (e.g., all the other ways in which you can set volumes on modules = Tone Generators, and on operators) where we may have the mapping wrong. If you discover any discrepancy between the real thing and MiniDexed, please let us know. These parameter changes are easy to fix once we know how they should be.

> And hey, operator mute messages are now working!! Test passed.
Did I miss this change in the last builds?

I had worked on this fix in 

* https://github.com/probonopd/MiniDexed/pull/896

and it has been merged to the main branch recently.

probonopd:
Merged. Please let me know if further changes are needed.

Skerjanc:
**Portamento Memory** is implemented wrong: currently it only enables/disables portamento. 
But on the HW it does the following (**only in Monophonic Mode**):
- Off: Portamento is disabled when two notes are played non-legato. Played legato, portamento is enabled
- On: always portamento

Another deviation: **Pitch Bend Step**:
The quantisation range is too low. E.g. set PB to 12 and PBstep to 1, you shall hear a chromatic glissando until the full octave. 
But is does stop  at the 11th note. This error propagates also with higher PBstep values.

probonopd:
This kind of feedback is so extremely valuable @Skerjanc. Thank you very much. I have opened separate issues for each bug:

* https://github.com/probonopd/MiniDexed/issues/916
* https://github.com/probonopd/MiniDexed/issues/917

Please feel free to open more such issues for any deviation you discover. Thanks!

PR #890, opened 2025-04-25
Performance first menu, PerformanceSelectChannel Omni by default

### LINKED ISSUE #384: Initial brass sound stays throughout all patches

probonopd:
Closes #384, #860

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -199,7 +199,7 @@ void CConfig::Load (void)
 	m_bMIDIDumpEnabled  = m_Properties.GetNumber ("MIDIDumpEnabled", 0) != 0;
 	m_bProfileEnabled = m_Properties.GetNumber ("ProfileEnabled", 0) != 0;
 	m_bPerformanceSelectToLoad = m_Properties.GetNumber ("PerformanceSelectToLoad", 0) != 0;
-	m_bPerformanceSelectChannel = m_Properties.GetNumber ("PerformanceSelectChannel", 0);
+	m_bPerformanceSelectChannel = m_Properties.GetNumber ("PerformanceSelectChannel", 17);
 	
 	// Network
 	m_bNetworkEnabled  = m_Properties.GetNumber ("NetworkEnabled", 0) != 0;
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -2322,7 +2322,7 @@ void CMiniDexed::UpdateNetwork()
 			LOGNOTE("FTP daemon not started (NetworkFTPEnabled=0)");
 		}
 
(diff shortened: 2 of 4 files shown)


diyelectromusic:
As commented here - https://github.com/probonopd/MiniDexed/discussions/860#discussioncomment-12951100, as you might expect from our previous conversations, I don't agree this is the approach to take as it doesn't actually solve the usability issue - it just hides it.

Kevin

probonopd:
I hear your opinion -- but would it actually make anything worse?

probonopd:
I believe your concerns need to be addressed by something else than the question which menu item shows up after booting. See my ideas (here](https://github.com/probonopd/MiniDexed/discussions/860#discussioncomment-12951350).

diyelectromusic:
> I believe your concerns need to be addressed by something else than the question which menu item shows up after booting

Yes, absolutely - that is what I keep saying!

I have no strong opinion on what the default is.  Just be sure to make sure the "home" action still works properly for multi-TG devices (PI 2+) and single-TG devices (Pi 1) and that the optional "up to 16 TGs" still works fine on a Pi 4/5.

To really satisfy what I think you're aiming for here, should the home now be "performance -> load" not just "performance" so users see a performance name and the first actions allow it to change?  But then it might not be obvious how to get "back" and arguably going "back" from what now seems to be the "home" screen is a little weird...

There is also an argument that the most useful starting menu might now be the master volume... :)

(I did wonder about allowing a range of options for "home menu" and "startup menu" but haven't looked into it in detail yet)

Kevin

probonopd:
> going "back" from what now seems to be the "home" screen is a little weird...

Indeed, the synth should not start up somewhere in the middle of the menu tree ;-)

PR #892, opened 2025-04-27
Implement Unison Mode

### LINKED ISSUE #315: Unison mode

probonopd:
Closes #315

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -31,6 +31,9 @@
 #include <assert.h>
 #include "arm_float_to_q23.h"
 
+// Forward declaration for getPhysicalTG
+static unsigned getPhysicalTG(unsigned logicalTG, unsigned unisonVoice, unsigned unisonVoices);
+
 const char WLANFirmwarePath[] = "SD:firmware/";
 const char WLANConfigFile[]   = "SD:wpa_supplicant.conf";
 #define FTPUSERNAME "admin"
@@ -126,6 +129,9 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 			m_pTG[i]->setEngineType(pConfig->GetEngineType ());
 			m_pTG[i]->activate ();
 		}
+		m_nUnisonVoices[i] = 1;
+		m_nUnisonDetune[i] = 0;
+		m_nUnisonSpread[i] = 0;
 	}
 
 	unsigned nUSBGadgetPin = pConfig->GetUSBGadgetPin();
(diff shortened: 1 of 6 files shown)


probonopd:
Can't wait to see what @Banana71 thinks about this. (Note: The meaning of the numbers in `UnisonDetuneN` and `UnisonSpreadN` is still subject to change, so don't create performances using this feature yet.)

probonopd:
Absolutely no clue why the 64-bit version fails to fetch the firmware now (the 32-bit version doesn't have the feature anyway). Maybe an intermittent issue.

Local build seems to work rather well.

probonopd:
Something is still wrong, we are not really getting 32 physical tone generators. So for one TG (e.g., if there is only TG1 in the performance) it seems to work well, but if there are also TG2...TG8 in use then things get mixed up. I am looking into it.

probonopd:
The issue is that not enough TGs get created at boot time (e.g., for 8 logical TGs, 32 physical ones need to get created, because we can have up to 4 voices per logical TG).

Banana71:
Unison would be a great feature.
I don't quite understand how you're trying to solve it.
Is this how it's supposed to work?:
If TG1 Unison Voices is set to 4, TG1 and 3 new sub TGs are set to the same value (VoiceData, Volume, ReverbSend, etc.), **while all other TGs remain TG2 to TG8**. Detune and PAN are set according to the Detune and Spread parameters.



What else I noticed: In Main and this version, the Performance Bank name is no longer transferred from the folder name. (001_default - 001-Seed)

Banana71:
Is it possible that for the Sub TGs the voice data is not taken from the performance but voice is loaded from the SD card (voice number and bank number)?

Banana71:
Detune seems far too strong; the gradations should be much finer.

probonopd:
This is how it is __supposed__ to work:

For each TG visible to the user ("logical TGs"), 4 physical TGs get created. Depending on how many voices the user chooses to use, one to four are used when the logical TG plays notes.

In the performance, for each logical TG, the voice data should be saved (like it always was). But also the union related parameters (number of voices, unison detune, and unison spread) shall get saved.

Banana71:
The sub-TGs are visible; I can see 32 TGs.
My thoughts:
My Raspberry Pi 3 Model A is overloaded if I activate Unison twice. To avoid overloading the Raspberry Pi, perhaps you should limit the polyphony and/or the maximum number of sub-TGs. For example, for drum sounds, `Unison-Voices=2` with a maximum polyphony of 4 voices would be sufficient.

probonopd:
> The sub-TGs are visible; I can see 32 TGs.

Where?

I don't think all 32 are being created, and this is why we are seeing some strange effects going on as soon as one enables more than just TG1. In other words, it seems to behave like I envisioned it if I set MIDI channel OFF for TG2-TG8 for now. I think we need to create the correct number (32) of physical TGs for TG2-TG8 to also work. Once that is done, we can start to think about limiting CPU load.

Banana71:
I can actually see the 32 TGs in the miniDexed UI; I took a photo of it yesterday.
32TGs

RPi 3 Model A

probonopd:
Argh. I think the whole codebase mixes up logical TGs (what the user sees), auxiliary TGs (used if there is more than one voice in unsion), and physical TGs (meaning both logical and auxiliary TGs). For example, currently when saving performances, all physical TGs get saved whereas only the logical ones should. I will look into it.

Issue #894, opened 2025-04-27
Performance Bank organization is not consistent with the Voice Banks


probonopd:
### Discussed in https://github.com/probonopd/MiniDexed/discussions/893

<div type='discussions-op-text'>

<sup>Originally posted by **Banana71** April 27, 2025</sup>
Why don't the default performances have their own folder?
Couldn't the folder structure for performances look like this? 
Or would that be too complex to implement?
It bothers me more than it should. :boar:

![Screenshot 2025-04-27 170039](https://github.com/user-attachments/assets/fb876452-175e-4e1f-a778-21bc0790467a)

I know the whole thing has grown and gotten bigger over time. The default performance.ini file in the root directory should be considered, and the first performance file in the 001 folder should not be used.

Peter :raising_hand:
</div>

probonopd:
Closed by [#895](https://github.com/probonopd/MiniDexed/pull/895)

PR #895, opened 2025-04-27
BREAKING CHANGE: Load default performances from performance/001_Default

probonopd:
rather than performance/

Closes https://github.com/probonopd/MiniDexed/discussions/893, https://github.com/probonopd/MiniDexed/issues/894

Users need to create a `performance/001_Default/` directory and move `performance/*.ini` there.

### DIFF
--- a/src/performanceconfig.cpp
+++ b/src/performanceconfig.cpp
@@ -1290,12 +1290,6 @@ std::string CPerformanceConfig::AddPerformanceBankDirName(unsigned nBankID)
 	{
 		// Performance Banks directories in format "001_Bank Name"
 		std::string Index;
-		if (nBankID == 0)
-		{
-			// Legacy: Bank 1 is the default performance directory
-			return "";
-		}
-
 		if (nBankID < 9)
 		{
 			Index = "00" + std::to_string(nBankID+1);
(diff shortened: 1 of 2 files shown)


diyelectromusic:
It's been a while since I last looked a performances, so could be mistaken, but I wonder if you're moving the default performance, then there are quite a few special cases around performanceconfig.cpp for the handling of DEFAULT_PERFORMANCE_FILENAME that I'm sure would need fiddling with too...?

Also, probably need to consider that in ListPerformances, the default performance name is assumed to always exist even if no directory is present, so the performance subsystem will always work in some capacity.  There are a number of assumptions in how it all works based on m_bPerformanceDirectoryExists that probably need to be re-examined if that all changes.

There are a number of calls into the performance system that would fail if there is no performance directory and that might need some extra error checking in minidexed.cpp too.

I'm also trying to remember what happens to performances in the /performances/* area that aren't in banks... I know there was lots of special cases.

I seem to have had concerns when this was last raised too (and then I was buried deep in performance code so would have seen a lot more of the issues that I can't remember right now).  See https://github.com/probonopd/MiniDexed/pull/581#issuecomment-1872561868 - in short, I think there is a quite a bit more to it if we were to remove the legacy behaviours around performances I'm afraid...

Kevin

diyelectromusic:
Ok, just seen the comments in the discussion.

Changing the default directory and leaving the original default performance.ini in the root is a lot safer change :)

Kevin

probonopd:
@diyelectromusic any objections to merging this?

diyelectromusic:
I've not looked in detail, and its been a while since thinking about performances, but I think this has probably caught most of the issues and it seems to have been tested ok thanks to you both...

So probably ok if this is how you prefer things? :)

Kevin

probonopd:
@Banana71 please ping me as soon as Soundplantage is changed, then I will merge this. Thanks!

Issue #889, opened 2025-04-25
Glissando effect seems to affect all operators differently


probonopd:
@skerjanc [reported](https://github.com/probonopd/MiniDexed/pull/884#issuecomment-2828982690):

> Glissando SYSEX catches the right function, but the glissando effect seems to affect all operators differently, thus it sounds detuned in steps. An issue of Dexed also?

probonopd:
I think we need a way of testing this on MiniDexed and on real hardware for comparison.

Would you be able to run the script below with MiniDexed and module 1 of your TX816 for comparison? Do you hear any differences between the two? (I know it is a bit much to ask, but if you hear differences, could you record them for comparison?)

```
#!/usr/bin/env python
# -*- coding: utf-8 -*- 

# Test MIDI portamento and glissando features of T816

import time
import sys
import signal
import rtmidi

def ask(question, play_fn=None):
    while True:
        reply = input(f"{question} (y/n/r): ").strip().lower()
        if reply == 'r':
            if play_fn:
                play_fn()
            continue
        if reply in ['y', 'n']:
            return reply == 'y'

def send_sysex(midiout, msg, label=""):
    # Force all SysEx messages to MIDI channel 0 (0x10)
    if len(msg) > 2 and msg[0] == 0xF0 and msg[1] == 0x43:
        msg = msg.copy()
        msg[2] = 0x10
    midiout.send_message(msg)
    print(f"Sent: {label}")

def send_parameter(midiout, parameter, value, label):
    """Helper function to send SysEx for a single parameter."""
    msg = [0xF0, 0x43, 0x10, 0x04, parameter, value, 0xF7]
    send_sysex(midiout, msg, label)

def note_on(midiout, note, velocity=100, channel=0):
    midiout.send_message([0x90 | 0, note, velocity])

def note_off(midiout, note, channel=0):
    midiout.send_message([0x80 | 0, note, 0])

def play_legato(midiout, notes, delay=0.5, channel=0):
    midiout.send_message([0x90 | 0, notes[0], 100])
    time.sleep(delay)
    midiout.send_message([0x90 | 0, notes[1], 100])
    time.sleep(delay)
    midiout.send_message([0x80 | 0, notes[0], 0])
    midiout.send_message([0x80 | 0, notes[1], 0])

def play_sequential(midiout, notes, delay=0.5, channel=0):
    note_on(midiout, notes[0], channel=0)
    time.sleep(delay)
    note_off(midiout, notes[0], channel=0)
[...]

PR #897, opened 2025-04-29
Patch Synth_Dexed to fix glissando/portamento detune

probonopd:
Patch Synth_Dexed until the fix is merged upstream; see https://github.com/probonopd/MiniDexed/issues/889

I don't know whether this is the best possible solution but it does seem to improve the situation for me.


probonopd:
@Skerjanc does this solve the issue for you? Thanks.

probonopd:
Thanks for confirming @Skerjanc.

I will now merge this to the main branch and then update the Pull Request that deals with MIDI performance changes, so that we can test both in combination.

soyersoyer:
It would be nice to fix this in the Synth_Dexed repo.

PR #910, opened 2025-05-03
Fix click/Pop sound on startup

### LINKED ISSUE #908: Click/Pop sound on startup

probonopd:
Closes #908

As suggested by @soyersoyer in https://github.com/probonopd/MiniDexed/issues/908#issuecomment-2848657586

Thanks @soyersoyer

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1414,6 +1414,8 @@ void CMiniDexed::ProcessSound (void)
 				// END create SampleBuffer for holding audio data
 
 				// get the mix of all TGs
+				arm_fill_f32(0.0f, SampleBuffer[indexL], nFrames);
+				arm_fill_f32(0.0f, SampleBuffer[indexR], nFrames);
 				tg_mixer->getMix(SampleBuffer[indexL], SampleBuffer[indexR]);
 
 				// BEGIN adding reverb


Octet-nl:
Works! Thanks!

probonopd:
Thing is, I still hear _something_. You don't?

Octet-nl:
No, nothing. Amp at full volume, total silent startup, master volume=64.
What do you hear? The same click/pop?

Banana71:
The nasty, high-pitched, one-sided click is gone.
I can still hear a very faint double click. Absolutely tolerable, not a problem in my opinion. My SY99 is louder when turned on.

Many thanks to everyone and, in this case, specifically to @soyersoyer

probonopd:
The remaining double-click (on the PCM5102A) is most likely coming from the hardware being powered up and is a separate issue. I think @soyersoyer had started to look into that one as well, it may or may not be possible to work around that in software. Possibly it would help to use the mute hardware pin and mute the device during power up? (That would require a wiring change in hardware.)

@soyersoyer do you think we should apply this fix, or are you working on another one?

soyersoyer:
> @soyersoyer do you think we should apply this fix, or are you working on another one?

I'm working on another one, but keep this one open until then.

soyersoyer:
superseded by #912

PR #912, opened 2025-05-04
limit the number of processed frames to chunksize / 2

### LINKED ISSUE #908: Click/Pop sound on startup

soyersoyer:
The tg_mixer cannot process more franes,
So if it has to process more, uninitialized frames will be played, which causes strange sounds, e.g. at startup.

Closes #908

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -1305,6 +1305,10 @@ void CMiniDexed::ProcessSound (void)
 	unsigned nFrames = m_nQueueSizeFrames - m_pSoundDevice->GetQueueFramesAvail ();
 	if (nFrames >= m_nQueueSizeFrames/2)
 	{
+		// only process the minimum number of frames (== chunksize / 2)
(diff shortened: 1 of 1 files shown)


Octet-nl:
Hi,

It works, startup is silent now.

Thanks for your fast response and perfect fix!

Kind regards,

Hans.

PR #923, opened 2025-05-19
MIDI CC 07 (Volume) on PCCH sets Master Volume

probonopd:
Closes #920

- When a MIDI CC 07 message is received on the PCCH (if enabled), it sets the global Master Volume (`setMasterVolume`), allowing direct control from a MIDI keyboard slider.
- When PCCH is off, normal per-TG volume operation is preserved.
- The change is implemented in mididevice.cpp and includes a log message for traceability.

**Code excerpt:**
```cpp
else if (pMessage[1] == MIDI_CC_VOLUME) // CC 07 on PCCH: Master Volume
{
    float fMasterVolume = (float)pMessage[2] / 127.0f;
    m_pSynthesizer->setMasterVolume(fMasterVolume);
    LOGNOTE("MIDI CC07 (Volume) on PCCH: Set Master Volume to %d (%.3f)", pMessage[2], fMasterVolume);
    break; // Do not process further for TGs
}
```

**Testing:**  
- Send MIDI CC 07 on the PCCH: Master Volume changes.
- Send MIDI CC 07 on other channels: Only TG volume changes.

### DIFF
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -284,7 +284,7 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 						float fMasterVolume = (float)pMessage[2] / 127.0f;
 						m_pSynthesizer->setMasterVolume(fMasterVolume);
 						LOGNOTE("MIDI CC07 (Volume) on PCCH: Set Master Volume to %d (%.3f)", pMessage[2], fMasterVolume);
-						break; // Do not process further for TGs
+						return; // Do not process further for TGs
 					}
 					else
 					{
--- a/src/mididevice.cpp
+++ b/src/mididevice.cpp
@@ -279,12 +279,12 @@ void CMIDIDevice::MIDIMessageHandler (const u8 *pMessage, size_t nLength, unsign
 					{
 						m_pSynthesizer->BankSelectLSBPerformance (pMessage[2]);
 					}
-					else if (pMessage[1] == MIDI_CC_VOLUME) // CC 07 on PCCH: Master Volume
+					else if (pMessage[1] == MIDI_CC_VOLUME && nLength == 3) // CC 07 on PCCH: Master Volume
 					{
-						float fMasterVolume = (float)pMessage[2] / 127.0f;
-						m_pSynthesizer->setMasterVolume(fMasterVolume);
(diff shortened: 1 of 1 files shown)


probonopd:
image

Banana71:
I tested:
- :white_check_mark:The master volume changes when PCCH = MIDI channels of MIDI CC 07. 
- :x:The volumes of the TGs whose MIDI channel is PCCH also change, which should then be suppressed. 
- :white_check_mark:With different MIDI channels, the master volume does not change, but only the TGs that have the same MIDI channel as the incoming MIDI CC 07 command.

Banana71:
New Test:

- ❌The Raspberry crashes when changing the volume via MIDI CC 07 Channel = PCCH Channel.
- ✅With different MIDI channels, the master volume does not change, but only the TGs that have the same MIDI channel as the incoming MIDI CC 07 command.

Banana71:
No, the Raspberry still crashes

probonopd:
Now that is interesting, because mine doesn't seem to. Can you describe exactly what you are doing?

Banana71:
> Can you describe exactly what you are doing?

Raspberry Pi 4
- PCCH=Off MIDI CC 07 works as always, and all TGs with the same MIDI channel number as the MIDI CC command change the volume.
- PCCH=2 and volume changes via MIDI CC 07 on channel 1 affect the volume of TGs with channel number 1.
- PCCH=1 and volume changes via MIDI CC 07 on channel 1 cause the Raspberry Pi to crash immediately.

probonopd:
Ah, the issue might be that you are setting the PCCH to a channel on which you also have a TG listening, whereas I was using channel 16 as PCCH (which have no TG set to). Will look into it, thanks for your patience.

SianthDK:
Hi - thanks for all the great work here - I am struggling to get this to work, but perhaps my perception of what should be working is wrong - I hope someone could guide me a little with this.
I understand this to be exactly what I want, i.e. when CC 07 is adjusted the main volume (for all TG's) is adjusted accordingly, so I have PerformanceSelectChannel=2 in .ini, but when I send CC07 on e.g. channel 1 the main volume is changed on my mini Dexed even though I've set PerformanceSelectChannel to 2.
I have also tried with PerformanceSelectChannel=16 and still the main volume on my Mini Dexed is changed even though I'm sending on MIDI channel 1.
Performance change is perfect on the given channel and is not affected by values on other channels
Raspberry Pi 4 with Pirate Audio line out DAC/Display
Have I totally misunderstood this or is this not yet implemented on this otherwise fantastic job?

PR #945, opened 2025-07-07
use the same style for MIDI channel selection as for master volume

soyersoyer:
This is just a style update, and so there is no need to call unlock separately.
But is this code good? According to the manual, wouldn't the 3rd byte be the new channel?
image
image


probonopd:
> But is this code good? According to the manual, wouldn't the 3rd byte be the new channel?

Not entirely sure I understand your question fully, but

> we interpret the device parameter as the number of the TG (unlike the TX816/TX216 which has a hardware switch to select the TG)

PR #953, opened 2025-07-09
mixer: fix pan channel order

soyersoyer:
0.0 full left
1.0 full right

## Summary by Sourcery

Fix stereo panorama channel order and remove temporary HDMI channel swap workaround

Bug Fixes:
- Swap sine and cosine calculations to correct left/right pan channel ordering

Chores:
- Remove HDMI sound driver channel swap inversion workaround

### DIFF
--- a/src/effect_mixer.hpp
+++ b/src/effect_mixer.hpp
@@ -113,8 +113,8 @@ public:
 			pan = MIN_PANORAMA;
 
 		// From: https://stackoverflow.com/questions/67062207/how-to-pan-audio-sample-data-naturally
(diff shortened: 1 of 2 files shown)


probonopd:
@soyersoyer I don't quite understand how this fixes the swapped channels - or are they not swapped for you?

soyersoyer:
First I noticed that pan works the other way around.
I looked at this code afterwards, and the left channel is calculated with the sine, which is 0 at 0, and there should be 1 there, that would be the cosine. So it's basically wrong, and the stackoverflow comment above it also says it the other way around.

https://stackoverflow.com/a/67064448

And after that the pan works fine.

Don't you have the two sides swapped?

probonopd:
@sourcery-ai review

PR #955, opened 2025-07-11
[WIP] Noiseless parameter change

soyersoyer:
It would be nice to be able to change any parameter while playing without any pops and clicks.

In principle, it is possible if the parameter change occurs at zero cross.

I wrote zero cross detection for panning, it works.
However, some performances have a DC component, and there is a bit of noise with those.

I will change the other parameters as well.
Until then, you can try this out :-)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,15 +64,11 @@ jobs:
         run: |
           set -ex
           export PATH=$(readlink -f ./gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/bin):$PATH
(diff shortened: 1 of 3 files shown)


soyersoyer:
I'll try it out to see what happens if the parameter doesn't change immediately, but continuously.

Why can't you test the HDMI output?

probonopd:
Wondering if this could be done on a Syth_Dexed level for all parameters...

soyersoyer:
I tried doing it for cutoff and resonance, but it didn't work this way. (or I overlooked something)

PR #959, opened 2025-07-19
Revert "Use WLAN firmware from zip for now"

soyersoyer:
This reverts commit eb3a9273b4bc58d32c30abf07f91799de055fd39.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,10 +83,10 @@ jobs:
           mkdir -p ./sdcard/hardware/
           cp -r ./hwconfig/minidexed_* ./sdcard/minidexed.ini ./sdcard/hardware/
           # WLAN firmware
-          # mkdir -p sdcard/firmware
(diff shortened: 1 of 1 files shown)


soyersoyer:
Yes.

But I took a closer look at Wi-Fi support, I would rather not use it due to the security flaws in [hostapd](https://github.com/rsta2/circle/issues/601) and the Broadcom [firmwares](https://github.com/rsta2/circle/issues/602), and I think it shouldn't be included in the release for now, because it might end up in a product that its owners think doesn't need to be updated.
Or they need to be warned not to use it.

probonopd:
Thank you very much for looking into this @soyersoyer.
I have added a note on https://github.com/probonopd/MiniDexed/wiki/Networking.

soyersoyer:
Ok, thanks. I think the Broadcom firmwares are also important, because otherwise they wouldn't have been released.

probonopd:
Yes. Let's update both once they are available. Feel free to edit the wiki. Thank you!

PR #964, opened 2025-07-24
use x^4 for mastervolume

soyersoyer:
the effect_mixer also uses this
see: https://www.dr-lex.be/info-stuff/volumecontrols.html#ideal2


probonopd:
Will this change the sound? Then we should keep @banana71 in the loop

soyersoyer:
Only the MasterVolume 0-127 -> volume scaler mapping.

Banana71:
Why did you change the master volume scale in this direction?

The MiniDexed isn't a clock radio, which needs many small quiet levels. Below -30 dB, the noise-to-signal ratio is no longer recommended.

Is there a reason for this that I don't understand?

MasterVolume Scale

soyersoyer:
Ah sorry, I should have measured it myself and it would have been clear.
The AudioMixer [gain](https://github.com/probonopd/MiniDexed/blob/main/src/effect_mixer.hpp#L55) confused me, and I thought it was used everywhere, but it is not.
The TG volume/expression is linear.
The MasterVolume is x^2
The ReverbSend is x^4 (only this uses the AudioMixer gain)

Thanks!

probonopd:
Thanks to both of you. Benefits of community development :-)

PR #965, opened 2025-07-25
update the UI in setMasterVolume()

soyersoyer:
Update the parameter in the UI if setMasterVolume() is called from the MIDI side.


Banana71:
> Bug Fixes:

> Ensure Master Volume changes received via MIDI update the on-screen display

Hey @soyersoyer, 
the master volume settings work fine, the rest of the system runs flawlessly but the master volume display is not updated when changes come via MIDI Sysex.

Peter

MasterVolume MIDI-OX

soyersoyer:
Hey @Banana71,
I tried it now, it works for me, but only if I open the MiniDexed MIDI output port with Dexed or Drumstick.
But this is the case with other MIDI devices as well, probably because of ALSA (on Linux).
Does the volume change but not the GUI?
What happens if you send this MIDI Note on message?
90  3c  44

Banana71:
> Does the volume change but not the GUI?

Yes, the volume changes but the bar in the miniDexed display does not change
> 
> What happens if you send this MIDI Note on message? 90 3c 44

The note C3 is played with moderate velocity.

> 
> Or if you set the Master Volume to 0? f0 7f 7f 04 01 7f 00 f7

Then the sound from the miniDexed cannot be heard.

> 7f 3f is close to the default value and so the change is not visible on the UI.

7F corresponds to maximum volume.
3F is 50% of the volume.

Two bytes are sent, but only the second byte is used.

```
F0 7F 7F 04 01 7F 3F F7
or
F0 7F 7F 04 01 00 3F F7
```
both set the volume to 50% (3F)

The Sysex commands all do what they're supposed to, except the bar graph in the UI remains unchanged.

soyersoyer:
> The Sysex commands all do what they're supposed to, except the bar graph in the UI remains unchanged.

Are you sure you are using version 2025-07-25-7d9b39f?

> Two bytes are sent, but only the second byte is used.

Thinking about it, with this change we lose the lower 7 bits of the SysEx Volume message, as it will only be 7 bits inside the MiniDexed. Should the Master Volume be 14-bit inside the MiniDexed?

Banana71:
> Should the Master Volume be 14-bit inside the MiniDexed?

No, 7 bits are perfectly sufficient. The 7 bits are scaled to 16 divisions via the UI anyway.

Banana71:
> Are you sure you are using version 2025-07-25-7d9b39f?

Yes, but I'll check again :-)

Banana71:
Strange things are happening here.

[MiniDexed_1245_2025-07-25-5d9e136_64bit] - This version works, including updating the bar graph in the UI when using the Sysex command.


[MiniDexed_1246_2025-07-25-7d9b39f_64bit] - This version won't start at all anymore; the display remains dark and no sound is played.

I have to check a few things again; I'm pretty lost right now.

soyersoyer:
> No, 7 bits are perfectly sufficient. The 7 bits are scaled to 16 divisions via the UI anyway.

Ok. I made it 14bit, but it can remain 7bit.
If anyone needs it for SysEx, let me know.

Banana71:
> This works for me.
> 
> ```
> $ sha256sum kernel8.img 
> 6187962ca55b6987e42338cde9dedaf1796f7526f2c30a17e4a40830110bd954  kernel8.img
> ```

The RPi 4 loads kernel8-rpi.img, right? That's how it's described in the config.txt file.

Banana71:
The checksum is correct, but the RPi 4 won't boot anymore.

soyersoyer:
Sometimes it helps me if I copy the kernel to the SD card again.

soyersoyer:
SysEx volume was scaled incorrectly.

soyersoyer:
SetParameter has an assert for reverb, and SetParameter (ParameterMasterVolume, ...) was called before reverb initialization. It's strange that it still worked for me.

soyersoyer:
Very thanks for testing, this would have been difficult to catch later.

Banana71:
It starts again :-) great
The SysEx command also works, and the display is updated. Everything's fine.

I just noticed one thing: you divide the display into 14 fields, but it scales in 16 units?

Banana71:
Wow, you just changed a lot of code. Is there anything special I should pay attention to?

soyersoyer:
In the 'CUIMenu: pass nWide parameter to ToVolume functions' commit, the Volume and Pan handling has changed.

In the 'CUIMenu: do not use the full std namespace', I removed the using namespace std, which is a bad construction for some reason.

Banana71:
Oops, something went wrong
The line break is gone.

Display

soyersoyer:
Oops, fixed :)

Banana71:
I don't think anything has changed.

Should a larger display (20 characters) allow for the volume to be adjusted in finer increments?
I can adjust the volume more precisely than the bar indicates.

https://github.com/user-attachments/assets/ce443088-3b0d-4b15-8943-a2be5974c2a9

Banana71:
Actually, I'm not a fan of this bar display. I'd prefer a digital display from 0 to 11. :-)

https://www.youtube.com/watch?v=4xgx4k83zzc&t=36s&ab_channel=goto11tv

soyersoyer:
For me it displays 18 wide + the arrows.

Banana71:
> @Banana71 Could you try this one more time with [30a528b](https://github.com/probonopd/MiniDexed/commit/30a528b556a9af58475fe9f1b899fedf8e8b56b6)?

is now active again on my miniDexed. What should I look out for or try?

Banana71:
In this version, the bar is displayed across the entire width of the display. At two locations (approximately 25% and 75%), two bars are changed per click.

soyersoyer:
I added that it can also step one by one, and also print the volume and pan values.

Banana71:
Thank you @soyersoyer,
The ability to fine-tune the volume ( 0 - 127 ) is an absolute dream. 👍 

But this option was discussed here before and rejected by users who operate their MiniDexed exclusively with buttons. (I can understand that.)

I believe that the parameters should be fully utilized. With the original DX7, with only one voice, a coarse adjustment was perfectly sufficient, but with an 8-voice multi-timbral device, it is necessary to fine-tune the volume balance.

@probonopd, what do you think of the result?

Peter

PR #971, opened 2025-07-29
kernel: RegisterSystemThrottledHandler

soyersoyer:
It can be useful to log various system throttlings.

## Summary by Sourcery

Register a system throttling event handler to log throttle reasons and CPU status, and dump current CPU throttle status during kernel initialization.

New Features:
- Add SystemThrottledHandler to log under-voltage, frequency capping, throttling, and soft temperature limit events
- Register the SystemThrottledHandler during kernel initialization to handle system throttled states

Enhancements:
- Dump CPU throttle status immediately after configuration load in CKernel::Initialize

### DIFF
--- a/src/kernel.cpp
+++ b/src/kernel.cpp
@@ -88,10 +88,13 @@ bool CKernel::Initialize (void)
 	m_Config.Load ();
 
 	m_CPUThrottle.DumpStatus ();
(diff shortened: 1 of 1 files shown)


probonopd:
Can we make this opt-in via the config file?

soyersoyer:
> Can we make this opt-in via the config file?

Sure!

PR #973, opened 2025-07-30
move compressor to TG menu and add parameters

soyersoyer:
## Summary by Sourcery

Introduce per-tone generator compressor controls by adding enable, pregain, attack, release, threshold, and ratio parameters to performance configuration, synthesizer engine, and the UI, and move compressor settings from global effects to the TG menu.

New Features:
- Add per-TG compressor parameters (enable, pregain, attack, release, threshold, ratio) to CPerformanceConfig and CMiniDexed
- Expose compressor controls in the UI under a dedicated TG submenu with custom formatting for dB, ms, and ratio values

Enhancements:
- Move compressor settings from the global effects menu into the tone generator menu for individual TG configuration
- Provide compatibility handling for the legacy global compressor enable flag in performance loading

### DIFF
--- a/src/minidexed.cpp
+++ b/src/minidexed.cpp
@@ -116,6 +116,8 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 		
 		m_nReverbSend[i] = 0;
 
+		m_bCompressorEnable[i] = 1;
+
 		// Active the required number of active TGs
 		if (i<m_nToneGenerators)
 		{
@@ -253,8 +255,6 @@ CMiniDexed::CMiniDexed (CConfig *pConfig, CInterruptSystem *pInterrupt,
 	SetParameter (ParameterReverbDiffusion, 65);
 	SetParameter (ParameterReverbLevel, 99);
 	// END setup reverb
-
-	SetParameter (ParameterCompressorEnable, 1);
 
 	SetPerformanceSelectChannel(m_pConfig->GetPerformanceSelectChannel());
 		
@@ -1004,14 +1004,6 @@ void CMiniDexed::SetParameter (TParameter Parameter, int nValue)
 
 	switch (Parameter)
 	{
-	case ParameterCompressorEnable:
-		for (unsigned nTG = 0; nTG < m_nToneGenerators; nTG++)
-		{
-			assert (m_pTG[nTG]);
-			m_pTG[nTG]->setCompressor (!!nValue);
-		}
-		break;
-
 	case ParameterReverbEnable:
 		nValue=constrain((int)nValue,0,1);
 		m_ReverbSpinLock.Acquire ();
@@ -1132,6 +1124,8 @@ void CMiniDexed::SetTGParameter (TTGParameter Parameter, int nValue, unsigned nT
 
 	case TGParameterReverbSend:	SetReverbSend (nValue, nTG);	break;
 
+	case TGParameterCompressorEnable:	SetCompressorEnable (nValue, nTG); break;
+
 	default:
 		assert (0);
 		break;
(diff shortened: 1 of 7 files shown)


Banana71:
I'm impressed. 👍 

It's only possible to set the compressor for TG1. I can't access the compressor for TG2 - TG8. When you try to control the compressor via the TG2 - TG8 UI, it always jumps directly to TG1. The shortcut (pressing and turning the encoder) doesn't work in the compressor menu either.

The compressor itself works really well :-)

probonopd:
As we still have a comprehensive FX patch waiting to become ready, I'm hesitant to merge this. Also I guess this would break the performance format?

soyersoyer:
> It's only possible to set the compressor for TG1.

This surprised me, I fixed it. Thanks.

> Does the compressor only work with an RPi 4 or higher, or also with the RPi 2, RPi Zero 2, and RPi 3?

Just like before, from RPi1 as well. Only the configurability of the compressors has changed.

> Is your goal to combine multiple parts (TG9 - TG16, TG17 - TG24) and the dynamic extension?

Not necessarily like that. Part2, Part3 PR was more of a demo to see if it was worth doing something like that. It would be nice to somehow link the TGs together, to find a solution that would be good for larger Performances, Unison, Part things.

> What about compressed layered sounds? Would compressor groups perhaps be better (e.g., route TG1 - TG4 to compressor A, TG5 to compressor B, TG6 & TG7 to compressor C, etc.)

This is also a good idea. We'll keep that in mind when linking to TGs.

> Also I guess this would break the performance format?

This can open the current format, thus disabling all compressors if CompressorEnable is set and its value is 0.
However, previous MiniDexed versions will not catch the new parameters and a default compressor will be enabled. (even it is disabled for some TGs)

Banana71:
Hey, hey, hey, the compressor is fun. 
You can control it really well and it does whatever you want. Great.
If the ratio is set to 1:1, you can easily adjust the drums to a reasonable volume using pre-gain.
Saving and loading performances also works perfectly with the compressor parameters.

Crying over nothing serious:  A ratio of 5:1 is more of a limiter than a musical compressor. I think a starting value of 2:1 is more suitable.

Just a thought: Since I like to overdo it sometimes, a warning about clipping would be helpful. Is it possible to evaluate the volume and, if the volume exceeds, say, -3 dB, set a digital output for a warning LED?

Peter

soyersoyer:
The compressor in Synth_Dexed could be placed after MasterVolume and parameterized to act as a limiter. Is this a good idea?

Could it be parameterized in the Performance files also?

soyersoyer:
> The parameterizable compressor that you have integrated into the TGs, is it technically the same as the one we have always known as a limiter under effects?

Yes, it's the same. Look at this current [implementation](https://github.com/probonopd/MiniDexed/blob/8f2c638a675e69866920a9b096a3c90c0dd43dea/src/minidexed.cpp#L1007): if you enable the compressor in the Effects menu, it will enable all TG compressors.

> The compressor is well implemented, and I wish it were adopted for this project. Then I could bring all the performances to the same volume level. 

That's sounds good! With 32 TGs, layering of performances is also possible.

I'm going to add a limiter at the end of the chain.

Banana71:
> I'm going to add a limiter at the end of the chain.

If it were possible for you, a fully parameterizable compressor would be ideal (with the default parameters as a limiter)

soyersoyer:
I added the limiter before MasterVolume.

Parameters:
- LimiterEnable
- LimiterPreGain
- LimiterAttack
- LimiterRelease
- LimiterThresh
- LimiterRatio
- LimiterHPFilterEnable

soyersoyer:
Then we just have to wait for the integration :)

soyersoyer:
@probonopd What do you think? Do you need else in this?

Banana71:
I could offer to extend the wiki after the merge.🤖

Banana71:
I tested something else: Performance saving and loading works great.

Peter

PR #977, opened 2025-08-04
Fix Compilation with GCC 13.3 and upgrade to newer CMSIS

soyersoyer:


marchingband:
What is the plan with this? Will it make it into main?

marchingband:
What worked for me was cloning from https://github.com/soyersoyer/MiniDexed/tree/fixgcc133
Then I was able to compile under macos on apple silicone for 64bit arch (RPI > 2)

probonopd:
Thanks @soyersoyer

PR #980, opened 2025-08-08
InitNetwork IPAddress related fixes

soyersoyer:
This fixes the IP address setting error by improving IP address handling.

## Summary by Sourcery

Improve IP address handling by fixing configuration loading, switching getters to return const references, and enhancing network initialization to robustly select between DHCP and static addressing with detailed logging.

Enhancements:
- Change IP address getters to return const references to avoid unnecessary copies
- Update CConfig::Load to initialize IP address fields only when properties are present
- Revamp InitNetwork to branch between DHCP, static IP, or fallback modes, default missing gateway/DNS to zero, and log network parameters

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -206,18 +206,13 @@ void CConfig::Load (void)
 	m_bNetworkDHCP  = m_Properties.GetNumber ("NetworkDHCP", 0) != 0;
 	m_NetworkType = m_Properties.GetString ("NetworkType", "wlan");
 	m_NetworkHostname = m_Properties.GetString ("NetworkHostname", "MiniDexed");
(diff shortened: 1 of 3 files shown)


probonopd:
Thanks @soyersoyer. Please have a look at https://github.com/probonopd/MiniDexed/pull/980#pullrequestreview-3101569235, is there something to it?

soyersoyer:
> is there something to it?
Yes, but it won't match the style of the file, so no.

For example, unique_ptr could be used, but then the code would have to be rewritten everywhere so that it would not be inconsistent.

probonopd:
Thanks @soyersoyer

PR #1004, opened 2025-12-23
WIP: UI can be sent over udp

penfold42:
For your review only at this stage.

VERY! incomplete, but it does work.

If you're interested in getting this in I will:
- [ ]  Make it configurable!
- [ ] Look to eliminate Msg2. I may have to tweak the existing menu display code to explicitly move the cursor to the second line.

## Summary by Sourcery

Add support for mirroring the LCD user interface over UDP when networking is available.

New Features:
- Send the rendered two-line LCD menu display over UDP to a configured destination when the network stack becomes ready.

Enhancements:
- Extend the user interface component with UDP socket initialization and write helpers tied into the existing network startup flow.


probonopd:
What do you use on "the other side"?

penfold42:
Oops - forgot my cleanup would also tickle this PR.

> What do you use on "the other side"?

Just netcat for viewing
`nc -l -u 1306`
More details including a horrible shell script to fire udp midi keypresses here: 
https://github.com/probonopd/MiniDexed/discussions/1003

PR #1011, opened 2026-01-25
SH1106 display support

Sgw32:
**Summary**

Add support for SH1106 display 

**Problem**

Currently there was only support for similar SSD1306 devices. SH1106 is bigger and slightly better. 

**Solution**

Added a new class, highly derived from SSD1306 from Circle, but slightly different. Can be later used in Circle itself, eliminating the need to have it in MiniDexed project. 

**Testing**

Tested on BulkyMidi MiniDexed with SH1106 display on Raspberry Pi 3 by me.


penfold42:
Great work, i look forward to testing it on a jumbo sh1106 display :)

It might magically work on those integrated oled/rotary boards - IIRC they had a ssd1309 which is more like a sh1106 than ssd1306.

Have you considered getting this into circle first? @rsta2 has been pretty quick with merging stuff into the develop branch there. If so, it'll minimize rework here after you've forgotten what you did :)

Sgw32:
> Great work, i look forward to testing it on a jumbo sh1106 display :)
> 
> It might magically work on those integrated oled/rotary boards - IIRC they had a ssd1309 which is more like a sh1106 than ssd1306.
> 
> Have you considered getting this into circle first? @rsta2 has been pretty quick with merging stuff into the develop branch there. If so, it'll minimize rework here after you've forgotten what you did :)

Yes, implementing it directly in circle would be perfect. 

The important thing for MiniDexed and SH1106 is that it will work on original BulkyMidi MT32Pi by tebl, not on BulkyMidi Dexed so a lot of people having BulkyMidi can run MiniDexed on it as well.

blixabloxa:
> Great work, i look forward to testing it on a jumbo sh1106 display :)
> 
> It might magically work on those integrated oled/rotary boards - IIRC they had a ssd1309 which is more like a sh1106 than ssd1306.
> 
> Have you considered getting this into circle first? @rsta2 has been pretty quick with merging stuff into the develop branch there. If so, it'll minimize rework here after you've forgotten what you did :)

I directly swapped out a ssd1306 for an ssd1309. Worked perfectly with the same ssd1306 config in minidexed.ini. Great for my aging eyes!

penfold42:
So rsta2/circle dont want to add support:
```
The C*Device classes for dot-matrix displays in
addon/display/ are deprecated and the class CTerminalDevice should be
used instead. 

```
I propose we merge this PR into MiniDexed - not ideal, but gets SH1106 support in now.

PR #1037, opened 2026-06-09
Suppress serial error -1 on startup.

penfold42:
This is caused by the serial hardware detecting a break condition on startup.
In serial init just perform a single dummy read after a 5mS delay

Also provide more detail in mDNS publish log message


probonopd:
Thanks @penfold42. Please test the build above and let me know if you think it is ready for being merged.

penfold42:
> Thanks @penfold42. Please test the build above and let me know if you think it is ready for being merged.

it works :) merge away

